PR #70, opened 2019-03-28
README: note submodules are required in environment setup

pchickey:
A handful of folks have hit this :( . Blog post update is also in flight


pchickey:
bypassing travis for NFC

PR #75, opened 2019-03-29
lucet-wasi: use number_of_values(1) for the --dir option

jedisct1:
`multiple(true)` for the --dir option doesn't play well with modules also accepting multiple values for their arguments, resulting in convoluted command lines when both have to be used simultaneously.

This fixes it as recommended in the clap documentation for that very situation. `Arg::number_of_values(1)` doesn't prevent `--dir a1:b1 --dir a2:b2` from being used, but prevents `--dir a1:b1 a2:b2` to remove ambiguities.

### DIFF
--- a/lucet-wasi/src/main.rs
+++ b/lucet-wasi/src/main.rs
@@ -29,21 +29,17 @@ fn main() {
                 .help("Entrypoint to run within the WASI module"),
         )
         .arg(
-            Arg::with_name("lucet_module")
-                .required(true)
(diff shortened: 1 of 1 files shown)


jedisct1:
It does. 

As stated above:

`Arg::number_of_values(1) doesn't prevent --dir a1:b1 --dir a2:b2 from being used, but prevents --dir a1:b1 a2:b2 to remove ambiguities.`

The `--dir` option can be used multiple times.

Issue #76, opened 2019-03-29
lucet-runtime: heap size limit is hardcoded to 1 MB


jedisct1:
The max heap size defaults to 1 MB:

```rust
    fn default() -> Limits {
        Limits {
            heap_memory_size: 16 * 64 * 1024,
            heap_address_space_size: 0x200000000,
            stack_size: 128 * 1024,
            globals_size: 4096,
        }
    }
```

We don't seem to ever set it to what is stored in the wasm module, effectively limiting all applications to 1 MB.

acfoltzer-fastly:
Closed by #77

PR #77, opened 2019-03-29
Add CLI configuration for memory limits and improve performance for large initial heaps

### LINKED ISSUE #74: How to change memory limits for lucet-wasi
I get the following error when trying to run my wasm app:

```
$ lucet-wasi demo.so
thread 'main' panicked at 'instance can be created: LimitsExceeded("heap spec initial size: HeapSpec { reserved_size: 4194304, guard_size: 4194304, initial_size: 1114112, max_size: None }")', src/libcore/result.rs:997:5
```

It indeed requires much more memory than that, how do I allow it to allocate more?

update: it seems that limits are specified during the compilation. I've tried to set them to 1 Gb:

```
lucetc --reserved-size 1073741824 --guard-size 1073741824 pedersen_demo_bg.wasm -o demo.so
```

But I still get the same error

acfoltzer-fastly:
This adds command-line parameters to `lucet-wasi` to control the memory limits, and raises the default values of those limits substantially. Most users will only ever need to turn them down, not up.

Also, this reworks the way reserved size is set in `lucetc` to provide fewer surprises; see https://github.com/fastly/lucet/pull/77/commits/b563370b6b9e1b5821f2bfe288ec956da953638f.

Finally, this switches to relying on `madvise` to zero the instance heap when an instance is reset or created, substantially improving performance for programs with large initial heaps.

Closes #74 and #76

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -534,6 +534,7 @@ dependencies = [
  "cast 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.32.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "human-size 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.50 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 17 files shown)


jedisct1:
Hey Adam,

Can you fix the conflicts, so this can be merged? That would be great :)

PR #84, opened 2019-03-30
Implement the __wasi_poll_oneoff system call

jedisct1:
This implements the `__wasi_poll_oneoff` system call, which is required for a variety of tasks, such as waiting on descriptors with a timeout, as well as `sleep()`-like functions.


lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+            .unwrap();
+    let input_slice = unsafe { slice::from_raw_parts(input_slice_.0, input_slice_.1) };
+
+    let output_slice_ =
+        unsafe { dec_slice_of::<wasm32::__wasi_event_t>(&mut vmctx, output, nsubscriptions) }
+            .unwrap();
+    let output_slice = unsafe { slice::from_raw_parts_mut(output_slice_.0, output_slice_.1) };
+
+    let input: Vec<_> = input_slice.iter().map(|x| dec_subscription(x)).collect();
+
```
acfoltzer-fastly: are we missing functionality that wasmtime has? https://github.com/CraneStation/wasmtime/blob/master/wasmtime-wasi/sandboxed-system-primitives/src/posix.c#L2348

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+        .filter(|event| match event {
+            Ok(event) => event.type_ == wasm32::__WASI_EVENTTYPE_CLOCK,
+            _ => false,
+        })
+        .map(|event| {
+            let event = event.unwrap();
+            unsafe {
+                (
+                    wasi_clock_to_relative_ns_delay(event.u.clock) / 1_000_000,
+                    event.userdata,
```
acfoltzer-fastly: `unsafe` should go here, rather than around the whole tuple

lucet-wasi/src/memory.rs
```diff
@@ -357,3 +357,72 @@ dec_enc_scalar!(
...
+    dec_eventtype,
+    dec_eventtype_byref,
+    enc_eventtype,
+    enc_eventtype_byref
+);
+
+pub fn dec_subscription(
+    subscription: &wasm32::__wasi_subscription_t,
+) -> Result<host::__wasi_subscription_t, host::__wasi_errno_t> {
+    let userdata = subscription.userdata;
```
acfoltzer-fastly: needs `dec_userdata`

lucet-wasi/src/memory.rs
```diff
@@ -357,3 +357,72 @@ dec_enc_scalar!(
...
+    let type_ = dec_eventtype(subscription.type_);
+    let u_orig = subscription.__bindgen_anon_1;
+    let u = match type_ {
+        wasm32::__WASI_EVENTTYPE_CLOCK => host::__wasi_subscription_t___wasi_subscription_u {
+            clock: unsafe {
+                host::__wasi_subscription_t___wasi_subscription_u___wasi_subscription_u_clock_t {
+                    identifier: u_orig.clock.identifier,
```
acfoltzer-fastly: another `dec_userdata`

lucet-wasi/src/memory.rs
```diff
@@ -357,3 +357,72 @@ dec_enc_scalar!(
...
+        },
+        _  => return Err(wasm32::__WASI_EINVAL)
+    };
+    Ok(host::__wasi_subscription_t { userdata, type_, u })
+}
+
+pub fn enc_event(event: host::__wasi_event_t) -> wasm32::__wasi_event_t {
+    let fd_readwrite = unsafe { event.u.fd_readwrite };
+    wasm32::__wasi_event_t {
+        userdata: event.userdata,
```
acfoltzer-fastly: `enc_userdata`

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+    let now: u128 = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("Current date is before the epoch")
+        .as_nanos();
+    let deadline = wasi_clock.timeout as u128;
+    deadline.saturating_sub(now)
+}
+
+#[no_mangle]
+pub extern "C" fn __wasi_poll_oneoff(
```
acfoltzer-fastly: In terms of text, this is a beast of a function, but it doesn't actually seem all that complicated now that I've read through it. I think that if you added comments explaining each of the major steps (e.g., decoding subscriptions, splitting subscription types, converting to pollfd, polling, gathering results), it would be much easier to understand at first glance. Given that this is largely a pipeline rather than repeated operations on the same values, it would probably also be suitable for splitting some of these steps into helper functions and having a (relatively) simple driver function for the body of the hostcall.

lucet-wasi/src/memory.rs
jedisct1: `userdata` is an opaque pointer passed back and forth by the wasm module, and that only makes sense it its own context. It's something that the syscall should never ever touch. Is it worth decoding/encoding it rather than assume that it will always be in wasm representation?

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+                _ => unreachable!(),
+            };
+            nix::poll::PollFd::new(event.0, flags)
+        })
+        .collect();
+    let timeout = timeout
+        .map(|(timeout, userdata)| (cmp::min(timeout, c_int::max_value() as u128), userdata));
+    let poll_timeout = timeout.map(|timeout| timeout.0 as c_int).unwrap_or(-1);
+    let ready = loop {
+        break {
```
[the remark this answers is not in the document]
jedisct1: The `loop { break{` pattern is insanely cool and unique to Rust, I had to use it :)

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+    let input: Vec<_> = input_slice.iter().map(|x| dec_subscription(x)).collect();
+
+    let timeout = input
+        .iter()
+        .filter(|event| match event {
+            Ok(event) => event.type_ == wasm32::__WASI_EVENTTYPE_CLOCK,
+            _ => false,
+        })
+        .map(|event| {
+            let event = event.unwrap();
```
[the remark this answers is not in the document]
jedisct1: We may want to panic. If we cant' unwrap `event`, there is definitely a bug in the syscall code, not in the caller.

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+    input: wasm32::uintptr_t,
+    output: wasm32::uintptr_t,
+    nsubscriptions: wasm32::size_t,
+    nevents: wasm32::uintptr_t,
+) -> wasm32::__wasi_errno_t {
+    if nsubscriptions as u64 > wasm32::__wasi_filesize_t::max_value() {
+        return wasm32::__WASI_EINVAL;
+    }
+    let mut vmctx = unsafe { Vmctx::from_raw(vmctx) };
+    unsafe { enc_pointee(&mut vmctx, nevents, 0) }.unwrap();
```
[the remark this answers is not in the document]
jedisct1: Sure, thanks for pointing this out.

lucet-wasi/src/hostcalls.rs
jedisct1: wasmtime has a short path for the case where there is a single event, and that event is a timer. We do a couple useless operations in that exact case, but the functionality is the same.

lucet-wasi/src/memory.rs
acfoltzer-fastly: we can't be sure of that without overriding what bindgen produces for the host side, unfortunately

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+            nix::poll::PollFd::new(event.0, flags)
+        })
+        .collect();
+    let timeout = timeout
+        .map(|(timeout, userdata)| (cmp::min(timeout, c_int::max_value() as u128), userdata));
+    let poll_timeout = timeout.map(|timeout| timeout.0 as c_int).unwrap_or(-1);
+    let ready = loop {
+        break {
+            match nix::poll::poll(&mut poll_fds, poll_timeout) {
+                Err(_) | Ok(-1) => {
```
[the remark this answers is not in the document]
jedisct1: Just being paranoid here, but no problem, I will remove this.

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+            unsafe { enc_slice_of(&mut vmctx, &output_slice[..1], output) }.unwrap();
+            unsafe { enc_pointee(&mut vmctx, nevents, 1) }.unwrap();
+        }
+        return wasm32::__WASI_ESUCCESS;
+    }
+    let mut fd_events_cur = fd_events.iter();
+    let mut poll_fds_cur = poll_fds.iter();
+    let mut output_slice_cur = output_slice.iter_mut();
+    let mut revents_count = 0;
+    for _ in 0..ready {
```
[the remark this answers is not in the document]
jedisct1: Oh yes, good suggestion.

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
acfoltzer-fastly: oh, I misunderstood what this is unwrapping. could you change `filter` to `filter_map` so we don't have to reason across the two closures?

lucet-wasi/src/hostcalls.rs
acfoltzer-fastly: just to make sure I understand: we're using the timeout argument to `poll` in lieu of `clock_nanosleep` etc, and we're making all times relative by using `Instant::now()`?

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+                Ok(ready) => ready,
+            }
+        };
+    };
+    if ready == 0 {
+        if let Some((_, userdata)) = timeout {
+            let output_event = host::__wasi_event_t {
+                userdata,
+                type_: wasm32::__WASI_EVENTTYPE_CLOCK,
+                error: wasm32::__WASI_ESUCCESS,
```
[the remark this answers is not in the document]
jedisct1: For timers, we don't care about the value in that field. But if a module ever reads it, `ESCUCCESS` looks like the safest choice.

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+                            nbytes: 0,
+                            flags: 0,
+                        },
+                },
+            };
+            output_slice[0] = enc_event(output_event);
+            unsafe { enc_slice_of(&mut vmctx, &output_slice[..1], output) }.unwrap();
+            unsafe { enc_pointee(&mut vmctx, nevents, 1) }.unwrap();
+        }
+        return wasm32::__WASI_ESUCCESS;
```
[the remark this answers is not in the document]
jedisct1: `nevents` is the pointer where we store the number of events that happened (here, `1` - a timer fired).

Granted, `nevents` is a confusing name, but this is the one used in the WASI prototypes.

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
     return wasm32::__WASI_ESUCCESS;
 }
 
+nix::ioctl_read_bad!(fionread, nix::libc::FIONREAD, c_int);
```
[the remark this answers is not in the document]
jedisct1: done, thanks!

lucet-wasi/src/hostcalls.rs
jedisct1: done, thanks!

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +985,211 @@ pub extern "C" fn __wasi_random_get(
...
+        return wasm32::__WASI_ESUCCESS;
+    }
+    let mut poll_fds: Vec<_> = fd_events
+        .iter()
+        .map(|event| {
+            let mut flags = nix::poll::EventFlags::empty();
+            match event.1 {
+                wasm32::__WASI_EVENTTYPE_FD_READ => flags.insert(nix::poll::EventFlags::POLLIN),
+                wasm32::__WASI_EVENTTYPE_FD_WRITE => flags.insert(nix::poll::EventFlags::POLLOUT),
+                _ => unreachable!(),
```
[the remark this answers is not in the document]
jedisct1: done, thanks!

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
jedisct1: done, but it doesn't make me feel safe.

lucet-wasi/src/memory.rs
jedisct1: done.

lucet-wasi/src/memory.rs
jedisct1: done.

lucet-wasi/src/memory.rs
jedisct1: done.

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
acfoltzer-fastly: > (here, `1` - a timer fired).

is that true, though? if we hit the empty `else` of the `if let` above, wouldn't `nevents` still be set to `0` by L1016?

lucet-wasi/src/hostcalls.rs
jedisct1: Yes, exactly. Which is also what has to happen when there are also file descriptors to poll.

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
acfoltzer-fastly: hmm, do you have other suggestions that would address potential race conditions?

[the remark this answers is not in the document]
acfoltzer-fastly: oh, sorry, now that I see your subsequent commit I realize I wasn't being clear. when nix returns `Err(e)`, you can get the errno from the `e` using `e.as_errno().unwrap()` (I need to do a pass and remove all of those `unwrap()`s before someone accuses me of hypocrisy :wink:)

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
jedisct1: done.

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
jedisct1: done.

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
jedisct1: Oh, I see. Comment added.

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
jedisct1: done.

lucet-wasi/src/hostcalls.rs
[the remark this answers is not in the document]
jedisct1: Done.

lucet-wasi/src/hostcalls.rs
jedisct1: done

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +988,241 @@ pub extern "C" fn __wasi_random_get(
...
+            u: host::__wasi_event_t___wasi_event_u {
+                fd_readwrite: host::__wasi_event_t___wasi_event_u___wasi_event_u_fd_readwrite_t {
+                    nbytes: 0,
+                    flags: 0,
+                },
+            },
+        };
+        output_slice[0] = enc_event(output_event);
+        if let Err(e) = unsafe { enc_pointee(vmctx, nevents, 1) } {
+            return e;
```
acfoltzer-fastly: needs to be `enc_errno(e)` whenever we return an error from an `enc_` or `dec_`. it would be nice if bindgen could make these as newtype structs, so we could have the compiler remind us to do this :frowning_face:

lucet-wasi/src/hostcalls.rs
```diff
@@ -980,6 +988,241 @@ pub extern "C" fn __wasi_random_get(
...
+        }
+    }
+    return wasm32::__WASI_ESUCCESS;
+}
+
+fn __wasi_poll_oneoff_handle_fd_event(
+    vmctx: &mut Vmctx,
+    output_slice: &mut [wasm32::__wasi_event_t],
+    nevents: wasm32::uintptr_t,
+    events: Take<Zip<Iter<'_, FdEventData>, Iter<'_, nix::poll::PollFd>>>,
```
acfoltzer-fastly: `<I: Iterator<Item = (FdEventData, PollFd)>>` would be more clear here

lucet-wasi/src/memory.rs
```diff
@@ -357,3 +357,80 @@ dec_enc_scalar!(
...
+    subscription: &wasm32::__wasi_subscription_t,
+) -> Result<host::__wasi_subscription_t, host::__wasi_errno_t> {
+    let userdata = dec_userdata(subscription.userdata);
+    let type_ = dec_eventtype(subscription.type_);
+    let u_orig = subscription.__bindgen_anon_1;
+    let u = match type_ {
+        wasm32::__WASI_EVENTTYPE_CLOCK => host::__wasi_subscription_t___wasi_subscription_u {
+            clock: unsafe {
+                host::__wasi_subscription_t___wasi_subscription_u___wasi_subscription_u_clock_t {
+                    identifier: u_orig.clock.identifier,
```
acfoltzer-fastly: still need a `dec_userdata` here

jedisct1: My eyes can barely read anything any more, but I can't see any `userdata` here :/

acfoltzer-fastly: `identifier` is a `__wasi_userdata_t` :frowning_face:

jedisct1: Done.

lucet-wasi/src/hostcalls.rs
jedisct1: Done with `impl Iterator<Item = (&'t FdEventData, &'t nix::poll::PollFd)>`.

lucet-wasi/src/hostcalls.rs
jedisct1: done

PR #88, opened 2019-04-02
Add a more comprehensive .gitignore

freebroccolo:
Generated with the following:

```
#!/bin/sh
echo -e "host\n" > .gitignore
gibo dump Archives Autotools Backup C C++ CMake Diff Dropbox Emacs Images macOS Linux Ninja Node Patch Python Rust SublimeText Tags TextMate VirtualEnv VisualStudio VisualStudioCode vim Windows Xcode >> .gitignore
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,1129 @@
-target/
-*.rs.bk
-*.pyc
 host
+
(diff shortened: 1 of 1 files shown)


pchickey:
Hi Darin - we'd prefer that users maintain their own global gitignore file (https://stackoverflow.com/questions/7335420/global-git-ignore) with all of these sorts of extensions, and keep just project-specific stuff in the .gitignore file. *.pyc probably shouldn't even be in the project's gitignore.

PR #90, opened 2019-04-02
Add .gitattributes

freebroccolo:
This adds a `.gitattributes` file. This is mostly for the benefit of folks on Windows.

For reference: [rust-lang/rust/.gitattributes](https://github.com/rust-lang/rust/blob/f694222887cf31f51e68927716c25736e62f037f/.gitattributes#L3).


pchickey:
Thanks!

PR #91, opened 2019-04-02
Pin rustup toolchain to stable

freebroccolo:
This pins the rustup toolchain to "stable" within this project. This is mostly for the benefit of users that have defaulted to a different toolchain version globally and are not using the docker based environment.


jedisct1:
Merged, thanks!

freebroccolo:
Does `master` actually compile on `stable` at the moment? When I'm trying that just now I'm getting errors but it works with `nightly-2019-03-23` which I also have installed.

jedisct1:
It should compile on `stable` if `rust-std-1.33.0-wasm32-unknown-emscripten` is also installed.

jedisct1:
The Docker environment only includes stable version `1.33.0` and everything seems to compile just fine with it :)

PR #92, opened 2019-04-03
Fix url in comments

11Takanori:
`https://webassembly.github.io/spec/syntax/modules.html#data-segments` is not found.


jedisct1:
Good catch, thank you!

PR #93, opened 2019-04-03
Add Csmith-based fuzzing

acfoltzer-fastly:
This executable:

- Generates C programs with [Csmith](https://embed.cs.utah.edu/csmith/)
- Compiles and runs them natively via `cc`
- Compiles and runs them via `wasi-sdk`, `lucetc`, and `lucet-wasi`
- Reports the Csmith seed used when the output of these two pipelines doesn't match

Additionally, the `lucet-wasi-fuzz --seed N` argument allows you to rerun a single test with a given seed.


jedisct1:
Cool stuff!

jedisct1:
Why do we need WASI at all here?

Wouldn't it be faster to not even try to link WASI?

lucet-wasi-fuzz/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+        .arg("-o")
+        .arg(gen_c_path.as_ref())
+        .status()?;
+    Ok(())
+}
+
+fn run_native<P: AsRef<Path>>(tmpdir: &TempDir, gen_c_path: P) -> Result<Option<Vec<u8>>, Error> {
+    let gen_path = tmpdir.path().join("gen");
+
+    Command::new("cc")
```
jedisct1: Wouldn't using the `cc` crate be better? It will improve portability, and the compiler and default options could be changed in a standard way.

acfoltzer-fastly:
> Why do we need WASI at all here?
> 
> Wouldn't it be faster to not even try to link WASI?

Csmith-generated programs output their checksum with `printf`, so we at least need stdio.

> Wouldn't using the `cc` crate be better? It will improve portability, and the compiler and default options could be changed in a standard way.

I didn't think `cc` was a general-purpose tool that could produce executables; I thought it was only for building libraries in `build.rs`?

lucet-wasi-fuzz/src/main.rs
acfoltzer-fastly: I didn't think `cc` was a general-purpose tool that could produce executables; I thought it was only for building libraries in `build.rs`?

jedisct1: As long as `out_dir`, `target`, `opt_level` and `host` are defined, `cc` works anywhere.

jedisct1: But this fuzzer is mainly for us, and it will only be used on one platform, so directly building the command line doesn't make a big difference anyway.

acfoltzer-fastly: Hmm, I tried this and still get a whole bunch of output like:

```
...
cargo:warning=                   ^
cargo:warning=/tmp/.tmp8uDaNl/gen.c:200:22: warning: 'g_2528' defined but not used [-Wunused-variable]
cargo:warning= static struct S1 ****g_2528 = (void*)0;
cargo:warning=                      ^
exit code: 0
AR_x86_64-unknown-linux = None
AR_x86_64_unknown_linux = None
HOST_AR = None
AR = None
running: "ar" "crs" "/tmp/.tmp8uDaNl/libgen.a" "/tmp/.tmp8uDaNl/gen.o"
exit code: 0
cargo:rustc-link-lib=static=gen
cargo:rustc-link-search=native=/tmp/.tmp8uDaNl
```

jedisct1: `.flag_if_supported("-w")`

acfoltzer-fastly: The problem isn't the warnings, it's all the `cargo:*` lines for `build.rs` programs, which I discovered `build.cargo_metadata(false)` can disable. However it's still wanting to produce a library at the end, rather than an executable. I think I give up for now :slightly_smiling_face:

PR #94, opened 2019-04-03
Build a "lucet" layer on top of "lucet-dev"

jedisct1:
We end up with two images.

The former being just the base development environment, suitable for CI.

The later having the release builds populated in /opt/lucet.
This is the one we can upload to the Docker Hub, as people can use it to get immediately started with Lucet without having to recompile it.

Also remove the wasi debian package after installation by the way.


acfoltzer:
Is this PR ready to be reviewed? I know we talked about having an executables-only image for distribution so that the extra libraries needed for fuzzing don't inflate downloads too much; this PR might be at cross-purposes with that?

jedisct1:
This PR is still waiting for a review.

Even with the forthcoming split into two images, this PR is a first step and is still worth merging now. Travis only builds the dev environment, but users get the precompiled binaries ready to use.

PR #95, opened 2019-04-04
teach lucet-wasi to be smarter about build failure modes

awortman-fastly:
if something in lucet-wasi's build.rs fails right now, it can result in bindgen just not running and odd failures about missing methods later on


acfoltzer:
Going ahead and merging this, as the test failures are the spurious Travis variety

PR #96, opened 2019-04-04
Fix benchmarks/shootout makefile

awortman-fastly:
Some options used here seem to have bitrotted some. `--reserved-size` was split into `--{min,max}-reserved-size` but it looks to be that the spirit of that is best expressed as `--max-reserved-size` now

### DIFF
--- a/benchmarks/shootout/Makefile
+++ b/benchmarks/shootout/Makefile
@@ -1,6 +1,6 @@
 LUCET_ROOT:=$(abspath ../..)
 
-WASI_SDK?=/opt/wasi
+WASI_SDK?=/opt/wasi-sdk
 WASI_CC=$(WASI_SDK)/bin/clang
 
 LUCETC:=$(LUCET_ROOT)/target/release/lucetc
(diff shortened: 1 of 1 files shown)


jedisct1:
We'll be measuring two different things with `--min` vs `--max`. With `--max` Cranelift is indeed unlikely to do bound checking elision. I think the previous behavior was similar to `--max`.

acfoltzer-fastly:
Previously, `--reserved-size` just set the reserved size unconditionally. Now, since we take the max of `--min-reserved-size` and the size of the module's initial heap, we probably want to set the `min` to get the same behavior (unless some of the benchmarks request an initial heap that large)

awortman-fastly:
I hadn't thought about the bounds check elision - I went and measured both and right now things are 10%ish faster across the board with `min`, surprise! `min` does in fact get the same `HeapSpec` built as was there before the setting was split, so I'll switch that over

acfoltzer:
Going ahead and merging this, as the test failures are the spurious Travis variety

PR #97, opened 2019-04-04
Fuzzing: Creduce driver and test predicate improvements

acfoltzer-fastly:
Notably, when compiling the native code with `-m32`, the error rate is low enough that a run of 10,000 tests yielded no failures. Yay!


acfoltzer-fastly:
@jedisct1 when you're available, could you look at the Dockerfile changes here and make sure it's not blowing up the size of the image too much? Docker Hub says our image is ~600MB but this image on my local machine is 2GB. Is that just because of how Docker Hub counts sizes of images vs. layers?

jedisct1:
The Docker changes add 288 MB to the image, 147 MB to the compressed image.

That's quite a bit. Maybe these dependencies can be installed on demand?

PR #106, opened 2019-04-08
Remove unexpected build flag --reserve-size

jlb6740:
When building using the shootout makefile in the container with:

cd /lucet
make bench

or outside with:
./devenv_run.sh make bench

an error occurs:

error: Found argument '--reserved-size' which wasn't expected, or isn't valid in this context
        Did you mean --min-reserved-size?

This patch removes the unrecognized flag.


jlb6740:
Looks like this has been patched. Will abandon.

PR #109, opened 2019-04-09
Just a massive update to lucet-idl

jedisct1:
This PR changes pretty much everything over the previous code.


lucet-idl/src/data_description_helper.rs
```diff
@@ -0,0 +1,171 @@
...
+/// Transforms a `DataDescription`
+/// We definitely need a better name for it
+#[derive(Debug, Clone)]
+pub struct DataDescriptionHelper {
+    pub data_description: DataDescription,
+}
+
+impl DataDescriptionHelper {
+    /// Retrieve information about a data type given its identifier
+    pub fn get(&self, id: DataTypeId) -> DataTypeEntry<'_> {
```
awortman-fastly: Could this move to be on `impl DataDescription` itself? The remainder would then be just functions writing out for whichever backend, making for a name like `DataDescriptionWriter` or something along those lines.

jedisct1: `DataDescription` only belongs to the lexer. It isn't supposed to know anything about code generation.

Unfortunately, that structure is not very convenient to work with for code generation.

Granted, `DataDescriptionHelper` is a bad and confusing name. I'm not good at naming things, so suggestions are welcome :)

pchickey:
CI indicates that this requires a minor update to Cargo.lock to be checked in

pchickey:
Rebased on master to correct the CI problem. Once it passes we can merge

PR #111, opened 2019-04-09
Add microbenchmark suite

acfoltzer:
As a complement to the shootout benchmarks, these benchmarks measure Lucet performance at a very granular level.


benchmarks/lucet-microbenchmarks/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+use std::path::Path;
+use std::sync::Arc;
+use tempfile::TempDir;
+
+fn wasi_bindings() -> Bindings {
+    Bindings::from_file("../../lucet-wasi/bindings.json").unwrap()
+}
+
+fn compile_hello<P: AsRef<Path>>(so_file: P) {
+    let wasm_build = Lucetc::new(&["../../lucet-wasi/examples/hello.c"])
```
pchickey: I would rather have a duplicate of this source file inside the lucet-microbenchmarks directory, these sort of cross-repo file paths have created trouble when we've refactored stuff in the past.

lucet-wasi-sdk/src/lib.rs
```diff
@@ -42,35 +59,49 @@ pub struct Compile {
...
+impl CompileOpts for Compile {
+    fn cflag<S: AsRef<str>>(&mut self, cflag: S) {
+        self.cflags.push(cflag.as_ref().to_owned());
     }
 
-    pub fn cflag<S: AsRef<str>>(mut self, cflag: S) -> Self {
-        self.with_cflag(cflag);
+    fn with_cflag<S: AsRef<str>>(mut self, cflag: S) -> Self {
+        self.cflag(cflag);
         self
```
pchickey: sorry i got these wrong :/ its hard to remember the conventions

benchmarks/lucet-microbenchmarks/src/lib.rs
acfoltzer: for sure; this is temporary until I finish the PR

lucet-wasi-sdk/src/lib.rs
```diff
@@ -186,6 +185,120 @@ impl Link {
...
+    tmpdir: TempDir,
+    wasm_file: PathBuf,
+}
+
+impl Lucetc {
+    pub fn new<P: AsRef<Path>>(input: &[P]) -> Self {
+        let link = Link {
+            input: input.iter().map(|p| PathBuf::from(p.as_ref())).collect(),
+            cflags: Vec::new(),
+            ldflags: Vec::new(),
```
awortman-fastly: this skips around the `--no-threads` option that was just added in `Link::new` - I think you could phrase this as `Link::new(input)` and get that plus less code here

acfoltzer: You found my copypasta...

Should we just have `--no-threads` as a default option?

awortman-fastly: that's what the hope of doing it in `new` was :) I didn't see any places `Link` was constructed outside that

acfoltzer: Ah, my branch must not be up to date then. Time to rebase, because yes we do want that option everywhere

PR #119, opened 2019-04-11
Update Rust to version 1.34.0

jedisct1:
No apparent regression.


acfoltzer:
It sounds like there's going to be a point release very soon to fix some compiler crashes, so let's hold off until 1.34.1 is available

PR #120, opened 2019-04-12
Make lucetc work on macOS

jedisct1:
Also allow the linker and its flags to be overridden using standard environment variables.


jedisct1:
Travis supports macOS.

But if we use Travis for this, it will significantly increase the build time, and this is not really something we can afford.

PR #125, opened 2019-04-16
Update sightglass again

jedisct1:


pchickey:
We can't figure out why PR #105 didn't work - it merged but the submodule never got updated.

PR #126, opened 2019-04-16
Cargo.lock update after sightglass update

jedisct1:


pchickey:
so hwloc is still in the cargo.toml but it wont be built on linux, right? I am testing locally but lets let CI double check that we got it right

PR #130, opened 2019-04-17
Fix setenv script name

daisukefuji:
In [this PR](https://github.com/fastly/lucet/pull/49), `lucet_setenv` was moved to `devenv_setenv`. So now you will see this error.
```
# source /opt/lucet/bin/lucet_setenv.sh
bash: /opt/lucet/bin/lucet_setenv.sh: No such file or directory
```


jedisct1:
Good catch, thanks!

PR #131, opened 2019-04-17
gitattributes: note that png and jpg are binary files

pchickey:
the assets/*.png files were getting CRLF'd up during rebases

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,7 @@
 # Auto detect text and normalize to LF on commit regardless of "core.autocrlf".
 # See: https://help.github.com/en/articles/dealing-with-line-endings#example
 * text=auto eol=lf
+
+# Denote all files that are truly binary and should not be modified.
+*.png binary
+*.jpg binary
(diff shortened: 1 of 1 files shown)


jedisct1:
Shouldn't `*.wasm` be added as well?

jedisct1:
We have quite a few other binary files around (ex: `deploy_key.enc`, `pldi2017.pdf`). Wondering whether keeping this setting around is a good idea or if it may cause more issues that it solves.

pchickey:
Those are in the wasm spec submodule, and haven't caused a problem so far. I'm also not sure if keeping this setting around is worthwhile, if it causes a problem again lets delete it.

pchickey:
@freebroccolo do you have any advice on this issue?

freebroccolo:
@pchickey This looks good. I would probably adopt most of what's in [rust-lang/rust/.gitattributes](https://raw.githubusercontent.com/rust-lang/rust/master/.gitattributes) too. I notice now that they had a comment at the end about images. Sorry I missed that in the original PR.

You also might want to have a look at some of the templates at [alexkaratarakis/gitattributes](alexkaratarakis/gitattributes).

PR #132, opened 2019-04-17
Move trap information to lucet-module-data

awortman-fastly:
This is a start to moving several pieces of information to `lucet-module-data`, from more ad-hoc serialization/deserialization elsewhere in `lucetc`/`lucet-runtime`


lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -71,31 +73,106 @@ impl DlModule {
...
+        // the trap manifest, which should point to the table in a readonly section
+
+        for (idx, record) in code_metadata.trap_manifest.iter_mut().enumerate() {
+            let function = function_manifest.get(record.func_index as usize).ok_or(
+                lucet_incorrect_module!(
+                    "trap manifest includes an invalid function index: {}",
+                    record.func_index,
+                ),
+            )?;
+            let fn_name = unsafe {
```
awortman-fastly: This is super hairy, but works for the moment, so at least it's a working changeset to think about

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -188,7 +265,11 @@ impl ModuleInternal for DlModule {
     }
 
     fn trap_manifest(&self) -> &[TrapManifestRecord] {
-        self.trap_manifest
+        &self.code_metadata.trap_manifest
```
awortman-fastly: Also not very sold on exposing `trap_manifest` here since it's moved, but if this reads fine, cool

lucetc/src/decls.rs
```diff
@@ -113,7 +113,7 @@ impl<'a> ModuleDecls<'a> {
                 if exportable_sigix.export_names.is_empty() {
                     let def_symbol = format!("guest_func_{}", ix);
                     let funcid = clif_module
-                        .declare_function(&def_symbol, Linkage::Local, signature)
+                        .declare_function(&def_symbol, Linkage::Export, signature)
```
awortman-fastly: This is only for the `dladdr` hackery above

lucetc/src/function.rs
```diff
@@ -256,3 +263,75 @@ impl<'a> FuncEnvironment for FuncInfo<'a> {
...
+     *
+     * `.link()` is declare a relocation in the artifact
+     * relocating the bytes that HAPPEN to be the "addr"
+     * part of a FunctionSpec to point to the right function
+     *
+     * If `addr` moves in `FunctionSpec`, this will break!
+     */
+    for sink in manifest.sinks.iter() {
+        function_manifest.push(FunctionSpec::new(
+            0, // TODO: This is a lie. We don't have function address information at this point.
```
awortman-fastly: Is there any kind of hint or initial value we can place here? I'd prefer to have a real address to the function, and a relocation to fix up when opened, but I think that also isn't exposed yet?

lucetc/src/output.rs
```diff
@@ -43,7 +73,22 @@ impl ObjectFile {
...
+
+        // TODO: at this moment there is no way to get a full list of functions and sizes
+        // at this point in compilation.
+        //
+        // For now, we need the list of functions with traps, which we can get here, and
+        // reuse that when writing out the trap manifest. When a full function list is
+        // available, `write_function_manifest` should take the function manifest, rather
+        // than compute it
+        let function_manifest = write_function_manifest(trap_manifest, &mut product.artifact)?;
+        let internal_trap_manifest = write_trap_tables(trap_manifest, &mut product.artifact)?;
```
awortman-fastly: `write_trap_tables` happens to not need `function_manifest` as a coincidence at this moment, but if `function_manifest` were a full list, or reordered, it would be necessary to look up functions there.

I'll actually change `write_trap_manifest` to assume `function_manifest` might be any ol list, just so this isn't an issue in the future

lucetc/src/traps.rs
```diff
@@ -1,101 +1,52 @@
...
+    obj: &mut Artifact,
+) -> Result<Vec<TrapManifestRecord>, Error> {
+    let mut trap_manifest: Vec<TrapManifestRecord> = Vec::with_capacity(manifest.sinks.len());
 
-    let manifest_sym = "lucet_trap_manifest";
-    obj.declare(&manifest_sym, Decl::data().global())
-        .context(format!("declaring {}", &manifest_sym))?;
+    for (i, sink) in manifest.sinks.iter().enumerate() {
+        trap_manifest.push(TrapManifestRecord {
+            table_addr: 0, // This will be fixed up when loaded
```
awortman-fastly: Same hinting question applies here, is there something other than `0` I could place here?

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
awortman-fastly: Thinking about it more, I think avoiding having a relocation in `TrapManifestRecord` is the wrong way to go about this - the address of the slice must be be somewhere, this just gets to it in a roundabout `dladdr` matter. Before the last commit on this branch I still had `TrapManifestRecord` written out with a relocation on `table_addr`, for comparison.

Upside is, I think, that we could get away with zero-copy deserialization of a module's `[TrapManifestRecord]` just like trap sites, and the same for other similar kinds of data (thinking unwind information, primarily)

pchickey: yep, lets avoid depending on `dladdr` in any of the new code, and also using `dlsym` (via `lib.get` for `trap_manifest_ptr`) to construct symbols from strings. Instead, I'd change `FunctionSpec` to have the fields `pub code: *const c_void, pub traps: *const c_void, len: u32` and, when serializing, use relocations for the first two members to point to the code and trap symbols.

lucetc/src/function.rs
pchickey: I think `0` is fine as an initial value. We'll need testing to make sure the relocation ends up working correctly

lucetc/src/output.rs
pchickey: I think traps can just be a member of each function manifest. That isn't the design we wrote down originally, but I don't see any reason for them to be different, do you?

awortman-fastly: Only if some data associated with a function might be sparse enough to justify not wanting a bunch of null pointers. Since most functions have some possibility of trapping, I agree

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -186,8 +196,8 @@ impl ModuleInternal for DlModule {
         }
     }
 
-    fn trap_manifest(&self) -> &[TrapManifestRecord] {
```
awortman-fastly: the irony is not lost on me that keeping code and trap table pointers together results in a change that's more renaming `TrapManifest` to `FunctionManifest`

lucet-module-data/src/functions.rs
```diff
@@ -0,0 +1,62 @@
...
+        FunctionSpec { code_addr, code_len, traps_addr, traps_len }
+    }
+    pub fn code_len(&self) -> u32 {
+        self.code_len
+    }
+    pub fn traps_len(&self) -> u64 {
+        self.traps_len
+    }
+    pub fn contains(&self, addr: u64) -> bool {
+        // TODO This *may* be an off by one - replicating the check in
```
tyler: Did you end up confirming this?

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -71,31 +69,43 @@ impl DlModule {
...
-                from_raw_parts(records, *len as usize)
-            } else {
-                &[]
+                    let manifest = ptr.as_ref().ok_or(lucet_incorrect_module!(
+                        "`lucet_function_manifest` is defined but null"
+                    ))?;
+
+                    from_raw_parts(manifest, *manifest_len as usize)
+                }
+                // TODO: Can a module be data-only?
```
pchickey: technically, it is valid for a webassembly module to define no functions

lucetc/src/traps.rs
```diff
@@ -109,19 +49,19 @@ fn trap_sym_for_func(sym: &str) -> String {
...
+        ir::TrapCode::HeapOutOfBounds => lucet_module_data::TrapCode::HeapOutOfBounds,
+        ir::TrapCode::OutOfBounds => lucet_module_data::TrapCode::OutOfBounds,
+        ir::TrapCode::IndirectCallToNull => lucet_module_data::TrapCode::IndirectCallToNull,
+        ir::TrapCode::BadSignature => lucet_module_data::TrapCode::BadSignature,
+        ir::TrapCode::IntegerOverflow => lucet_module_data::TrapCode::IntegerOverflow,
+        ir::TrapCode::IntegerDivisionByZero => lucet_module_data::TrapCode::IntegerDivByZero,
+        ir::TrapCode::BadConversionToInteger => lucet_module_data::TrapCode::BadConversionToInteger,
+        ir::TrapCode::Interrupt => lucet_module_data::TrapCode::Interrupt,
+        ir::TrapCode::TableOutOfBounds => lucet_module_data::TrapCode::TableOutOfBounds,
+        ir::TrapCode::UnreachableCodeReached => lucet_module_data::TrapCode::Unreachable,
```
pchickey: yesss

lucetc/src/function.rs
```diff
@@ -259,3 +266,94 @@ impl<'a> FuncEnvironment for FuncInfo<'a> {
         Ok(*pos.func.dfg.inst_results(inst).first().unwrap())
     }
 }
+
```
pchickey: My only substantive comment is that these fns could probably live in another file (func_manifest.rs) because the rest of this file is dedicated to implementing the FuncEnvironment trait used by cranelift-wasm. The two concepts are pretty disjoint.

lucet-module-data/src/functions.rs
awortman-fastly: yep! By looking through a compiled module. The test case for this is convoluted enough it'd be tricky enough.

Since cranelift pads to 16-byte alignment on x86_64, this is only an observable bug if two functions are adjacent with no padding and the first byte of the second function is in an instruction that could and does trap. I'm not sure if it's possible to convince cranelift to not emit `push rbp` as the first instruction, so that would have to be a stack overflow right now

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
awortman-fastly: cool. adjusted this match arm in light of that

PR #133, opened 2019-04-18
Check that INSTANCE_PADDING is correct

jedisct1:
This hard-coded padding size has been the cause of headaches in the original C version, because it may need to be adjusted every time a change is made to the `Instance` structure.

Furthermore, the context is platform-dependent, so the padding cannot be one-size-fits-all.

The globals pointer must be right at the end of the page, and the last member of the structure. This is documented, but not enforced in code.

Reintroduce the checks making sure that this is the case.

The `INSTANCE_PADDING` constant remains fragile, and we can probably find a better way to handle this. Even on Linux, I'm not convinced that the current value is correct on non-x86_64 architectures.

The check can save quite a bit of debugging time when it's off.

Also add the current value for macOS.

We probably don't want a catchall value, as it is likely to be wrong on untested platforms.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance.rs
@@ -3,7 +3,7 @@ pub mod signals;
 
 pub use crate::instance::signals::{signal_handler_none, SignalBehavior, SignalHandler};
 
-use crate::alloc::Alloc;
+use crate::alloc::{host_page_size, Alloc};
 use crate::context::Context;
 use crate::embed_ctx::CtxMap;
 use crate::error::Error;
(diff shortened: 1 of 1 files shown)


acfoltzer-fastly:
Was this test not already catching this? https://github.com/fastly/lucet/blob/9e9d34b14c1f3cc1b988a340e6677fac5f4289b8/lucet-runtime/lucet-runtime-internals/src/instance.rs#L858

jedisct1:
When the padding is incorrect, the global containing the stack address is wrong, and half the tests are failing. Finding the root cause is not obvious.

With the assertion being hit as soon as an instance is created, it immediately becomes obvious.

The rest of the code assumes that the structure has that specific and unusual layout. That check being present in the code itself documents this and improves clarity.

acfoltzer-fastly:
Yeah, it's unfortunate that you can't enforce ordering of tests, though we could via the Makefile by specifying `instance` tests before the others. But I don't think that's great either.

Perhaps we could combine this check with the way we assert host page size at runtime? https://github.com/fastly/lucet/blob/9e9d34b14c1f3cc1b988a340e6677fac5f4289b8/lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs#L15

This also reminds me that we should be making some of these checks conditionally compile only in debug mode, but I can take care of that separately if you want.

jedisct1:
Closing, as we are going to remove hard-coded padding length completely 🎉

PR #138, opened 2019-04-19
Replace hard-coded link flags with portable link options

jedisct1:
Instead of hard-coded, linker-specific command-line flags, define a portable set of options to express the intent.

These options are then converted into the correct set of command-line flags for the given linker.

We may eventually want to do the same thing for the compiler, but this is not really necessary for now since clang is all we support.

Also make `ucontext` inclusion more portable to non-glibc environments.

### DIFF
--- a/lucet-builtins/Makefile
+++ b/lucet-builtins/Makefile
@@ -17,6 +17,7 @@ build/%.o: src/%.c
 	$(CC) $(COMMON_CFLAGS) -c $^ -o $@
 
 build/libbuiltins.so: $(LIBBUILTINS_OBJS)
+	$(CC) -Wl,-U,_lucet_vmctx_current_memory -Wl,-U,_lucet_vmctx_get_heap -shared $^ -o $@ 2> /dev/null || \
 	$(CC) -shared $^ -o $@
 
 .PHONY: clean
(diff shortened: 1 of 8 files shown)


lucet-builtins/Makefile
```diff
@@ -17,6 +17,7 @@ build/%.o: src/%.c
 	$(CC) $(COMMON_CFLAGS) -c $^ -o $@
 
 build/libbuiltins.so: $(LIBBUILTINS_OBJS)
+	$(CC) -Wl,-U,_lucet_vmctx_current_memory -Wl,-U,_lucet_vmctx_get_heap -shared $^ -o $@ 2> /dev/null || \
 	$(CC) -shared $^ -o $@
```
pchickey: I think you meant to delete the second invocation of CC

lucet-wasi-sdk/src/lib.rs
```diff
@@ -221,28 +221,88 @@ impl AsLink for Link {
     }
 }
```
pchickey: This is a great design!

lucet-builtins/Makefile
jedisct1: This is intentional - the second invocation of `CC` is called if the first one fails (notice the `||` operator between them). Declaring only specific symbols as weak doesn't work on all llvm versions.

pchickey: Oh, nice. I totally missed the `||`. Could you put a comment explaining this? Its not intuitive, but quite smart.

jedisct1: Added.

PR #144, opened 2019-04-22
lucet-runtime: fix deb packaging by deleting ref to deleted file

pchickey:


pchickey:
unfortunately our CI doesnt cover this but i will vouch for it having passed in private CI

PR #147, opened 2019-04-23
Upgrade to wasi-sdk 4.0

pchickey:
Just upgrades the package installed in the Dockerfile. The only fix required went in as part of #146


jedisct1:
Good!

I'm going to update the Docker image with the new SDK right away.

PR #150, opened 2019-04-23
Timeouts

tyler:


singpolyma-shopify:
This looks really great.  Is there anything someone like me could be doing to help move this feature forward, or would that involve more effort than it's worth from you?

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -174,37 +183,64 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
             }
             SignalBehavior::Terminate => {
                 // set the state before jumping back to the host context
+                //
+                // we must clear the terminable flag so that no KillSwitch accidentally signals a
+                // host thread.
+                inst.kill_state.terminable.store(false, Ordering::SeqCst);
```
awortman-fastly: This `store` and the store added on line 214 are the two most questionable lines of this PR, as far as correctness is concerned, IMO.

I want to prevent the race where _while_ this handler is _finishing_ with a SIGSEGV, some timer decides the instance has timed out, sends a SIGALRM, and ends up sending it to host code. Since this handler masks SIGALRM among others, an alarm raised at or before this point will be blocked. If the instance were terminable (so we got here through a non-alarm fault), exactly one `KillSwitch` may fire, and if it occurs after these `store` that lead to host code, it will fail to get the flag, and not signal.

Because a `KillSwitch` may have fired before (to get us here), or during, this handler, the flag may have been taken, and we can't assert the flag is either `true` or `false`. That leaves just overwriting the flag, and is what makes me question these lines. Because in both cases we will exit to the host, I'm at least confident we won't end up with an unterminable guest.

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -164,6 +166,13 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
                 .as_mut()
         };
 
+        if signal == Signal::SIGALRM {
```
awortman-fastly: We could `assert!(!inst.kill_state.terminable.load(Ordering::SeqCst))` here, just because an alarm without marking the instance unterminable would be an error, but I'm not very opinionated on this. I'm pretty sure it would just serve to make a spurious sigalrm into a host-crashing assert fail.

lucet-runtime/lucet-runtime-internals/src/vmctx.rs
```diff
@@ -363,6 +364,16 @@ impl Vmctx {
             val: YieldedVal::new(val),
             expecting: expecting as Box<dyn Any>,
         };
+
+        let terminable = inst.kill_state.terminable.swap(false, Ordering::SeqCst);
+
+        if !terminable {
```
awortman-fastly: This and the small `jz _hang` loop in `context_asm.S` are two places I think it's best to just spin for a moment - if a processor stays in them for even microseconds I'd be surprised and concerned, but it's entirely dependent on how long signal delivery takes.

lucet-runtime/lucet-runtime-tests/src/host.rs
```diff
@@ -269,7 +269,6 @@ macro_rules! host_tests {
             }
         }
 
-        #[test]
```
awortman-fastly: uh. i'll put this back.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -688,6 +692,51 @@ impl Instance {
...
+                panic!("Invalid state: Instance marked as in guest code while exiting a hostcall.");
+            }
+            Domain::Hostcall => {
+                *current_domain = Domain::Guest;
+            }
+            Domain::Terminated => {
+                // The instance was stopped in the hostcall we were executing.
+                std::mem::drop(current_domain);
+                unsafe {
+                    self.terminate(TerminationDetails::Remote);
```
acfoltzer: We need to remember to change this to a panic once the unwinding support is merged, as this currently will otherwise drop frames.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -709,6 +758,12 @@ impl Instance {
             module,
             ctx: Context::new(),
             state: State::Ready,
+            kill_state: Arc::new(KillState {
+                terminable: AtomicBool::new(false),
+                tid_change_notifier: Condvar::new(),
+                execution_domain: Mutex::new(Domain::Guest),
+                thread_id: Mutex::new(None),
+            }),
```
acfoltzer: Since there are no free variables here, I'd factor this out into something like `KillState::new_arc()`

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -825,6 +881,11 @@ impl Instance {
         );
         self.state = State::Running;
 
+        {
+            let mut kill_state = self.kill_state.thread_id.lock().unwrap();
+            *kill_state = Some(unsafe { pthread_self() });
```
acfoltzer: Could we rename the variable used on the left-hand side, or just do the lock and assignment in a single statement?

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -837,6 +898,22 @@ impl Instance {
         });
 
         self.with_signals_on(|i| {
+            // set up the guest to set itself as terminable, then continue to
+            // whatever guest code we want to run
+            //
+            // lucet_context_activate takes the guest code address in `rsi`,
+            // replacing the guest return address with itself so it can run
```
acfoltzer: replacing the guest return address where? and `itself` is unclear here
...
ah, the comments below make it more clear, but maybe this could be reworded still?

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1119,3 +1204,150 @@ pub(crate) struct EmptyYieldVal;
...
+///
+/// Some definitions for reference in this struct's documentation:
+/// * "stopped" means "stop executing at some point before reaching the end of the entrypoint
+/// wasm function".
+/// * "critical section" means what it typically means - an uninterruptable region of code. The
+/// detail here is that currently "critical section" and "hostcall" are interchangeable, but in
+/// the future this may change. Hostcalls may one day be able to opt out of criticalness, or
+/// perhaps guest code may include critical sections.
+///
+/// "Stopped" is a particularly loose word here because it ecompasses the worst case: trying to
```
acfoltzer: typo `ecompasses`

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1119,3 +1204,150 @@ pub(crate) struct EmptyYieldVal;
...
+/// stop a guest that is currently in a critical section. Because the signal will only be checked
+/// when exiting the critical section, the latency is bounded by whatever embedder guarantees are
+/// made. In fact, it is possible for a kill signal to be successfully sent and still never
+/// impactful, if a hostcall itself invokes `lucet_hostcall_terminate!`.
```
acfoltzer: I think this needs to be explained in more detail

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1119,3 +1204,150 @@ pub(crate) struct EmptyYieldVal;
...
+pub(crate) struct KillState {
```
acfoltzer: Can we move this struct and related code to a separate module? `instance.rs` is pretty huge already

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1119,3 +1204,150 @@ pub(crate) struct EmptyYieldVal;
...
+#[derive(Debug, PartialEq)]
+pub enum KillError {
+    NotTerminable,
+}
+
+type KillResult = Result<KillSuccess, KillError>;
+
+/// An object that can be used to terminate an instance's execution from a separate thread.
+impl KillSwitch {
+    /// Signal the instance associated with this `KillSwitch` to stop, is possible.
```
acfoltzer: typo `if`

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1119,3 +1204,150 @@ pub(crate) struct EmptyYieldVal;
...
+                        // Something else (another KillSwitch?) has already signalled this instance
+                        // to exit when it has completed its hostcall. Nothing to do here.
+                        Err(KillError::NotTerminable)
+                    }
+                };
+                // we must hold the lock on this bool at least until we set the "timed_out" flag.
+                mem::drop(execution_domain);
+                result
+            } else {
+                Err(KillError::NotTerminable)
```
acfoltzer: Could these be rearranged to use `ok_or`/`?` so they're next to the condition that failed?

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
acfoltzer: Let's do a `debug_assert!` for now, and remember to log it properly later when we have some sort of stats interface

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
acfoltzer: I think it'd help to enumerate the potential interleavings in documentation. It seems like we two variables to explore: currently handling signal raised by a `KillSwitch` or no? `KillSwitch` signal arrives before or after this `store`?

lucet-runtime/lucet-runtime-internals/src/vmctx.rs
acfoltzer: Will our demise ever arrive? This is only called from a hostcall, so the execution domain shouldn't allow a signal to be sent. I think we instead need to continue on to swap back to the host, but then check for `Domain::Terminated` on the host side.

lucet-wasi/src/main.rs
```diff
@@ -237,6 +248,17 @@ fn run(config: Config<'_>) {
             .build()
             .expect("instance can be created");
 
+        if let Some(timeout) = config.timeout {
+            let kill_switch = inst.kill_switch();
+            thread::spawn(move || {
+                thread::sleep(timeout);
+                // We may hit this line exactly when the guest exits, so sometimes `terminate` can
+                // fail. That's still acceptable, so just ignore the result.
+                #[allow(unused_must_use)]
```
acfoltzer: this attribute isn't preventing this warning, yet if I move it right above `thread::spawn` it works. must be some weird interaction between attributes and the grammar

lucet-runtime/lucet-runtime-internals/src/context/context_asm.S
```diff
@@ -66,9 +74,27 @@ _lucet_context_backstop:
...
+    // x86 xchg has an implicit "lock" semantic, so we don't need the prefix here.
+    xchgb (%rax), %cl
+    test %cl, %cl
+
+    // if we got 0 from the `xchg`, some terminator beat us in trying to
+    // stop this instance, and it's not safe to start swapping contexts.
+    //
+    // so, just hang out for a bit and wait for the terminator to do its thing
+_hang:
+    jz _hang
```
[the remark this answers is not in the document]
awortman-fastly: It is, but this might deserve some clarification as to _why_ - if the zero flag is set, `cl` is zero, and the attempt to acquire the flag failed. In that case, we're about to exit, but something else intends to terminate the guest, so we do spin lest we enter a critical section (context swap) or are outright in host code.

In the case we got the flag, `cl` will be `1`, the zero flag will be unset, and we fall right through the loop, since we acquired the termination flag and nothing else will try stopping in a subsequent critical section.

I was going to write out what this looks like less golf'd but it turns out to be clearer anyway:
```
; test if we actually acquired the flag
test %cl, %cl
jnz lucet_context_swap
; we didn't get the flag, so something else will be terminating this guest
_await_termination:
  jmp _await_termination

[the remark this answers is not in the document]
Comment (acfoltzer) [reply lucet-runtime/lucet-runtime-internals/src/context/context_asm.S]: Got it, that makes sense and the new version makes that a lot more evident

File: lucet-runtime/lucet-runtime-internals/src/instance.rs (hunk 13 above)
Comment (awortman-fastly) [reply lucet-runtime/lucet-runtime-internals/src/instance.rs]: Done! This was a great improvement.

File: lucet-runtime/lucet-runtime-internals/src/instance.rs (hunk 8 above)
Comment (awortman-fastly) [reply lucet-runtime/lucet-runtime-internals/src/instance.rs]: rephrased this comment, I think it's better but I'll leave that up to you

File: lucet-runtime/lucet-runtime-internals/src/instance.rs (hunk 10 above)
Comment (awortman-fastly) [reply lucet-runtime/lucet-runtime-internals/src/instance.rs]: Tacked on another sentence to expand on this case. Not sure if this gets too rambly, there's a surprising doc-to-code ratio here :D

File: lucet-runtime/lucet-runtime-internals/src/instance/execution.rs [hunk 16]
```diff
@@ -0,0 +1,318 @@
...
+                *execution_domain = Domain::Terminated;
+                Ok(KillSuccess::Pending)
+            }
+            Domain::Terminated => {
+                // Something else (another KillSwitch?) has already signalled this instance
+                // to exit when it has completed its hostcall. Nothing to do here.
+                Err(KillError::NotTerminable)
+            }
+        };
+        // we must hold the lock on this bool at least until we set the "timed_out" flag.
```
acfoltzer: This comment seems out of date?

awortman-fastly: oh gee, yes

lucet-wasi/src/main.rs
awortman-fastly: I've since learned the "cool" trick to `.ok()` a Result to ignore it - I think the warning changing is a result of building this on 1.35 vs nightly

awortman-fastly:
it's green! I'm pressing the button!

PR #153, opened 2019-04-25
Fix sigaltstack(2) usage in lucet-runtime

jedisct1:
`sigaltstack(2)`'s second argument is not exactly the "old" `stack_t` when registering a stack for signal handling.

Only the flags are supposed to be read, in order to know whether we currently are on the alternative stack, and/or if one has been set.

The regular stack is automatically restored after a signal handler returns.

Disabling an alternative stack should be done by passing the `SS_DISABLE` flag, not necessarily by copying what was returned in the optional second argument to the function.

Even on Linux, the address of the "old" `stack_t` appears to be always set to `NULL`, but it may be undefined. So it is not relevant, and is ignored when `SS_DISABLE` is set anyway.

On MacOS, providing the stack size is required, even with `SS_DISABLE`.

This commit changes a few things.

First, in instead of returning the "old" `stack_t`, our `sigaltstack(2)` wrapper now returns the current flags, as this is the only information actually being returned.

A new function to only get these flags is added, so we can avoid setting something that is already set.

A new constructor for the `SigStack` is also added, and sets the required values to disable the signal stack.

This fixes all previous Lucet's issues with signal handling on macOS.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
@@ -59,9 +59,16 @@ impl Instance {
             SigStackFlags::empty(),
             libc::SIGSTKSZ,
         );
-        let saved_sigstack =
-            unsafe { sigaltstack(&guest_sigstack).expect("saving sigaltstack succeeds") };
-
+        let previous_sigstack = unsafe { sigaltstack(Some(guest_sigstack)) }
+            .expect("enabling or changing the signal stack succeeds");
+        if let Some(previous_sigstack) = previous_sigstack {
+            assert!(
+                !previous_sigstack
+                    .flags()
+                    .contains(SigStackFlags::SS_ONSTACK),
+                "an instance was created with a signal stack"
+            );
+        }
         let mut ostate = LUCET_SIGNAL_STATE.lock().unwrap();
         if let Some(ref mut state) = *ostate {
             state.counter += 1;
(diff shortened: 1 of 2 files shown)


acfoltzer:
How does this affect the case where there is an existing alt stack installed when the Lucet alt stack is set up?

jedisct1:
The alt stack can still be changed as frequently as needed, as long as it's not active.

You're not supposed to change the alt stack while it's active.

I don't know what Linux does if it doesn't return an error when this is attempted, but it's documented as returning `EPERM` in that case. And as expected, this is what happens on macOS and BSD.

jedisct1:
In fact, it fails with `EPERM` on Linux as well (at least on an up-to-date Arch Linux): https://gist.github.com/jedisct1/8220833d3ec09020389447a93ac905ba

acfoltzer:
Sorry, I didn't mean that we'd be messing with the altstack while handlers are running. Rather that the following sequence of events happens:

1. The host application starts and installs its own altstack
2. The application runs a Lucet instance, which installs a signal handler and a Lucet altstack
3. The Lucet instance finishes
4. Lucet reinstalls the altstack that was installed in step 1.

This playground shows the idea: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7196075e154cfadb388fda6e2b5642ad

My read of this patch is that rather than reinstalling the altstack in step 4, it just gets disabled, so the altstack set in step 1 is lost.

The Lucet signal handler can potentially call previously-installed signal handlers if it detects a signal is not originating from Lucet code, however that handler ends up running on the Lucet altstack because we cannot switch altstacks in a signal handler.

jedisct1:
I see, thanks for clarifying! Maybe it could be worth adding a test for this.

The following changes make the intent clear, and still solve the altstack issues on macOS.

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -357,13 +377,35 @@ bitflags! {
     }
 }
 
-pub unsafe fn sigaltstack(ss: &SigStack) -> nix::Result<SigStack> {
-    let mut oldstack = std::mem::uninitialized::<libc::stack_t>();
-
+pub unsafe fn sigaltstack(new_sigstack: Option<SigStack>) -> nix::Result<Option<SigStack>> {
```
acfoltzer: We might want to revisit this interface if we decide to upstream to `nix`, but this does the job we need it to do in this project

PR #154, opened 2019-04-25
Update to rust-1.34.1

jedisct1:


pchickey:
`rustfmt` output changed so this PR will have to reformat everything as well :upside_down_face:

PR #156, opened 2019-04-25
Resetting the heap did not zero it on non-Linux systems

jedisct1:


lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -126,6 +126,8 @@ impl RegionInternal for MmapRegion {
             let heap_size = alloc.slot().limits.heap_address_space_size;
 
             unsafe {
+                #[cfg(not(target_os = "linux"))]
+                memset(heap, 0, heap_size);
```
pchickey: can you add a comment here just for documentation? "mprotect and madvise are sufficient to zero a page on linux, but not necessarily on all POSIX operating systems, and on Mac OS in particular"

jedisct1: Done!

PR #157, opened 2019-04-26
Hostcall macros and Vmctx rework

acfoltzer:
This PR includes a few things developed while updating Terrarium to use newer versions of Lucet.

## Termination and unwinding

We now use a combination of `panic!` and `catch_unwind` to ensure that any resources that were held at the time an instance is terminated (such as `Mutex`es or `File`s) are properly disposed of.

Note that so far, this only works for the case where we are in at most one hostcall context. If we have a stack with `guest -> host -> guest -> host`, the first `host` segment will not be properly unwound.

We believe this is a reasonable limitation at this point, but more work will be required to support unwinding across guest stacks.

### `lucet_hostcall_terminate!`

This replaces `Vmctx::terminate()`, and is implemented in terms of `panic!`. As before, it allows the user to return an `Any` value for the termination details. It must only be used within the scope of...

### `lucet_hostcalls!`

This is a macro for defining hostcalls in Rust. In particular, it uses `catch_unwind` to make sure that panics raised by `lucet_hostcall_terminate!` are properly caught, with the instance shutting down in a terminated state.

## `Vmctx` interface rework

Because so many resources are accessed through `Vmctx`, practical applications quickly ran into limitations with borrow-checking. In particular, holding a mutable reference to the heap meant that one could not hold even an immutable reference to an embedder context, which is a big problem if one needed to copy data from one to the other. Even if `Vmctx`'s resources were accessed via fields, allowing split borrows, we would still have similar borrowing issues if multiple embedder contexts are used.

This PR uses `RefCell`s to instead enforce borrowing rules dynamically. In practice, the innocuous situations described above will work fine, but if the borrowing rules get violated, the instance will terminate using the machinery described above.

The `Vmctx` `heap`, `globals`, and `get_embed_ctx` interfaces are now all subject to this dynamic borrowing. The latter is now a bit awkward to access from outside of a hostcall, as it returns an `Option<Result<Ref<T>, BorrowError>>`, so maybe we'll want more `lucet_runtime::Error` variants down the road.

## WASI stdio default

We're required to provide working file descriptors for stdio, but we now default to providing `/dev/null`, with the option to inherit stdio. When we have a chance to write a properly Rusty system for managing WASI capabilities, we should turn these into virtual handles that mimic `/dev/null`'s behavior, but currently we must have a real file descriptor there.

## The return of `--reserved-size` for `lucetc`

Overrides `--{min,max}-reserved-size` if set, making them equal.


tyler:
Amazing work.

Have you looked at how much overhead the hostcall wrapping adds at runtime to simple hostcalls?

jedisct1:
👍

lucet-runtime/lucet-runtime-internals/src/vmctx.rs
```diff
@@ -96,13 +99,24 @@ impl Vmctx {
...
             .try_borrow_mut()
             .unwrap_or_else(|_| panic!(TerminationDetails::BorrowError("heap_mut")));
         RefMut::map(r, |b| b.borrow_mut())
     }
 
+    unsafe fn reconstitute_heap_if_needed(&self) {
+        let inst = self.instance_mut();
+        if inst.heap_mut().len() != self.heap.borrow().len() {
+            let old_heap = self.heap.replace(Box::<[u8]>::from_raw(inst.heap_mut()));
+            Box::leak(old_heap);
```
awortman-fastly: I'd appreciate a note here that this must be `Box::leak`'d to avoid `Box` trying to drop the heap, since all we're trying to do here is update the length of the boxed slice

lucet-runtime/lucet-runtime-internals/src/hostcall_macros.rs
```diff
@@ -0,0 +1,92 @@
...
+/// It is important to use this macro for hostcalls, rather than exporting them directly, as it
+/// installs unwind protection that prevents panics from unwinding into the guest stack.
+///
+/// Since this is not yet a proc macro, the syntax is unfortunately fairly brittle. The functions it
+/// encloses must be of the form:
+///
+/// ```ignore
+/// #[$attr1]
+/// #[$attr2]
+/// ... // any number of attributes are supported; in most cases you will want `#[no_mangle]`
```
jedisct1: What would be a case where would *not* want `#[no_mangle]`?

acfoltzer: It's useful when we're defining hostcalls for a `MockModuleBuilder`, so hostcalls for different tests don't interfere with each other.

acfoltzer: Actually, these aren't technically hostcalls, but are the fake guest functions that go into the module

PR #160, opened 2019-04-26
Add another explicit page zeroing when MADV_DONTNEED is not enough

jedisct1:
Needless to say that clearing the whole heap has some performance implications :(

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
@@ -91,7 +91,8 @@ impl RegionInternal for MmapRegion {
 
         // clear and disable access to the heap, stack, globals, and sigstack
         for (ptr, len) in [
-            (slot.heap, slot.limits.heap_address_space_size),
+            // We don't ever shrink the heap, so we only need to zero up until the accessible size
+            (slot.heap, alloc.heap_accessible_size),
             (slot.stack, slot.limits.stack_size),
             (slot.globals, slot.limits.globals_size),
             (slot.sigstack, SIGSTKSZ),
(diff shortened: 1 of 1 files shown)


jedisct1:
Do we really need to clear the whole heap?

Can we have non-zeroed pages after `alloc.heap_accessible_size`?

acfoltzer:
We don't ever shrink the heap, so I think we only need to zero up until the accessible size.

You'll also likely need to add similar code in the implementation of `reset`.

I'd also appreciate if you could keep that commented `eprintln!` around; it's useful to have for debugging things like this.

jedisct1:
Thanks Adam. The commented out debugging statement has been re-added.

Explicit zeroing was already done in `reset()`.

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -100,6 +101,13 @@ impl RegionInternal for MmapRegion {
         {
             // eprintln!("setting none {:p}[{:x}]", *ptr, len);
             unsafe {
+                // MADV_DONTNEED is not guaranteed to clear pages on non-Linux systems
+                #[cfg(not(target_os = "linux"))]
+                {
+                    mprotect(*ptr, *len, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE)
+                        .expect("mprotect succeeds during drop");
+                    memset(*ptr, 0, *len);
```
acfoltzer: should we set `PROT_NONE` again after this `memset`?

awortman-fastly: that's unconditionally done a few lines down from here :)

acfoltzer: so it is, lol :+1:

PR #163, opened 2019-04-29
lucet-idl refactoring

pchickey:


pchickey:
Frank and I discussed the difficulties surrounding tagged unions. Since the WASI and Terrarium interfaces don't actually need them, I'm going to get rid of them in the language in a subsequent PR. It is tricky to support them properly in C, and probably impossible to do a zero-cost abstraction in Rust as well.

PR #164, opened 2019-04-29
Zero out instance magic more safely

acfoltzer:
This was previously segfaulting if the region got dropped after the `mem::replace` line. Our tests
previously always had the region remain in scope, which is why we didn't catch it till now


jedisct1:
Wouldn’t `ptr::write_volatile` be potentially faster?

acfoltzer:
Instead of which part? If we write instead of replacing, the `Instance` doesn't get dropped

PR #168, opened 2019-05-03
Ensure indirect calls with duplicate types check correctly

### LINKED ISSUE #158: Calling call_indirect with argument and with a function defined prior to type definitions will cause exception
I discovered a backtrace when calling `call_indirect` in a much more complicated program but simplified it down to the following reproducer:

```
(module $calli

  (table funcref (elem $fn0 $fn1))

  ;; Must have a function before the types to trigger the issue
  (func $early_fn (param $arg i32) (result i32) (i32.const 0))

  (type $fnT0 (func (param i32) (result i32)))
  (type $fnT1 (func (result i32)))

  (func $fn0 (param $arg i32) (result i32) (i32.const 0))
  (func $fn1 (result i32) (i32.const 0))

  (func $main0 (drop (call_indirect (type $fnT0) (i32.const 0) (i32.const 0))))
  (func $main1 (drop (call_indirect (type $fnT1) (i32.const 1))))

  ;; This causes a backtrace
  (export "_start0" (func $main0))

  ;; This does not cause a backtrace
  (export "_start1" (func $main1))

  ;; Default _start (for wasmtime)
  (export "_start" (func $main0))
)
```

Here is how the problem manifests:

```
$ /opt/lucet/bin/lucetc-wasi -o calli.so calli.wasm
$ RUST_BACKTRACE=1 /opt/lucet/bin/lucet-wasi --entrypoint _start0 calli.so
thread 'main' panicked at 'lucet-wasi runtime error: Runtime fault: FaultDetails { fatal: false, trapcode: Some(BadSignature), rip_addr: 139639523104562, rip_addr_details: Some(AddrDetails { in_module_code: true, file_name: Some("/mal/wasm/calli.so"), sym_name: Some("guest_func__start0") }) }', lucet-wasi/src/main.rs:196:23
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:70
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:58
             at src/libstd/panicking.rs:200
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:215
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:478
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:385
   6: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:340
   7: lucet_wasi::main
   8: std::rt::lang_start::{{closure}}
   9: std::panicking::try::do_call
             at src/libstd/rt.rs:49
             at src/libstd/panicking.rs:297
  10: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:92
  11: std::rt::lang_start_internal
             at src/libstd/panicking.rs:276
             at src/libstd/panic.rs:388
             at src/libstd/rt.rs:48
  12: main
  13: __libc_start_main
  14: _start
$ RUST_BACKTRACE=1 /opt/lucet/bin/lucet-wasi --entrypoint _start1 calli.so
$ wasmtime calli.wasm
```

Note that calling the `_start1` entrypoint doesn't not cause the problem only calling `_start0` will trigger the problem. Also, wasmtime is able to call the `_start1` entrypoint without issue.

The problem appears to be triggered by a combination of two things. The first is a `call_indirect` call that has an argument to use with the called function (`_start0`). Using call_indirect without the argument does not trigger the issue (`_start1`). The second necessary part to trigger this is that there must be a function (`early_fn`) declared prior to the definition of the two types. If that function is omitted or moved below the type definition then the problem does not reproduce.

awortman-fastly:
This PR smushes (technical word) duplicate types together behind a single `UniqueSignatureIndex`, then uses that to typecheck `call_indirect`, to comply with the wasm spec (and fix #158) on what constitutes a permissible indirect call.

We now pass both the `call_indirect` and `func` spectests! :tada:

### DIFF
--- a/lucetc/src/decls.rs
+++ b/lucetc/src/decls.rs
@@ -5,7 +5,7 @@ use crate::module::ModuleInfo;
 pub use crate::module::{Exportable, TableElems};
 use crate::name::Name;
 use crate::runtime::{Runtime, RuntimeFunc};
-use cranelift_codegen::entity::{EntityRef, PrimaryMap};
+use cranelift_codegen::entity::{entity_impl, EntityRef, PrimaryMap};
 use cranelift_codegen::ir;
 use cranelift_codegen::isa::TargetFrontendConfig;
 use cranelift_module::{Backend as ClifBackend, Linkage, Module as ClifModule};
@@ -20,11 +20,18 @@ use lucet_module_data::{
 };
 use std::collections::HashMap;
 
+/// UniqueSignatureIndex names a signature after collapsing duplicate signatures to a single
+/// identifier, whereas SignatureIndex is directly what the original module specifies, and may
+/// specify duplicates of types that are structurally equal.
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
+pub struct UniqueSignatureIndex(u32);
+entity_impl!(UniqueSignatureIndex);
+
 #[derive(Debug)]
 pub struct FunctionDecl<'a> {
     pub import_name: Option<(&'a str, &'a str)>,
     pub export_names: Vec<&'a str>,
-    pub signature_index: SignatureIndex,
+    pub signature_index: UniqueSignatureIndex,
     pub signature: &'a ir::Signature,
     pub name: Name,
 }
@@ -101,7 +108,8 @@ impl<'a> ModuleDecls<'a> {
         for ix in 0..info.functions.len() {
             let func_index = FuncIndex::new(ix);
             let exportable_sigix = info.functions.get(func_index).unwrap();
-            let signature = info.signatures.get(exportable_sigix.entity).unwrap();
+            let inner_sig_index = info.signature_mapping.get(exportable_sigix.entity).unwrap();
+            let signature = info.signatures.get(*inner_sig_index).unwrap();
             let name = if let Some((import_mod, import_field)) = info.imported_funcs.get(func_index)
             {
                 let import_symbol = bindings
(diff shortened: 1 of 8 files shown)


lucetc/src/decls.rs
```diff
@@ -347,12 +368,12 @@ impl<'a> ModuleDecls<'a> {
         }
     }
 
-    pub fn get_module_data(&self) -> ModuleData {
+    pub fn get_module_data(&self) -> Result<ModuleData, LucetcError> {
```
pchickey: Just curious why this was changed to be fallible. The current method body doesn't ever throw an error.

lucetc/src/module.rs
```diff
@@ -106,7 +111,22 @@ impl<'a> ModuleEnvironment<'a> for ModuleInfo<'a> {
...
         );
-        self.signatures.push(sig);
+
+        let mut maybe_match_key: Option<UniqueSignatureIndex> = None;
+
+        for (k, v) in self.signatures.iter() {
+            if v == &sig {
+                maybe_match_key = Some(k);
+            }
+        }
```
pchickey: I would use `find` for this, its short circuiting (stops iterating after a match) and shorter: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find

awortman-fastly: `find` is definitely better (and is what i wanted - i thought it wasn't there for some reason)

lucetc/src/decls.rs
awortman-fastly: ugh. shouldn't be, here. I had some conflict resolution issues with another branch (adding types to module data, where it's technically possible that cranelift declares a signature to us that isn't representable with wasm types)

kanaka:
I reported #158 and have built and verified that this branch addresses the problem I was seeing.

awortman-fastly:
@kanaka thank you for confirming this fixes your issue!

PR #174, opened 2019-05-08
Add function signatures to module data

awortman-fastly:
This PR adds function signatures to module data, as well as checking that `Instance::run` calls are made with matching arguments. This exposed a few instances in our tests where we were actually calling functions with incorrect arguments, but things happened to work out.

A small but noteworthy change is that the function manifest I was building in #136 unintentionally excluded imported functions. Those are included now, since they have signatures, can be called, and need to be recorded just the same.

### DIFF
--- a/benchmarks/lucet-benchmarks/src/modules.rs
+++ b/benchmarks/lucet-benchmarks/src/modules.rs
@@ -28,7 +28,7 @@ pub fn null_mock() -> Arc<dyn Module> {
 
     MockModuleBuilder::new()
         .with_export_func(MockExportBuilder::new(b"f",
-                                                 unsafe { *std::mem::transmute::<_, *const extern "C" fn()>(&f) })
+                                                 unsafe { std::mem::transmute::<_, extern "C" fn()>(f as u64) })
         )
         .build()
 }
@@ -53,7 +53,7 @@ pub fn large_dense_heap_mock(heap_kb: usize) -> Arc<dyn Module> {
     MockModuleBuilder::new()
         .with_export_func(
             MockExportBuilder::new(b"f",
-                                   unsafe { *std::mem::transmute::<_, *const extern "C" fn()>(&f) }
+                                   unsafe { std::mem::transmute::<_, extern "C" fn()>(f as u64) }
             )
         )
         .with_initial_heap(heap.as_slice())
@@ -88,7 +88,7 @@ pub fn large_sparse_heap_mock(heap_kb: usize, stride: usize) -> Arc<dyn Module> 
     MockModuleBuilder::new()
         .with_export_func(
             MockExportBuilder::new(b"f",
-                                   unsafe { *std::mem::transmute::<_, *const extern "C" fn()>(&f) }
+                                   unsafe { std::mem::transmute::<_, extern "C" fn()>(f as u64) }
             )
         )
         .with_initial_heap(heap.as_slice())
@@ -113,7 +113,7 @@ pub fn fib_mock() -> Arc<dyn Module> {
     MockModuleBuilder::new()
         .with_export_func(
             MockExportBuilder::new(b"f",
-                                   unsafe { *std::mem::transmute::<_, *const extern "C" fn()>(&f) }
+                                   unsafe { std::mem::transmute::<_, extern "C" fn()>(f as u64) }
             )
         )
         .build()
@@ -194,7 +194,7 @@ pub fn many_args_mock() -> Arc<dyn Module> {
     MockModuleBuilder::new()
         .with_export_func(
             MockExportBuilder::new(b"f",
-                                   unsafe { *std::mem::transmute::<_, *const extern "C" fn()>(&f) }
+                                   unsafe { std::mem::transmute::<_, extern "C" fn()>(f as u64) }
             )
         )
         .build()
@@ -273,12 +273,12 @@ pub fn hostcalls_mock() -> Arc<dyn Module> {
     MockModuleBuilder::new()
         .with_export_func(
             MockExportBuilder::new(b"wrapped",
-                                   unsafe { *std::mem::transmute::<_, *const extern "C" fn()>(&wrapped) }
+                                   unsafe { std::mem::transmute::<_, extern "C" fn()>(wrapped as u64) }
             )
         )
         .with_export_func(
             MockExportBuilder::new(b"raw",
-                                   unsafe { *std::mem::transmute::<_, *const extern "C" fn()>(&raw) }
+                                   unsafe { std::mem::transmute::<_, extern "C" fn()>(raw as u64) }
             )
         )
         .build()
--- a/benchmarks/lucet-benchmarks/src/modules.rs
+++ b/benchmarks/lucet-benchmarks/src/modules.rs
@@ -1,6 +1,7 @@
 use lucet_runtime::lucet_hostcalls;
 use lucet_runtime::vmctx::{lucet_vmctx, Vmctx};
 use lucet_runtime_internals::module::{HeapSpec, MockModuleBuilder, Module};
+use lucet_runtime_internals::module::Signature;
 use lucet_wasi_sdk::{CompileOpts, Lucetc};
 use lucetc::{Bindings, LucetcOpts, OptLevel};
 use std::path::Path;
@@ -26,7 +27,12 @@ pub fn null_mock() -> Arc<dyn Module> {
     extern "C" fn f(_vmctx: *mut lucet_vmctx) {}
 
     MockModuleBuilder::new()
-        .with_export_func(b"f", f as *const extern "C" fn())
+        .with_export_func(
+            b"f",
+            unsafe { std::slice::from_raw_parts(f as *const extern "C" fn() as *const u8, 1) },
+            &[],
+            Signature { params: vec![], ret_ty: None }
+        )
         .build()
 }
 
@@ -48,7 +54,12 @@ pub fn large_dense_heap_mock(heap_kb: usize) -> Arc<dyn Module> {
     });
 
     MockModuleBuilder::new()
-        .with_export_func(b"f", f as *const extern "C" fn())
+        .with_export_func(
+            b"f",
+            unsafe { std::slice::from_raw_parts(f as *const extern "C" fn() as *const u8, 1) },
+            &[],
+            Signature { params: vec![], ret_ty: None }
+        )
         .with_initial_heap(heap.as_slice())
         .with_heap_spec(heap_spec)
         .build()
@@ -79,7 +90,12 @@ pub fn large_sparse_heap_mock(heap_kb: usize, stride: usize) -> Arc<dyn Module> 
         });
 
     MockModuleBuilder::new()
-        .with_export_func(b"f", f as *const extern "C" fn())
+        .with_export_func(
+            b"f",
+            unsafe { std::slice::from_raw_parts(f as *const extern "C" fn() as *const u8, 1) },
+            &[],
+            Signature { params: vec![], ret_ty: None }
+        )
         .with_initial_heap(heap.as_slice())
         .with_heap_spec(heap_spec)
         .build()
@@ -100,7 +116,12 @@ pub fn fib_mock() -> Arc<dyn Module> {
     }
 
     MockModuleBuilder::new()
-        .with_export_func(b"f", f as *const extern "C" fn())
+        .with_export_func(
+            b"f",
+            unsafe { std::slice::from_raw_parts(f as *const extern "C" fn() as *const u8, 1) },
+            &[],
+            Signature { params: vec![], ret_ty: None }
+        )
         .build()
 }
 
@@ -177,7 +198,12 @@ pub fn many_args_mock() -> Arc<dyn Module> {
     }
 
     MockModuleBuilder::new()
-        .with_export_func(b"f", f as *const extern "C" fn())
+        .with_export_func(
+            b"f",
+            unsafe { std::slice::from_raw_parts(f as *const extern "C" fn() as *const u8, 1) },
+            &[],
+            Signature { params: vec![], ret_ty: None }
+        )
         .build()
 }
 
@@ -252,7 +278,17 @@ pub fn hostcalls_mock() -> Arc<dyn Module> {
     }
 
     MockModuleBuilder::new()
-        .with_export_func(b"wrapped", wrapped as *const extern "C" fn())
-        .with_export_func(b"raw", raw as *const extern "C" fn())
+        .with_export_func(
+            b"wrapped",
+            unsafe { std::slice::from_raw_parts(wrapped as *const extern "C" fn() as *const u8, 1) },
+            &[],
+            Signature { params: vec![], ret_ty: None }
+        )
+        .with_export_func(
+            b"raw",
+            unsafe { std::slice::from_raw_parts(raw as *const extern "C" fn() as *const u8, 1) },
+            &[],
+            Signature { params: vec![], ret_ty: None }
+        )
         .build()
 }
--- a/benchmarks/lucet-benchmarks/src/modules.rs
+++ b/benchmarks/lucet-benchmarks/src/modules.rs
@@ -1,7 +1,7 @@
 use lucet_runtime::lucet_hostcalls;
 use lucet_runtime::vmctx::{lucet_vmctx, Vmctx};
 use lucet_runtime_internals::module::{HeapSpec, MockModuleBuilder, Module};
-use lucet_runtime_internals::module::Signature;
+use lucet_runtime_tests::helpers::MockExportBuilder;
 use lucet_wasi_sdk::{CompileOpts, Lucetc};
 use lucetc::{Bindings, LucetcOpts, OptLevel};
 use std::path::Path;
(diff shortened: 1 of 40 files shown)


benchmarks/lucet-benchmarks/src/modules.rs
```diff
@@ -26,7 +27,12 @@ pub fn null_mock() -> Arc<dyn Module> {
     extern "C" fn f(_vmctx: *mut lucet_vmctx) {}
 
     MockModuleBuilder::new()
-        .with_export_func(b"f", f as *const extern "C" fn())
+        .with_export_func(
+            b"f",
+            unsafe { std::slice::from_raw_parts(f as *const extern "C" fn() as *const u8, 1) },
```
awortman-fastly: I added a [`function_bytes_slice`](https://github.com/fastly/lucet/pull/174/files#diff-af384e39937e7bd880a7bc1d0040fcfeR2) macro to make this a little more concise, but don't really feel great about making that available outside testing contexts. Very open to suggestions on a nice way for it to be available here, but not for, say, downstream use.

lucet-runtime/lucet-runtime-internals/src/module/mock.rs
```diff
@@ -99,8 +101,40 @@ impl MockModuleBuilder {
         self
     }
 
-    pub fn with_export_func(mut self, sym: &[u8], func: *const extern "C" fn()) -> Self {
-        self.export_funcs.insert(sym.to_vec(), func);
+    fn record_sig(&mut self, sig: Signature) -> UniqueSignatureIndex {
+        let idx = self
```
awortman-fastly: This might look familiar, which is because it's the same function as `declare_signature` in `ModuleInfo`. This greatly simplifies building up mock modules, instead of requiring us to track what signatures a mock module should have and their corresponding indices, so I think the copy is worthwhile.

lucetc/src/compiler.rs
```diff
@@ -139,13 +139,16 @@ impl<'a> Compiler<'a> {
         let function_manifest: Vec<(String, FunctionSpec)> = self
             .clif_module
             .declared_functions()
-            .filter_map(|f| {
-                f.compiled.as_ref().map(|compiled| {
```
awortman-fastly: oops. this is what rejected imported functions from the manifest.

lucetc/tests/wasi-sdk.rs
```diff
@@ -87,7 +87,7 @@ mod programs {
         let b = Bindings::empty();
         let h = HeapSettings::default();
         let c = Compiler::new(&m, OptLevel::Best, &b, h).expect("compile a");
-        let _mdata = c.module_data();
+        let _mdata = c.module_data().unwrap();
 
         /* FIXME: module data doesn't contain the information to check these properties:
         assert_eq!(p.import_functions().len(), 0, "import functions");
```
awortman-fastly: @pchickey can you think of a reason we'd discerning between import and export functions in a list of names/signatures/etc? if not, I'll uncomment relevant parts of the lucetc tests too

lucet-runtime/lucet-runtime-tests/src/strcmp.rs
```diff
@@ -39,20 +38,15 @@ macro_rules! strcmp_tests {
...
             heap[s2_ptr..s2_ptr + s2.len()].copy_from_slice(&s2);
 
             let res = c_int::from(
                 inst.run(
                     b"run_strcmp",
-                    &[
-                        Val::GuestPtr(s1_ptr as u32),
-                        Val::GuestPtr(s2_ptr as u32),
-                        Val::GuestPtr(res_ptr as u32),
-                    ],
```
pchickey: It took me a minute but this fixed a bug in the test suite! nice job :)

benchmarks/lucet-benchmarks/src/modules.rs
pchickey: We have a bunch of things in `lucet_runtime_internals` that are not for downstream consumption outside of testing, e.g. `MockModuleBuilder`, so I don't see harm in putting it in the internals module.

lucet-module-data/src/module_data.rs
```diff
@@ -52,6 +61,24 @@ impl<'a> ModuleData<'a> {
         &self.globals_spec
     }
 
+    // Function index here is a different index space than `get_func_from_idx`, which
+    // uses function index as an index into a table of function elements.
+    //
+    // This is an index of all functions in the module.
+    pub fn get_signature(&self, fn_id: u32) -> Option<&Signature> {
+        self.function_info.get(fn_id as usize).and_then(|func| {
+            self.signatures().get(func.signature.as_u32() as usize)
```
pchickey: if the function exists the signature will as well, right - otherwise its a malformed module data? i'd add a `expect` on the signatures().get() so that if this happens we detect it by crashing in tests, hopefully

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -565,6 +565,37 @@ impl Instance {
...
+        let fn_id: u32 = self
+            .module
+            .function_manifest()
+            .iter()
+            .enumerate()
+            .find(|(_, fn_spec)| fn_spec.relative_addr(func as u64) == Some(0))
+            .map(|(fn_id, _)| fn_id as u32)
+            .ok_or(Error::InvalidArgument(
+                "entrypoint function must be the start of a function declared in this module",
+            ))?;
```
pchickey: This is correct code, and I'm impressed by it, but rather than lookup into the module from the function pointer, lets instead have Module always return this metadata alongside the function pointer.

Can you have `Module::get_export_func`, `get_start_func`, and `get_func_from_idx` return structs that have both the function pointer, and a ref to the signature, inside it?

lucet-runtime/lucet-runtime-internals/src/module/mock.rs
pchickey: sgtm.

lucetc/tests/wasi-sdk.rs
pchickey: As future work, we do want some way to check the set of import functions and their signatures for a module, as a runtime safety feature that e.g. checks that all of the wasi imports match the types that the wasi runtime library implemented them as.  We will want to do the same in the compiler, but since deployment situations will doubtless be complicated, I would like to check them in both places since the check will be quick & easy on module load time.

The tests happened to be written this way against the old abstraction because its a reasonable way to check that a given wat's imports and exports match up against the compiled version. It might not end up making sense in the wasi-sdk test suite but it would be nice to try it. If it seems fragile or fiddly in this particular test, we can ignore it.

pchickey: Oh, we will also want signatures on import functions for when we implement per-instance dynamic imports.

pchickey:
wrt Module::module_data() being partial - I'd rather see a design that validates the module data as part of creating a Module, so that we don't have to unwrap that everywhere. If a module has an invalid manifest, we don't want to do anything with it, so it should fail as early as possible.

lucet-module-data/src/module_data.rs
awortman-fastly: :+1: good call

lucet-runtime/lucet-runtime-internals/src/instance.rs
awortman-fastly: I think having those functions return a pointer/signature pair ends up complicating downstream interfaces in a bad way.. I'm not as opposed to making all uses of internal functions take a struct like `FunctionHandle(*const extern "C" fn(), &Signature)` but that still obligates users to look up the signature when they might not need it, as well as becoming a change to `lucet_vmctx.h` to return structs or tuples or whatever (though that might be fine?)

Since signature lookup will end up happening by matching function pointers anyway[1], how's taking this chain and putting it behind a function like `Module::signature_for_func` sound? 

[1] Looking up by function pointers would be necessary to retrieve signatures for functions obtained by table index, where the space of `(table_idx, fn_idx)` doesn't very nicely to the space of functions with signatures in a Module. Not matching function pointers would mean tables have to have an index to the internal function space and adds more indirection for indirect calls :(

lucet-runtime/tests/globals.rs
```diff
@@ -1,3 +1,4 @@
+use lucet_runtime_tests::function_bytes_slice;
```
acfoltzer: multiple uses of the same module

lucet-runtime/lucet-runtime-internals/src/instance.rs
pchickey: I don't know of any uses of looking up a function that don't also require the signature eventually. And I'd ideally like to change Instance::run to take a function handle in substitute for the export name, and make users look up the function from a `Module::get_export_function` interface.

For the table lookup, we should add a module-data that maps table index to function index - it can just be a vector for now. We'll need to expand that representation when we implement table initializers that compute offsets from import globals. So, you can put the simple version of that in now.

I don't think its a problem to change the C API. We don't expect any of the runtime APIs to be stable as we roll out the changes required for spec compliance. We left out a lot of this complexity in order to have simpler interfaces that were easy to iterate on, but we're going to have to add some indirection.

pchickey: btw I don't think you need to change Instance::run and expose Module::get_export_function in this PR. but it may be a good thing to think about. It will require some way of checking that the export function passed to an instance belongs to the module of the instance. I'm not sure if you can use lifetimes to do that in the Rust type system or not, but for the C API itll require some sort of runtime check.

awortman-fastly: A mapping from table index to function index makes sense and cleans up that concern well.

I thought there was a cases where we would end up grabbing the signature, discard it, and look it up again to check, but on reread that's not actually a concern because we'd be passing the handle including signature along. So I'm totally on board :)

lucet-module-data/src/types.rs
```diff
@@ -0,0 +1,126 @@
...
+            _ => Err(ValueError::Unrepresentable)
+        }
+    }
+}
+
+/// A signature for a function in a wasm module.
+///
+/// Note that this does not explicitly name VMContext as a parameter! It is assumed that all wasm
+/// functions take VMContext as their first parameter.
+// VMContext is not explicitly named specifically so that value types can be tagged with eight bits
```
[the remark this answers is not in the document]
awortman-fastly: it is - the distinction I was making is that rustdoc comments are "what/why" about the structure as a user might care to see, where the non-rustdoc comment is context for someone looking at this structure who might change it. that said, the comment isn't even meaningful anymore :D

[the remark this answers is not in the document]
acfoltzer: gotcha, sounds good. on this project I've tended to make the rustdoc include the extra context for the crates that aren't meant to be public facing, but I'm also a weirdo who runs `cargo doc --document-private-items` regularly

lucet-runtime/lucet-runtime-tests/src/host.rs
```diff
@@ -8,6 +8,7 @@ macro_rules! host_tests {
             lucet_hostcall_terminate, lucet_hostcalls, DlModule, Error, Limits, Region,
             TerminationDetails, TrapCode,
         };
+        use lucet_runtime_internals::{module::Signature, val::Val};
```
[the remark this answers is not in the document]
awortman-fastly: sounds good, though I've addressed this by just removing the need for either of these imports. incidentally, `Signature` was never actually used in this macro.. do warnings inside macro expansions not get reported?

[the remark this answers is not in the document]
acfoltzer: they don't seem to in this case; I think it might be because of the cross-crate expansion?

lucet-runtime/lucet-runtime-tests/src/entrypoint.rs
```diff
@@ -119,14 +120,99 @@ pub fn mock_calculator_module() -> Arc<dyn Module> {
...
+                params: [ValueType::F64; 10].to_vec(),
+                ret_ty: Some(ValueType::F64),
+            },
+        )
+        .with_export_func(
+            b"add_mixed_20",
+            function_bytes_slice!(add_mixed_20),
+            &[],
+            Signature {
+                params: vec![
```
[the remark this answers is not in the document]
awortman-fastly: a macro like this would be nice but I think a bit more time than I want to put in for the handful of uses at the moment, unless you know a nice way to expand `F64` and friends to `ValType::F64`? I adjusted `with_export_func` to take a builder that fills in defaults that makes this a little less tedious to work with, so that might help a bit

lucet-module-data/src/functions.rs
```diff
@@ -1,7 +1,44 @@
...
+///
+/// This is split from but closely related to a [`FunctionSpec`]. The distinction is largely for
+/// serialization/deserialization simplicity, as [`FunctionSpec`] contains fields that need
+/// cooperation from a loader, with manual layout and serialization as a result.
+/// [`FunctionMetadata`] is the remainder of fields that can be automatically
+/// serialized/deserialied and are small enough copying isn't a large concern.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct FunctionMetadata<'a> {
+    pub signature: UniqueSignatureIndex,
+    pub name: Option<&'a str>,
```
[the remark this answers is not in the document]
awortman-fastly: I don't *think* so, I would have expected a derivation error if serde didn't know to borrow this from the input, but the docs only mention `&str` and `&[u8]`, not compositions on top of them. know of a nicer way to check for this being zero-copy than reading a derived `Deserialize`'s assembly?

lucet-runtime/lucet-runtime-tests/src/entrypoint.rs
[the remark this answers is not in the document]
acfoltzer: The components of a path like that are separate tokens, so you should be able to use the ident matcher. Something like

```rust
( ( $($arg_ty:ident),* ) ) => { vec![ $($crate::ValType::$arg_ty),* ] }
```

lucet-module-data/src/functions.rs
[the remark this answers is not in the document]
acfoltzer: I don't, but perhaps @pchickey picked up some tricks when he did the initial implementation?

lucet-runtime/lucet-runtime-tests/src/entrypoint.rs
```diff
@@ -728,8 +852,11 @@ macro_rules! entrypoint_tests {
                 let func = vmctx
                     .get_func_from_idx(0, cb_idx)
                     .expect("can get function by index");
-                let func = func as *const extern "C" fn(*mut lucet_vmctx, u64) -> u64;
```
awortman-fastly: This specific test was very handy because in switching to `FunctionHandle`, which has an `extern "C" fn()` rather than a `*const extern "C" fn()`. In following the types literally, I tried doing something like `&func.ptr as *const extern "C" fn...` and ran into the segfaults I mentioned earlier, with a pointer that had one too many indirections.

It turns out that most of our `*const extern "C" fn()` are _really_ `extern "C" fn()` -- they're function pointers, not pointers to function pointers. Because it's *much* nicer syntactically to work with (you can `as` them without issue), I'm actually not sure if that's intentional... it results in highly confusing errors when following the types literally though.

The last two commits push through a change to turn `*const extern "C" fn()` into `extern "C" fn()` where that's the actual pointer type of the underlying data, but I'm not sold on them because of how much it tosses transmutes across the codebase. Since fn-to-different-fn isn't a primitive cast, `as` is not legal, and transmute is what I end up having to use :(

lucet-runtime/src/c_api.rs
```diff
@@ -420,9 +420,7 @@ lucet_hostcalls! {
         vmctx.instance()
             .module()
             .get_func_from_idx(table_idx, func_idx)
-            // the Rust API actually returns a pointer to a function pointer, so we want to dereference
```
awortman-fastly: I didn't see any uses of this function in `lucet` but I think this would have resulted in the same incorrect double-deref of `fptr` for users of the C api?

lucet-runtime/lucet-runtime-tests/src/entrypoint.rs
acfoltzer: Hopefully a future PR will replace this with an `Instance::run_from_hostcall()` function that will both obviate the need for the user to care about `FunctionHandles` and typecheck the function rather than using `transmute`.

lucet-runtime/lucet-runtime-tests/src/host.rs
```diff
@@ -119,7 +119,7 @@ macro_rules! host_tests {
             let mut inst = region
                 .new_instance(module)
                 .expect("instance can be created");
-            inst.run(b"main", &[Val::GuestPtr(0), Val::I32(0)])
+            inst.run(b"main", &[0u32.into(), 0i32.into()])
```
acfoltzer: this actually changes the first argument from `Val::GuestPtr(0)` to `Val::U32(0)`; was this intentional?

lucet-runtime/src/c_api.rs
acfoltzer: a good reminder that we need a better test suite for the C API

PR #176, opened 2019-05-09
Implement the remaining hostcalls

jedisct1:
The large `hostcalls.rs` file becomes a module, split in multiple files.

The hostcalls implementations themselves are not in the macro any more. What is inside the macro is not valid Rust code, RLS cannot do anything with it, neither can `rustfmt`, so maintaining that code was tedious.

So, implementations are now outside the macro, with the content of the macro being nothing but thin wrappers.

(draft: not tested on Linux yet)
(update: now compatible with Linux)


acfoltzer:
I think we should do this refactoring either way, but if removing the `&mut vmctx` syntax from the `lucet_hostcalls!` macro would make rustfmt work, then let's do that as well.

jedisct1:
Unfortunately, removing the `&mut vmctx` syntax to make it valid Rust doesn't help much. 
Everything in the macro is seen as an unstructured blob of text.

But the wrappers are fine. It's actually very convenient to have all of the public definitions in a single file, while having some flexibility for the actual implementations.

acfoltzer:
Gotcha, that's what I thought, which is why I wasn't worrying too much about messing with the syntax. It definitely is a major downside of using this kind of macro. Eventually it'd be nice to make it a proc macro so that we can use attribute syntax instead.

I totally agree about the separation.

jedisct1:
Procedural macros are going to make a lot of things way easier and cleaner.

Looking forward to them being stabilized some day...

acfoltzer:
They are stable, or at least enough of them that we could do this. I just didn't have the time to dig in and do it properly when implementing the macro the first time around.

lucet-wasi/src/hostcalls/fs.rs
```diff
@@ -0,0 +1,1338 @@
...
+        Err(e) => return enc_errno(e),
+    };
+    let path_cstr = match std::ffi::CString::new(path.as_os_str().as_bytes()) {
+        Ok(path_cstr) => path_cstr,
+        Err(_) => return wasm32::__WASI_EINVAL,
+    };
+    // nix doesn't expose mkdirat() yet
+    match unsafe { mkdirat(dir, path_cstr.as_ptr(), 0o777) } {
+        0 => wasm32::__WASI_ESUCCESS,
+        _ => wasm32::errno_from_nix(errno::Errno::last()),
```
sunfishcode: It looks like `last()` translates into `ENOSYS`, so it'd be a little more obvious here if this (and other places that do this) just said `wasm32::errno_from_nix(errno::Errnor::ENOSYS)` directly.

lucet-wasi/src/memory.rs
```diff
@@ -397,6 +419,14 @@ dec_enc_scalar!(
     enc_timestamp_byref
 );
 
+pub fn dec_u32(x: u32) -> u32 {
+    u32::from_le(x)
+}
+
+pub fn enc_u32(x: u32) -> u32 {
+    x
```
sunfishcode: It's surprising that `enc_u32` isn't the inverse of `dec_u32` here. I don't use a use of `dec_u32` in the patch though.

On a tangential note, I'm starting to think that the "wasm32.rs" vs "host.rs" split (that I'm partly to blame for) is a little confusing, especially with all the `enc`/`dec`/ functions for scalar types that are effectively no-ops. I think we can unify things when they're defined to be fixed-size types that are the same on both sides. But that's a bigger topic that we can figure out later :-).

sunfishcode:
I've only made a brief skim through the code here, but it looks good to me as a base to build on!

lucet-wasi/src/hostcalls/fs.rs
jedisct1: Does it? `nix::errno::Errno::last()` is documented as being the last `errno` value, and `mkdirat()` can return many more error codes than `ENOSYS`.

lucet-wasi/src/memory.rs
jedisct1: I guess this is due to the fact that WASM is always assumed to be little-endian.

jedisct1: Actually, the `dec_enc_scalar()` properly converts to little-endian, so I guess not doing it in `enc_u32()` is an oversight. I'm gonna add the missing `.to_le()`.

jedisct1:
Thanks a ton for your review, Dan!

lucet-wasi/src/hostcalls/fs.rs
sunfishcode: I mean that `errno_from_nix` turns `nix::errno::Errno::last()` into `wasm32::__ENOSYS`, so `wasm32::errno_from_nix(errno::Errnor::ENOSYS)` should produce the same result.

lucet-wasi/src/memory.rs
sunfishcode: Wasm is little-endian, but the question is where the endian translation happens. It's surprising to have functions named `dec_u32` and `enc_u32` where the former does an endian translation going one way, and the later does not do an endian translation going the other way.

jedisct1: Done :)

lucet-wasi/src/hostcalls/fs.rs
jedisct1: Hey Dan.

I'm still a little bit confused by your comment.

[`nix::errno::Errno::last()`](https://github.com/nix-rust/nix/blob/63aec5066b6bca637f45a4acc23aa8f8079ca050/src/errno.rs#L132-L134) returns the current value of libc's `errno`.

And `mkdirat()` can fail with `EEXIST` that `wasm32::errno_from_nix()` turns into `__WASI_EEXIST`.  I added a trivial [test for this](https://github.com/fastly/lucet/blob/4321e538acd868e9303a9576a5138b64b43542ef/lucet-wasi/tests/guests/fs.c#L88) and this doesn't seem to return `ENOSYS`, but `EEXIST` as expected.

Can you clarify when `errno_from_nix` turns `nix::errno::Errno::last()` into `wasm32::__ENOSYS`?

sunfishcode: Ah, I somehow missed that `nix::errno::Errno::last()` returns the current `errno` value. You're right, and the code looks good!

jedisct1: 👍

PR #177, opened 2019-05-14
Add function import/export information to module data

awortman-fastly:
also uncomment parts of lucetc tests that are now expressible/testable

Adds FunctionIndex newtype instead of passing around raw u32, to help make the lifecycle of function indices clearer.

Overall this isn't too big, but is the last piece I wanted to have in module data before renaming, and spending a moment fixing up lucet-analyze to know about all the new stuff.

### DIFF
--- a/lucet-module-data/src/module_data.rs
+++ b/lucet-module-data/src/module_data.rs
@@ -1,5 +1,5 @@
 use crate::{
-    functions::{FunctionMetadata, OwnedFunctionMetadata},
+    functions::{FunctionIndex, FunctionMetadata, ImportFunction, OwnedFunctionMetadata},
     globals::GlobalSpec,
     linear_memory::{HeapSpec, LinearMemorySpec, SparseData},
     types::Signature,
@@ -23,6 +23,9 @@ pub struct ModuleData<'a> {
     globals_spec: Vec<GlobalSpec<'a>>,
     #[serde(borrow)]
     function_info: Vec<FunctionMetadata<'a>>,
+    #[serde(borrow)]
+    import_functions: Vec<ImportFunction<'a>>,
+    export_functions: Vec<FunctionIndex>,
     signatures: Vec<Signature>,
 }
 
@@ -31,12 +34,16 @@ impl<'a> ModuleData<'a> {
         linear_memory: Option<LinearMemorySpec<'a>>,
         globals_spec: Vec<GlobalSpec<'a>>,
         function_info: Vec<FunctionMetadata<'a>>,
+        import_functions: Vec<ImportFunction<'a>>,
+        export_functions: Vec<FunctionIndex>,
         signatures: Vec<Signature>,
     ) -> Self {
         Self {
             linear_memory,
             globals_spec,
             function_info,
+            import_functions,
+            export_functions,
             signatures,
         }
     }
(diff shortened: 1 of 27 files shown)


lucet-module-data/src/module_data.rs
```diff
@@ -23,6 +23,9 @@ pub struct ModuleData<'a> {
     globals_spec: Vec<GlobalSpec<'a>>,
     #[serde(borrow)]
     function_info: Vec<FunctionMetadata<'a>>,
+    #[serde(borrow)]
+    import_functions: Vec<ImportFunction<'a>>,
```
awortman-fastly: I have these also as `Vec` because I expect the most common uses of these to be enumerating all imports or all exports of a module when doing setup, and probably not caring about import/export info after that point. If y'all can think of a point where we need to performantly look up if a given function is imported or exported, this probably isn't a great representation

lucetc/tests/wasm.rs
```diff
@@ -79,18 +71,22 @@ mod module_data {
...
-        assert_eq!(p.defined_functions().len(), 4);
+        assert_eq!(mdata.import_functions().len(), 1);
+        assert_eq!(mdata.import_functions()[0].module, "env");
+        assert_eq!(
+            mdata.sym_for(mdata.import_functions()[0].fn_idx),
+            Some(&b"icalltarget"[..])
+        );
+        assert_eq!(mdata.function_info().len(), 5);
         assert_eq!(
-            p.defined_functions().get(0).unwrap().symbol(),
```
awortman-fastly: This goes from `0` to `1` because `function_info` also contains entries for imported functions. I imagine dynamic linking-style function import would look like walking imports, looking for those indices, and updating the corresponding `FunctionSpec` with a pointer that would otherwise have had a relocation anyway.

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -165,15 +167,12 @@ impl ModuleInternal for DlModule {
         Ok(unsafe { from_raw_parts(*p_table_segment, **p_table_segment_len as usize) })
     }
 
-    fn get_export_func(&self, sym: &[u8]) -> Result<FunctionHandle, Error> {
-        let mut guest_sym: Vec<u8> = b"guest_func_".to_vec();
-        guest_sym.extend_from_slice(sym);
-
+    fn get_export_func(&self, sym: &str) -> Result<FunctionHandle, Error> {
```
awortman-fastly: This became another one of those "oh, I'll start changing more stuff while I'm here!" situations.

@pchickey since we talked earlier and realized `function_id_by_name` is actually too broad for this use, I changed it to `get_export_func_id`, which more closely matches the primary user, `get_export_func`, anyway. Because it goes through export names, it compares functions without the `guest_func_` prefix. Because export names are valid utf8 by [The Spec](https://webassembly.github.io/spec/core/syntax/values.html#syntax-name) this can take `&str` instead of a `&[u8]`. That simplifies things and avoids questions about "what if sym is a byte slice that is not utf8", which should never happen anyway.

acfoltzer: Ooo, that's good to know! Rust makes bytestrings easy enough to write but I was always annoyed by that type showing up in the region traits

lucetc/src/decls.rs
```diff
@@ -125,12 +145,16 @@ impl<'a> ModuleDecls<'a> {
                     let funcid = clif_module
                         .declare_function(&export_symbol, Linkage::Export, signature)
                         .context(LucetcErrorKind::TranslatingModule)?;
+                    exports.push(ExportFunction {
+                        fn_idx: LucetFunctionIndex::from_u32(function_names.len() as u32),
+                        names: exportable_sigix.export_names.clone(),
```
awortman-fastly: All the export names for a function get bunched together, which makes this really straightforward!

awortman-fastly:
@pchickey I still need to think about reexports but I've since tweaked this to address your comments.

Another open question: is there a reason the `&[u8]` -> `&str` should not also change `Instance::run` to take `entrypoint: &str`? Any `&[u8]` that is not a valid `&str` will be invalid wasm, so unless someone is doing something really funky I think the only concern is that the C api would have an additional failure mode that the rust api wouldn't.

acfoltzer-fastly:
I say go for it. We can just return the symbol not found error enum if the C string fails to decode

lucetc/src/decls.rs
pchickey: I think the way the imports and exports collection live in different branches of this name calculation means we can't support an export of an imported function. Can you move those out to the higher level (check for them at the same indentation level as `let name =`, if that makes sense), and also add a test case somewhere that shows we can run an export of an imported function?

PR #185, opened 2019-05-31
Fix a path_get() bug with nested directories

jedisct1:
When traversing directories, we should keep the most recent file descriptor.

Add a test case originally triggering that bug.

Spotted by @kubkon -- thanks!


jedisct1:
Yes, capabilities are hard :(

Differential testing is badly needed. All WASI implementations must behave the same, and it's probably far from being the case right now.

PR #189, opened 2019-05-31
Update lucet-analyze

awortman-fastly:
Now things look like: lucet-analyze screenshot

Important points:
* import/export information can come from module data!
* signatures!
* trap records!
* (some) consistency checks!

### DIFF
--- a/lucet-analyze/src/main.rs
+++ b/lucet-analyze/src/main.rs
@@ -1,3 +1,5 @@
+use lucet_module_data::{Error, FunctionSpec, ModuleData, TrapManifest, TrapSite};
+
 use byteorder::{LittleEndian, ReadBytesExt};
 use colored::Colorize;
 use goblin::{elf, Object};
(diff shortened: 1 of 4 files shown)


jedisct1:
This is rad!

awortman-fastly:
Worth noting: the errors about sizes not matching is actually due to a lucetc bug I found while updating lucet-analyze. Not good, but lucet-analyze working as intended :)

lucet-analyze/src/main.rs
```diff
@@ -459,57 +321,263 @@ fn print_summary(summary: ArtifactSummary) {
...
     }
 
     println!("");
-    println!("Trap Manifest:");
-    if let Some(trap_manifest) = summary.trap_manifest {
-        for row in trap_manifest.records {
-            println!("  {:25} {} traps", row.func_name, row.trap_count);
-            println!("      {:?}", row.sites);
+    println!("Data Segments:");
+    if let Some(data_segments) = summary.data_segments {
```
awortman-fastly: important question: are data segments still a thing? I didn't see where in `lucetc` data under these symbols would be emitted, and it _seems_ like this turned into sparse data. Is that a correct understanding? (cc @tyler ?)

PR #190, opened 2019-05-31
Emit function information for lucetc-provided functions

awortman-fastly:
Runtime functions were not getting their supporting metadata (trap tables, names, ..) added to module data, which I think is mostly harmless, but violates some assumptions I'd made about what data is present in module data and function manifests. Specifically, that all functions in the module are present, and have a signature.

### DIFF
--- a/lucetc/src/decls.rs
+++ b/lucetc/src/decls.rs
@@ -59,9 +59,9 @@ pub struct TableDecl<'a> {
 }
 
 pub struct ModuleDecls<'a> {
-    info: ModuleInfo<'a>,
+    pub info: ModuleInfo<'a>,
     runtime: Runtime,
-    function_names: PrimaryMap<FuncIndex, Name>,
+    pub function_names: PrimaryMap<FuncIndex, Name>,
     imports: Vec<ImportFunction<'a>>,
     exports: Vec<ExportFunction<'a>>,
     table_names: PrimaryMap<TableIndex, (Name, Name)>,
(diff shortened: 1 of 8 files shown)


lucetc/src/decls.rs
```diff
@@ -72,15 +72,17 @@ pub struct ModuleDecls<'a> {
...
         runtime: Runtime,
         heap_settings: HeapSettings,
     ) -> Result<Self, LucetcError> {
-        let (function_names, imports, exports) = Self::declare_funcs(&info, clif_module, bindings)?;
+        let (mut function_names, imports, exports) =
+            Self::declare_funcs(&info, clif_module, bindings)?;
         let table_names = Self::declare_tables(&info, clif_module)?;
-        let runtime_names = Self::declare_runtime(&runtime, clif_module)?;
+        let runtime_names =
+            Self::declare_runtime(&mut info, &mut function_names, &runtime, clif_module)?;
```
awortman-fastly: Because we're tracking more information in `ModuleData` (built out of this and `ModuleInfo`), the amount of bookkeeping as functions are introduced has grown a bit. I'm thinking about factoring out most of it into one or two functions on `ModuleDecls` to set up signatures, names, and indices. Then that can be reused by the typical `ModuleEnvironment` path of function declaration, or for internal declarations like runtime functions or the stack probe.

awortman-fastly: did this

PR #191, opened 2019-05-31
Fix benchmark entrypoint function signature

awortman-fastly:
well. _add_ an entrypoint function signature. The code compiles but when running `lucet-benchmark`, specifically "run_many_args", it fails because the mock module has the default signature of `() -> ()` where we provide ~66 arguments.


awortman-fastly:
cc @acfoltzer I broke and then fixed your code, sorry! :sweat_smile:

PR #196, opened 2019-06-05
Export/import spectest fixes

awortman-fastly:
This actually includes changes for lucetc tests as well, which it turned out to be testing that a bug continued being a bug

Broadly this improves support for exported values - globals and functions (haven't gotten to tables/memory yet)

Draft PR because it's not complete (latest commit fails due to not handling exported imports properly with Faerie), but there's enough here I'd like to get eyes

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/module/dl.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/module/dl.rs
@@ -80,9 +80,9 @@ impl DlModule {
                     let manifest_len = len_ptr.as_ref().ok_or(lucet_incorrect_module!(
                         "`lucet_function_manifest_len` is defined but null"
                     ))?;
-                    let manifest: *mut FunctionSpec = *ptr; /*.ok_or(lucet_incorrect_module!(
-                                                                "`lucet_function_manifest` is defined but null"
-                                                            ))?;*/
+                    let manifest: *mut FunctionSpec = ptr.as_mut().ok_or(
+                        lucet_incorrect_module!("`lucet_function_manifest` is defined but null"),
+                    )?;
 
                     std::slice::from_raw_parts_mut(manifest, *manifest_len as usize)
                 }
@@ -117,7 +117,7 @@ impl DlModule {
             }
         };
 
-        // resolve imports (oh no ...)
+        // Fix up entries in function manifest for imported functions to have correct addresses
         unsafe {
             use libloading::os::unix;
             use nix::sys::mman::ProtFlags;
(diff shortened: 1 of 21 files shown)


lucetc/src/decls.rs
```diff
@@ -285,7 +320,7 @@ impl<'a> ModuleDecls<'a> {
                     def: GlobalDef::new(init_val),
                 }
             };
-            globals.push(GlobalSpec::new(g_variant, None));
```
awortman-fastly: This is the big change for global exports - lucetc was just not passing export names along for globals. Because they can be more than one name (one global exported under multiple names), I switched it to a Vec as well

lucet-spectest/src/bindings.rs
```diff
@@ -1,6 +1,29 @@
 use lucetc::Bindings;
 use serde_json::json;
 
+use lucet_runtime::lucet_hostcalls;
+
+lucet_hostcalls! {
+    #[no_mangle]
+    pub unsafe extern "C" fn print(&mut _vmctx,) -> () {
```
awortman-fastly: These and import/export declaration changes below fix the `imports` spectest!

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -117,6 +117,36 @@ impl DlModule {
             }
         };
 
+        // resolve imports (oh no ...)
```
awortman-fastly: This is the really questionable change. Since looking up exported functions is done by walking `function_manifest`, where an imported function has address `0` (no relocation to place there, so it's just left default), we need to fix this up at module instantiation time with the right hostcall addresses.

The other option I can see is inserting stub functions that just `jmp [import_addr]` that we can actually export, which trades off instantiation time for the extra indirection at runtime. I'm just not sure how that tradeoff plays out. If we don't declare the import in artifact headers anymore, and fix up addresses here, we're just avoiding work the loader would have done, so the net instantiation time _should_ be the same.

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -117,6 +117,36 @@ impl DlModule {
...
+
+            for import in module_data.import_functions() {
+                let ptr = runtime.get::<*const usize>(import.name.as_bytes()).map_err(Error::DlError)?;
+                let fn_addr = ptr.as_ref().unwrap() as *const usize as u64;
+                function_manifest.get_mut(import.fn_idx.as_u32() as usize).unwrap().code_addr = fn_addr;
+            }
+
+            // Module data ends up in the same page as code, potentially, so this gets +x so as to
+            // permit those functions to execute. this should be fixed, though..
+            mprotect(manifest_page_ptr, manifest_len,
```
awortman-fastly: last one for now, this `mprotect` requires `PROT_EXEC`  because the function manifest can end up in the same page as a function in the module. that's not great and I want to fix regardless of if this `fix up function manifest at load time` logic stays

lucetc/src/decls.rs
```diff
@@ -267,25 +307,42 @@ impl<'a> ModuleDecls<'a> {
...
-            let g_import = info.imported_globals.get(ix);
-            let g_variant = if let Some((module, field)) = g_import {
-                GlobalVariant::Import { module, field }
-            } else {
-                let init_val = match g_decl.entity.initializer {
-                    // Need to fix global spec in ModuleData and the runtime to support more:
-                    GlobalInit::I32Const(i) => i as i64,
-                    GlobalInit::I64Const(i) => i,
-                    _ => Err(format_err!(
-                        "non-integer global initializer: {:?}",
```
awortman-fastly: Supporting float globals gets us passing the `global` spectest, though this PR doesn't add import global support into `lucet-runtime` (yet?), which is the entire reason I started changing globals here! :grimacing:

awortman-fastly:
Closing to split these changes into their logically different parts

PR #201, opened 2019-06-07
Add support for small(er) Docker images to use (not work on) Lucet

jedisct1:
This adds the ability to build a new Docker image, `lucet-toolchain`, that only contains the Lucet toolchain, and not the boilerplate to build it.

Lucet tools can then be directly used as:

```sh
docker run --rm lucet-toolchain lucet-wasi ...
```

This image is based on the current Ubuntu, and uses the stock LLVM installation, so it can be used as a base image for projects willing to build upon Lucet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,16 @@
 dist: xenial
 
 env:
-    - DEVENV_NO_INSTALL=1
+    - UNOPTIMIZED_BUILD=true
 
 services:
     - docker
 
 script:
     - ./devenv_run.sh make indent-check test audit
+    - ./devenv_stop.sh
     - git diff --exit-code
+    - ./devenv_build_toolchain_only.sh
 
 notifications:
     email: false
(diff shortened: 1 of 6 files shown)


pchickey:
I like this, but it doesn't look like it gets used from CI at all, so we'd be depending on user reports if it bit rots and breaks. Is there a sensible way to test it with travis?

jedisct1:
I can add this to Travis, but as this is going to make the builds take more time to complete, we will get closer to the point where Travis will explode.

But let's try :)

pchickey:
Ok. If this is the straw that breaks Travis's back, we can merge it without CI. Thanks for giving it a try. Hopefully we wont be on Travis for too much longer.

jedisct1:
Some changes have been added to the Travis builds. We now build the `lucet` container (not just `lucet-dev`) as well as the `lucet-toolchain` one.

Just like in non-CI mode, except that what gets installed is the binaries compiled in debug mode. When the same scripts are run outside CI, they are still compiled in release mode.

PR #203, opened 2019-06-08
Improve import/export function support

awortman-fastly:
Split from #196 into changes I'd be comfortable squashing together:

This is a pair of changes (and some tests around what they change) that get us closer to spec compliance and plug a surprising failure mode:
* Exported imports do not get declared as an export function in the produced artifact
* Imported function addresses in a module's function manifest are fixed at load-time to the appropriate function pointers from the host

The first piece is necessary to support exported import functions at all - otherwise they result in a module with an export and no code defined for it, and an error generating artifacts.

Once we _can_ generate artifacts, the `FunctionHandle` for an exported import has a ptr of `0`, the default. The change to update imported function addresses in the manifest at load time fixes that. The result is we can correctly load and execute modules with an exported import as the entrypoint!


awortman-fastly:
ah. I've been testing with `cargo test` locally, which appears to not run tests for `lucet-wasi`, and I have thoroughly broken it. hm.

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -117,6 +118,94 @@ impl DlModule {
...
+                    .unwrap();
+
+                // TODO: if this import is ever resolved from another WASM module, we can
+                // read trap and code length information from that module. In fact, this might even
+                // be rewritable as a replace by a clone of the exported FunctionSpec that satisfies
+                // this import!
+                let new_entry = FunctionSpec::new(
+                    *ptr as u64,
+                    existing_entry.code_len(),
+                    existing_entry
```
acfoltzer: what is the situation where it'd be possible for an import function to have associated trap info?

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -117,6 +118,94 @@ impl DlModule {
...
+                        .map(|manifest| manifest.traps.as_ptr() as u64)
+                        .unwrap_or(0),
+                    existing_entry.traps_len(),
+                );
+
+                std::mem::replace(existing_entry, new_entry);
+            }
+
+            // The function manifest may end up in the same page as code, for now, so this gets +x to
+            // permit those functions to execute. This should be should be fixed so .rodata can be
```
acfoltzer: typo in comment, also could we make an issue so we don't forget this?

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
awortman-fastly: admittedly I'm getting ahead of myself here, these should all be 0 when we're updating them at load-time. Eventually if we import from another module, we'd want the trap information available here to look up in the case something goes sideways.

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
awortman-fastly: :+1: good catch

lucet-module-data/src/functions.rs
```diff
@@ -21,12 +21,13 @@ impl FunctionIndex {
...
 /// ImportFunction describes an internal function - its internal function index and the name/module
-/// pair that function should be found in.
+/// pair that function should be found in, as well as the name of a hostcall it maps to.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize)]
 pub struct ImportFunction<'a> {
     pub fn_idx: FunctionIndex,
     pub module: &'a str,
-    pub name: &'a str
+    pub name: &'a str,
+    pub mapped_to: Option<&'a str>,
```
pchickey: I think `symbol` would be a better name than `mapped_to`. Also, we only ever construct the `Some` case in lucetc/src/decls.rs, and if we encounter a `None` case in DlModule we throw an error. Can we just get rid of the Option for now?

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -117,13 +120,113 @@ impl DlModule {
...
         Ok(Arc::new(DlModule {
             lib,
             fbase,
             module_data,
             function_manifest,
         }))
     }
+
+    /// `resolve_imports` fixes up entries in function manifest for imported functions,
+    /// which initialize to 0, with the correct address of the referenced function
```
pchickey: I don't understand why we couldn't do the same thing by using a relocation and having the loader take care of this. In `lucetc::function_manifest`, Can we get rid of the `len > 0` comparison in `write_relocated_slice` so that we emit a relocation to the import function symbol? The import function has relocations that work at call sites, so I think it should be possible for it to work here as well.

Maybe there's some reason for this that I haven't thought of. Otherwise this looks like good code to me.

lucetc/src/decls.rs
```diff
@@ -188,16 +211,34 @@ impl<'a> ModuleDecls<'a> {
...
                 decl_linkage,
                 self.info.signature_for_function(func_ix),
             )
             .context(LucetcErrorKind::TranslatingModule)?;
-        self.function_names.push(Name::new_func(decl_sym, funcid));
-        Ok(FuncIndex::new(self.function_names.len() - 1))
+
+        if func_ix.as_u32() as usize >= self.function_names.len() {
+            // func_ix is new, so we need to add the name
+            //
```
pchickey: sentence fragment

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
pchickey: just realized that import funcs wont have a trap table, so we'd still need to handle that case properly. we could provide the `Linkage` of a func in the `function_manifest` vec to make it more clear what is going on there.

awortman-fastly: Using a relocation for these is a good idea. The `len > 0` check was to avoid writing a relocation to a blob of data we wouldn't write (the function body) but I'll dig around to see if we can get the loader to fix this up to the import address. The only complication is that we don't want this to be a pointer to to the entry in the GOT, we want the entry here to be the same as the one in the GOT.

re trap tables, I had imagined import funcs could have trap tables, iff they were imported from a wasm module and we copied their trap information here. But in retrospect that wouldn't be sufficient to be useful for trap lookup (a module's direct dependency's trap info would be present, but not transitive dependencies, if any ever exist), so that really should all be left nil.

What would linkage offer in `function_manifest`? We can learn a function is imported/exported by matching the function index with information in `ModuleData` - are you thinking of a scenario where that wouldn't be workable?

lucet-module-data/src/functions.rs
awortman-fastly: `symbol` is good. After our conversation today, I'll make this just `&'a str` and we'll see out that pans out later. Got ahead of myself thinking about how module imports would look.

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
pchickey: I hadn't thought about the GOT issue, hopefully that is resolvable.

Agreed re trap tables.

I only thought the linkage should be available as an alternative to checking `len > 0` when constructing the manifest reloc. It doesn't need to be in the manifest itself.

lucet-module-data/src/functions.rs
awortman-fastly: Since this was only present to support the load-time import fixing, and that works by relocation now, this field's just gone

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
awortman-fastly: thought about this a bit and realized part of the problem was calling `write_relocated_slice` with constructed trap table symbols that never really existed, expecting `write_relocated_slice` to happen to discard those by relying on implicit knowledge. Rewrote it to make the imported data/local data cases clearer and for callers to not provide junk data.

lucetc/src/function_manifest.rs
```diff
@@ -5,21 +5,62 @@ use failure::{Error, ResultExt};
...
-            at: buf.position(),
-        })
-        .context(format!("linking {} into function manifest", to))?;
+    match (to, len) {
+        (Some(to), 0) => {
+            // This is an imported slice of unknown size
+            let absolute_reloc = match obj.target.binary_format {
+                BinaryFormat::Elf => {
+                    faerie::artifact::Reloc::Raw {
+                        reloc: 1, // this is an ELF R_X86_64_64
```
pchickey: Can you import these constants from `goblin`?

awortman-fastly: sure thing, I was kind of hoping we could avoid depending on `goblin` (directly) just for two constants but oh well :)

pchickey: Its a transitive dependency through faerie. I don't think it hurts :)

PR #204, opened 2019-06-08
Float global support

awortman-fastly:
Split from #196 into changes I'd be comfortable squashing together:

This adjusts globals to allow for a global exported by multiple names, and adds support for globals of type `f32` and `f64`. This is sufficient to get us spec-compliant for globals!

I noticed `lucet-analyze` was missing information about globals, so I added that as well.

This makes global values a union and adds some unsafe as a consequence, but my understanding is that type confusion of globals would requite wasm that doesn't validate, so this "only" would require additional unsafe code on the embedder's side. Since I imagine it's relatively unlikely to be prodding an instance's globals outside initialization, that seems acceptable vs the additional implementation complexity of separating out globals by type.

### DIFF
--- a/lucetc/src/decls.rs
+++ b/lucetc/src/decls.rs
@@ -267,25 +267,42 @@ impl<'a> ModuleDecls<'a> {
         for ix in 0..info.globals.len() {
             let ix = GlobalIndex::new(ix);
             let g_decl = info.globals.get(ix).unwrap();
-            let g_import = info.imported_globals.get(ix);
-            let g_variant = if let Some((module, field)) = g_import {
-                GlobalVariant::Import { module, field }
-            } else {
-                let init_val = match g_decl.entity.initializer {
-                    // Need to fix global spec in ModuleData and the runtime to support more:
-                    GlobalInit::I32Const(i) => i as i64,
-                    GlobalInit::I64Const(i) => i,
-                    _ => Err(format_err!(
-                        "non-integer global initializer: {:?}",
-                        g_decl.entity
-                    ))
-                    .context(LucetcErrorKind::Unsupported)?,
-                };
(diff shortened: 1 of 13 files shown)


lucetc/src/decls.rs
```diff
@@ -267,25 +267,42 @@ impl<'a> ModuleDecls<'a> {
...
+                        }
+                    } else {
+                        // This WASM restriction may be loosened in the future:
+                        Err(format_err!("invalid global declarations: global {} is initialized by referencing another global value, but the referenced global is not an import", ix.as_u32()))
+                        .context(LucetcErrorKind::TranslatingModule)
+                    }
+                }
+                GlobalInit::Import => {
+                    if let Some((module, field)) = info.imported_globals.get(ix) {
+                        Ok(GlobalVariant::Import { module, field })
```
awortman-fastly: This turns import globals into a runtime error for the moment. It probably makes sense to run `module` and `field` through the same name mapping as functions (eg `bindings.json`) but aside from spec tests it seems like the most common use of globals would be to import from other modules, which is also something lucet-runtime does not yet support.

pchickey: Yes, we need to figure out the design changes to lucet-runtime for import globals, memories, tables etc - right now import functions is special cased to use the linker, which is also not always possible or useful. So I think we should continue to defer that to a runtime error. This looks like the correct implementation for the compiler.

awortman-fastly: I totally imagine an argument for embedding-provided globals, so there might be a path where this grows in the same way `ImportFunction` has a new field in #203. But maybe by then we'll support loading multiple wasm modules, and embedder context is just a wasm-module-looking thing. Ah, the problems of tomorrow!

PR #209, opened 2019-06-12
WASI fixes

jedisct1:
- Remove unneeded `unsafe` blocks; codecs don't need these any longer.
- Removing a directory without `REMOVEDIR` may return `EPERM` instead of  `EISDIR`. Adjust accordingly. (thanks, @kubkon)
- If a path cannot be opened, set the file descriptor to `wasi_fd_t::max_value()`

This is fixing tests from https://github.com/sunfishcode/misc-tests

### DIFF
--- a/lucet-wasi/src/hostcalls/fs.rs
+++ b/lucet-wasi/src/hostcalls/fs.rs
@@ -60,12 +60,8 @@ pub fn wasi_fd_fdstat_get(
     } else {
         wasm32::__WASI_EBADF
     };
-
-    unsafe {
-        enc_fdstat_byref(vmctx, fdstat_ptr, host_fdstat)
-            .expect("can write back into the pointer we read from");
-    }
-
+    enc_fdstat_byref(vmctx, fdstat_ptr, host_fdstat)
+        .expect("can write back into the pointer we read from");
     errno
 }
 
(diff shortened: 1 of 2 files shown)


awortman-fastly:
less unsafe! :tada:

lucet-wasi/src/hostcalls/fs.rs
```diff
@@ -601,7 +576,26 @@ pub fn wasi_path_unlink_file(
     // nix doesn't expose unlinkat() yet
     match unsafe { unlinkat(dir, path_cstr.as_ptr(), 0) } {
         0 => wasm32::__WASI_ESUCCESS,
-        _ => wasm32::errno_from_nix(errno::Errno::last()),
+        _ => {
+            let mut e = errno::Errno::last();
+            // Non-Linux implementations may return EPERM when attempting to remove a
```
awortman-fastly: I realized right after clicking approve: do we know *which* non-Linux `unlinkat` may return EPERM? I would imagine this is a macos-ism, but would like to be clear if we can

jedisct1: In fact, this is the expected POSIX behavior. Only Linux returns `EISDIR` for this case, and this is even documented in the Linux man page.

awortman-fastly: Ah, I see that now. Thanks!

kubkon:
My pleasure @jedisct1. I was meant to sync with your repo guys and open a PR but didn’t get the time to do that yet. Glad you could extrapolate the fixes yourself! :-)

PR #210, opened 2019-06-12
lucet-runtime: add missing lucet_error_no_linear_memory in C header

htfy96:
Error::NoLinearMemory was added in d03aaad7b198ab4e4eff442e213871801df5c36f, but it didn't update the C header, causing GDB to display lucet_error incorrectly.

This change makes the definition of `lucet_error` in `lucet_types.h` same with `Error` in [error.rs](https://github.com/fastly/lucet/blob/master/lucet-runtime/lucet-runtime-internals/src/error.rs)


jedisct1:
Good catch, thank you!

PR #212, opened 2019-06-13
Update AssemblyScript, WASA and the AssemblyScript example and instructions

jedisct1:
This is an update to the AssemblyScript module and example, after the new
runtimes landed into the AssemblyScript compiler.

WASA, the WASI bindings, got a major update, not only to support the new ASC,
but also to support most of WASI as it is today.
Breaking changes were made to the API to make is both safer and closer to the
NodeJS filesystem API.

wasm-crypto was updated for the new runtime as well.

The example was updated for the new runtime and the new WASA API.

The AssemblyScript compiler still has some rough edges regarding garbage
collection. So, switch the runtime from `full` (default) to `stub` for now,
until some known regressions have been ironed out.

### DIFF
--- a/assemblyscript/examples/signatures/README.md
+++ b/assemblyscript/examples/signatures/README.md
@@ -24,6 +24,7 @@ npm install
 npm run asbuild:optimized
 
 lucetc -o example \
+  --reserved-size=64MB \
   --bindings /opt/lucet/share/assemblyscript/modules/wasa/bindings.json \
   build/optimized.wasm
 ```
@@ -31,7 +32,7 @@ lucetc -o example \
 ## Running the example using lucet-wasi
 
 ```sh
-lucet-wasi --entrypoint main --dir /lucet:/lucet example help
+lucet-wasi --entrypoint main --dir .:. example help
 ```
 
 Unlike C and Rust applications, AssemblyScript requires an entry point to be explicitly defined.
@@ -43,7 +44,7 @@ Since files are going to be read and written to, a descriptor to a pre-opened di
 
 This is the purpose of the `--dir` command-line option. Without this, the webassembly module cannot access the filesystem at all.
 
-Here, the `/lucet` virtual directory, as seen by the application, maps to the `/lucet` directory in the container.
+Here, the current virtual directory (`.`), as seen by the application, maps to the current directory in the container.
 
 ## What the example does
 
@@ -54,7 +55,7 @@ The main code is in `assembly/index.ts`.
 Key pair creation:
 
 ```sh
-lucet-wasi --entrypoint main --dir /lucet:/lucet example keypair
+lucet-wasi --entrypoint main --dir .:. example keypair
 ```
 
 ```text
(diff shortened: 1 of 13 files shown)


jedisct1:
More changes coming, since breaking changes to the way string conversions are done in AssemblyScript recently landed.

assemblyscript/examples/signatures/README.md
```diff
@@ -72,19 +73,19 @@ The WASA wrapper automatically sets the minimum required WASI capabilities in or
...
 ```
 
-This command reads the `/lucet/LICENSE` file, as well as the key pair, and computes a signature of the file's content that can be verified using the public key.
+This command reads the `README.md` file, as well as the key pair, and computes a signature of the file's content that can be verified using the public key.
 
 Signature verification:
 
 ```sh
-lucet-wasi --entrypoint main --dir /lucet:/lucet example verify LICENSE 94b8eb14373eb245c1daaacb2c24e2cb554bdd723009423aae5a8ca5fa99fa16 deedf3910d5b166ca17e0e307312a422cb50efcbcc90754cf0e2d528a9159c4ad3ac973e3cd9b2c2986fb2e467a0506bc9a5ceb9c7d6d30e360fb4d1cef3c50d
+lucet-wasi --entrypoint main --dir .:. example verify README.md 94b8eb14373eb245c1daaacb2c24e2cb554bdd723009423aae5a8ca5fa99fa16 deedf3910d5b166ca17e0e307312a422cb50efcbcc90754cf0e2d528a9159c4ad3ac973e3cd9b2c2986fb2e467a0506bc9a5ceb9c7d6d30e360fb4d1cef3c50d
```
acfoltzer: Since we're now using the README as the example, we should add a note here saying that you should use the output from the `sign` command here rather than this literal signature, which will fail to verify.

acfoltzer: Actually, I couldn't get it to verify even after using the signature produced by `sign`. Is there a trick to get it to use `keypair.bin` for the `verify` step?

jedisct1: Mmm... seems to work here:

```sh
$ lucet-wasi --entrypoint main --dir .:. example keypair
Creating a new keypair...
Key pair created and saved into [keypair.bin]
Public key: [6ec6ad279c6335382fa72a6ee23dbcddf835dc438d248b4a224425e04b4abb61]

$ lucet-wasi --entrypoint main --dir .:. example sign README.md
Signature for that file: [4cbae15aeb812c642dbbd40e32c50c7bcdaf8e75519c220508407918645011152145c941ff7c451f6b66c70fdada82fa0e2c90678dff93bed58332ef0b6cf005]

$ lucet-wasi --entrypoint main --dir .:. example verify README.md 6ec6ad279c6335382fa72a6ee23dbcddf835dc438d248b4a224425e04b4abb61 4cbae15aeb812c642dbbd40e32c50c7bcdaf8e75519c220508407918645011152145c941ff7c451f6b66c70fdada82fa0e2c90678dff93bed58332ef0b6cf005
This is a valid signature for that file
```

The public key can be extracted from `keypair.bin`, but the verifier usually doesn't have that file.

jedisct1: > Since we're now using the README as the example, we should add a note here saying that you should use the output from the sign command here rather than this literal signature, which will fail to verify.

Done.

acfoltzer: Ah, I was using the wrong public key. Got it!

PR #214, opened 2019-06-14
Enable Golang wasi-wasm 

JanFalkin:
We are very interested in using this lucet for the backing of our content fabric API.  We are a Go shop and managed to get find an early preview git@github.com:neelance/go.git wasi branch.  Unfortunately, wherein wasmtime was able to load an execute the wasm produced by this go branch. Lucetc was able to compile but lucet-wasi was unable to load and execute.  This pull is an attempt to correct that.   
It should be noted that I am very new to Rust and have done my best to learn and provide changes, however I am not confident that my lifetime choices of using static &str are the best.  Also golang seems to want to call wasi_exit on close and the current implementation of panic! seems messy considering lucet-wasi is itself a process.  Thus I used exit to close out the process.  Perhaps there is a nicer way of doing that with Panic!.  I verified the tests pass/fail the same as before my changes. 

Regardless we are enamored with the technology and implementation and we look forward to using it in our content fabric.  

I hope you find this useful

### DIFF
--- a/benchmarks/lucet-benchmarks/src/seq.rs
+++ b/benchmarks/lucet-benchmarks/src/seq.rs
@@ -29,7 +29,7 @@ fn hello_load_mkregion_and_instantiate<R: RegionCreate + 'static>(c: &mut Criter
     let workdir = TempDir::new().expect("create working directory");
 
     let so_file = workdir.path().join("out.so");
-    compile_hello(&so_file, OptLevel::Best);
+    compile_hello(&so_file, OptLevel::Fast);
 
     c.bench_function(
         &format!("hello_load_mkregion_and_instantiate ({})", R::TYPE_NAME),
@@ -57,7 +57,7 @@ fn hello_instantiate<R: RegionCreate + 'static>(c: &mut Criterion) {
     let workdir = TempDir::new().expect("create working directory");
 
     let so_file = workdir.path().join("out.so");
-    compile_hello(&so_file, OptLevel::Best);
+    compile_hello(&so_file, OptLevel::Fast);
 
     let module = DlModule::load(&so_file).unwrap();
     let region = R::create(1, &Limits::default()).unwrap();
@@ -125,7 +125,7 @@ fn hello_drop_instance<R: RegionCreate + 'static>(c: &mut Criterion) {
     let workdir = TempDir::new().expect("create working directory");
 
     let so_file = workdir.path().join("out.so");
-    compile_hello(&so_file, OptLevel::Best);
+    compile_hello(&so_file, OptLevel::Fast);
 
     let module = DlModule::load(&so_file).unwrap();
     let region = R::create(1, &Limits::default()).unwrap();
@@ -246,7 +246,7 @@ fn run_hello<R: RegionCreate + 'static>(c: &mut Criterion) {
     let workdir = TempDir::new().expect("create working directory");
 
     let so_file = workdir.path().join("out.so");
-    compile_hello(&so_file, OptLevel::Best);
+    compile_hello(&so_file, OptLevel::Fast);
 
     let module = DlModule::load(&so_file).unwrap();
     let region = R::create(1, &Limits::default()).unwrap();
--- a/benchmarks/lucet-benchmarks/src/seq.rs
+++ b/benchmarks/lucet-benchmarks/src/seq.rs
@@ -240,8 +240,7 @@ fn run_fib<R: RegionCreate + 'static>(c: &mut Criterion) {
 /// Run a trivial WASI program.
 fn run_hello<R: RegionCreate + 'static>(c: &mut Criterion) {
     fn body(inst: &mut InstanceHandle) {
-        const START: &'static str = "_start";
-        inst.run(START, &[]).unwrap();
+        inst.run("_start", &[]).unwrap();
     }
 
     let workdir = TempDir::new().expect("create working directory");
@@ -251,7 +250,6 @@ fn run_hello<R: RegionCreate + 'static>(c: &mut Criterion) {
 
     let module = DlModule::load(&so_file).unwrap();
     let region = R::create(1, &Limits::default()).unwrap();
-    const START: &'static str = "_start";
 
     c.bench_function(&format!("run_hello ({})", R::TYPE_NAME), move |b| {
         b.iter_batched_ref(
@@ -263,7 +261,7 @@ fn run_hello<R: RegionCreate + 'static>(c: &mut Criterion) {
                     .build()
                     .unwrap();
                 region
-                    .new_instance_builder(module.clone(),START)
+                    .new_instance_builder(module.clone(),"_start")
                     .with_embed_ctx(ctx)
                     .build()
                     .unwrap()
--- a/benchmarks/lucet-benchmarks/src/seq.rs
+++ b/benchmarks/lucet-benchmarks/src/seq.rs
@@ -240,7 +240,8 @@ fn run_fib<R: RegionCreate + 'static>(c: &mut Criterion) {
 /// Run a trivial WASI program.
 fn run_hello<R: RegionCreate + 'static>(c: &mut Criterion) {
     fn body(inst: &mut InstanceHandle) {
-        inst.run("_start", &[]).unwrap();
+        const START: &'static str = "_start";
+        inst.run(START, &[]).unwrap();
     }
 
     let workdir = TempDir::new().expect("create working directory");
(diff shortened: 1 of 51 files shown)


awortman-fastly:
Hi! First off, thank you for the PR! We really appreciate you going and fixing your issue, especially picking up a new language to do it :)

It looks like the issue you ran into is that we're executing the function declared in the start section of the module, something like `(start $yourfn)`. Through a C toolchain, that would also be present as an exported `guest_func__start`, which is what it looks to me like your go fork is also writing. Can you try ensuring that the start section in your produced module references `_start`? That _should_ work with `lucet-wasi` out of the box.

There's a few comments for the PR as well, but don't worry about addressing them; I mean them to be more informational - let's figure out if setting a `start` section fixes your issue first.

benchmarks/lucet-benchmarks/src/seq.rs
```diff
@@ -240,7 +240,8 @@ fn run_fib<R: RegionCreate + 'static>(c: &mut Criterion) {
 /// Run a trivial WASI program.
 fn run_hello<R: RegionCreate + 'static>(c: &mut Criterion) {
     fn body(inst: &mut InstanceHandle) {
-        inst.run("_start", &[]).unwrap();
+        const START: &'static str = "_start";
```
awortman-fastly: `"_start"` _is_ a `&'static str` by definition, as all string literals are, so you don't have to actually store it anywhere before using - it's entirely valid and correct to write `inst.run("_start", &[])`, here, and similarly for other places that take `&str`

lucet-wasi/src/main.rs
```diff
@@ -182,22 +183,12 @@ fn run(config: Config) {
...
             .expect("instance can be created");
-
-        match inst.run(config.entrypoint, &[]) {
-            // normal termination implies 0 exit code
-            Ok(_) => 0,
-            Err(lucet_runtime::Error::RuntimeTerminated(
-                lucet_runtime::TerminationDetails::Provided(any),
-            )) => *any
-                .downcast_ref::<lucet_wasi::host::__wasi_exitcode_t>()
-                .expect("termination yields an exitcode"),
```
awortman-fastly: This looks like it has the unexpected side-effect of always returning an exit code of 0, even if the wasm that was executed returned a non-0 exit code. If that doesn't fail a test I suspect we might be missing tests :)

Just because you mentioned the `panic!` here: the panic case is for an aytpical runtime exit, such as a timeout or the runtime reaching a known-invalid state, in contrast to an execution where the module ran, everything worked fine, and the program that ran reports an error

JanFalkin: Thanks for the feedback.  I agree it will always return 0, but in effect the statement before will error out in erroneous cases.   
region
            .new_instance_builder(module as Arc<dyn Module>, config.entrypoint)
            .with_embed_ctx(ctx.build().expect("WASI ctx can be created"))
            .build()
Thus if a valid handle is returned we are exit 0.  At least that was my thinking but perhaps I am misunderstanding more profound.

benchmarks/lucet-benchmarks/src/seq.rs
JanFalkin: I thought I had tried that but as a newb perhaps I made a mistake.  I was getting compiler errors complaining about lifetimes of the &str which was why I made explicit static.  Regardless I will to do as you request. Thanks

JanFalkin: I see from above you were a bit tongue and cheek for this change but its done anyway.  Looks better too :-)

lucet-wasi/src/main.rs
JanFalkin: I may be misunderstanding you first comment about _start not being explicitly referenced by the go wasm but I believe it is.  lucetc seems to be acquiring it from its export table and calling what it believes is the correct function index (in my case that was 851). The problem that I see is that the go linker is creating a duplicate record for wasi_fd_write in imports causing the indexing to be off by one and having the lucet-wasi code end up in function 850 causing a barf.  I stepped through wasmtime to see how it was avoiding this issue, and it is doing something similar to the approach I took.  I am not certain lucet-wasi is folding duplicate imports properly. Wasmtime doesn't do it either.  At first I tried to alter the go linker to avoid putting out dupes, but that is tricky and while I could figure out how to remove them from imports table I could not correctly resolve all the old references to the removed duplicate.  I am really a C++ dev pretending to be a Go dev learning Rust hacking on compilers and linkers.  Hope that clarifies.  If you like I can attach sample go code and resultant wasm.

JanFalkin:
I may be misunderstanding you first comment about _start not being explicitly referenced by the go wasm but I believe it is. lucetc seems to be acquiring it from its export table and calling what it believes is the correct function index (in my case that was 851). The problem that I see is that the go linker is creating a duplicate record for wasi_fd_write in imports causing the indexing to be off by one and having the lucet-wasi code end up in function 850 causing a barf. I stepped through wasmtime to see how it was avoiding this issue, and it is doing something similar to the approach I took. I am not certain lucet-wasi is folding duplicate imports properly. Wasmtime doesn't do it either. At first I tried to alter the go linker to avoid putting out dupes, but that is tricky and while I could figure out how to remove them from imports table I could not correctly resolve all the old references to the removed duplicate. I am really a C++ dev pretending to be a Go dev learning Rust hacking on compilers and linkers. Hope that clarifies. If you like I can attach sample go code and resultant wasm.

awortman-fastly:
Ah! I would love a copy of the go code and resulting wasm (with steps to build please, I'm not very familiar with Go). I've suspected duplicate imports might still be an issue, but couldn't come up with a test case the last time I tried. If it results in function indices pointing to the wrong code, I think that's safe to say is a lucetc bug, which should handle merging imports together.

Technically duplicate imports are valid wasm, but definitely isn't ideal. We should be able to handle that correctly in any case.

JanFalkin:
I just took this simple go sample for quicksort
http://www.golangprograms.com/golang-program-for-implementation-of-quick-sort.html
I built this wasm using the custom Go I was mentioning and I have attached as a zip.  I wanted to use 7z as it was half the size but alas github will not cooperate.  I will send more detailed instructions on building you own Go.  Effectively you will need Go to build Go but it is very similar to rust in its build simplicity.  Let me know if the wasm inflates ok and you can make heads or tails of it.  I can explain some of the Go symbols in the resultant wasm as I walked through it all on the Go side.  
The Go custom branch is here and I will send instructions once I cook them up better than the Go readme has it.  git@github.com:neelance/go.git wasi branch
[quicksort.zip](https://github.com/fastly/lucet/files/3292619/quicksort.zip)

awortman-fastly:
You are definitely hitting a lucetc bug! This is a screenshot from `lucet-analyze`, looking at the quicksort wasm you attached, after building through lucetc:

lucet-analyze screenshot showing function names are misaligned

The important part is down towards the bottom and shows how indices are incorrect as you observed. `name` is the name associated with the function index in the module's metadata, but `lucet-analyze` also shows the exported symbol, if present, for the address associated with that function. At the bottom lucet-analyze reports that we're writing out that the function `__wasi_fd_write` has the pointer to `guest_func_12` (and subsequently all function records have the pointer for the _following_ function).

I'll take a closer look Monday but with a test case in hand I think it can be fixed fairly quickly.

I think I better understand the motivation for your change now. Having `get_start_func` take a function name lets you look up the function by symbol, which gives you the right function, rather than using something like `inst.get_export_func`, which was giving you an incorrect pointer. And with the wrong pointer, lucet-wasi was running the wrong code and .. probably crashing?

JanFalkin:
First sorry I didn't see your comment till just now.  I thought github sent email with every update and I neglected to refresh the page.  Got lucky and clicked on a mail that refreshed the page. 
The crash is absolutely what happened.  I will look to lucet-analyze more in the future and you description above is very helpful. 
I was thinking over the weekend that my fix will only help for start as any other exports will be similarly messed up if I get what's going on correctly.  
I am really looking forward to seeing the resulting changes (all part of the learning process) .  Thanks again for all of your feedback and let me know if I can be of assistance in any way

JanFalkin:
Also before I forget there is still a problem with exit.  Go is explicitly calling wasi_exit. I believe the code panics and unwinds ugly in the console.  I put those simple changes in for exit that I know you are not a fan of :-)

awortman-fastly:
Good news! With the changes in #219, `lucetc` no longer produces an incorrect (to my eye, at least) module. It does panic for other reasons, but I think you might be in a better position to understand what's going on:
quicksort not quicksorting

JanFalkin:
That was it. 219 just fixes it, no panic!!   I am curious to see if the go wasm still ends up calling exit in the same way.  Perhaps there was more there than met my eye initially and it all flowed from the one of indexing.  Nice job and thanks!

JanFalkin:
Desired functionality resolved in https://github.com/fastly/lucet/pull/219

Issue #193, opened 2019-06-03
Lucetc API to build from &[u8]


singpolyma-shopify:
Right now the `Lucetc` API requires passing a file path, but on calling `build` the first thing it does is slurp all the bytes into RAM.  I understand that the main purpose of this library is to implement the `lucetc` CLI and for that it doesn't much matter, but it would be very useful to be able to embed the compiler in my app, where I already have the WASM bytecode in a `Vec<u8>` and not in a file.

Would it be reasonable to have the `Lucetc` API build from `AsRef<u8>` and move the file-slurp call either into `main.rs` or into a second constructor for `Lucetc`?

jedisct1:
Hi,

This is a great suggestion, and definitely something we will add to the API.

Thanks!

data-pup:
I'd like to try and take care of this :slightly_smiling_face:

jedisct1:
Awesome, thank you @data-pup !

acfoltzer:
Closed by #215; thanks @data-pup!

PR #215, opened 2019-06-14
lucetc::Lucetc::from_bytes

data-pup:
This takes care of #193. :smile_cat: 

Adds a `from_bytes<B: AsRef<[u8]>>(bytes: B)` method so that `lucetc::Lucetc` can be constructed from bytes. This means that we aren't required to write a program to disk in certain situations, if we already have the wasm binary content in memory.

`from_bytes` is used rather than a `from_path` because I didn't want to change the behavior of `new`.

While poking around at this I found that the stack tests in `lucet-runtime/lucet-runtime-tests/src/stack` were a good candidate to switch from `Lucetc::new` to `Lucet::from_bytes`, skipping the need to write the wat content generated by `generate_test_wat` to a file in the tempdir. :tada: 

p.s. big thanks & kudos to @awortman-fastly for helping me out getting started w/ this :tada:

### DIFF
--- a/lucet-runtime/lucet-runtime-tests/src/stack.rs
+++ b/lucet-runtime/lucet-runtime-tests/src/stack.rs
@@ -1,20 +1,16 @@
 use failure::Error;
 use lucet_runtime_internals::module::DlModule;
 use lucetc::Lucetc;
-use std::fs::File;
-use std::io::prelude::*;
 use std::sync::Arc;
 use tempfile::TempDir;
+use wabt::wat2wasm;
 
 pub fn stack_testcase(num_locals: usize) -> Result<Arc<DlModule>, Error> {
+    let wasm_program = wat2wasm(generate_test_wat(num_locals)).expect("test wat is valid");
+
+    let native_build = Lucetc::from_bytes(wasm_program);
+
     let workdir = TempDir::new().expect("create working directory");
-
-    let wasm_path = workdir.path().join("out.wasm");
-
-    let mut wasm_file = File::create(&wasm_path)?;
-    wasm_file.write_all(generate_test_wat(num_locals).as_bytes())?;
-
-    let native_build = Lucetc::new(wasm_path);
 
     let so_file = workdir.path().join("out.so");
 
--- a/lucet-runtime/lucet-runtime-tests/src/stack.rs
+++ b/lucet-runtime/lucet-runtime-tests/src/stack.rs
@@ -3,12 +3,9 @@ use lucet_runtime_internals::module::DlModule;
 use lucetc::Lucetc;
 use std::sync::Arc;
 use tempfile::TempDir;
-use wabt::wat2wasm;
 
 pub fn stack_testcase(num_locals: usize) -> Result<Arc<DlModule>, Error> {
-    let wasm_program = wat2wasm(generate_test_wat(num_locals)).expect("test wat is valid");
-
-    let native_build = Lucetc::from_bytes(wasm_program);
+    let native_build = Lucetc::try_from_bytes(generate_test_wat(num_locals))?;
 
     let workdir = TempDir::new().expect("create working directory");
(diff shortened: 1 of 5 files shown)


lucetc/src/lib.rs
```diff
@@ -31,8 +31,13 @@ use std::env;
 use std::path::{Path, PathBuf};
 use tempfile;
 
+enum LucetcInput {
+    Bytes(Vec<u8>),
+    Path(PathBuf),
+}
```
data-pup: We have a `LucetcInput` enum rather than reading the bytes in `new`, and changing the type of `input` to a `Vec<u8>` because we don't always have a program _in_ the given path when `Lucetc` is constructed. (The `lucet-wasi-sdk/src/lib.rs:compile_to_lucet` test is an example of this)

Instead, we use a lil' enum so that the order of operations doesn't need to change, and can handle when we are directly given bytes, or when we have been given a path from which to read later.

lucetc/src/lib.rs
```diff
@@ -150,7 +155,17 @@ impl Lucetc {
     pub fn new<P: AsRef<Path>>(input: P) -> Self {
         let input = input.as_ref();
         Self {
-            input: input.to_owned(),
+            input: LucetcInput::Path(input.to_owned()),
```
data-pup: here's the only change in `new`, wrapping the path in the corresponding enum variant.

lucetc/src/lib.rs
```diff
@@ -162,7 +177,10 @@ impl Lucetc {
         use parity_wasm::elements::{deserialize_buffer, serialize};
 
         let mut builtins_bindings = vec![];
-        let mut module_binary = read_module(&self.input)?;
+        let mut module_binary = match &self.input {
+            LucetcInput::Bytes(bytes) => bytes.clone(),
+            LucetcInput::Path(path) => read_module(&path)?,
+        };
```
data-pup: Here's the difference in reading the module binary. We either have the binary itself already (`Bytes`) in which case, we can just clone. Otherwise, we have a path, and we can use the same call to `read_module` as before.

lucet-runtime/lucet-runtime-tests/src/stack.rs
```diff
@@ -1,20 +1,16 @@
...
-use std::io::prelude::*;
 use std::sync::Arc;
 use tempfile::TempDir;
+use wabt::wat2wasm;
 
 pub fn stack_testcase(num_locals: usize) -> Result<Arc<DlModule>, Error> {
-    let workdir = TempDir::new().expect("create working directory");
-
-    let wasm_path = workdir.path().join("out.wasm");
+    let wasm_program = wat2wasm(generate_test_wat(num_locals)).expect("test wat is valid");
```
awortman-fastly: Since `Lucetc` takes care of `wat2wasm` for the "from a file" case, I think it makes for a less surprising API for it to handle that translation in the "from bytes" case as well. Can you adjust `from_bytes` to also do a preamble check/translation? Might make sense to just split `lucetc/src/load.rs:read_module` up a bit for that

awortman-fastly:
:tada: Thank you! I've got just that one comment really. I'll defer to other folks for an actual approval though, since we've been chatting through this

lucet-runtime/lucet-runtime-tests/src/stack.rs
data-pup: ooh, good idea! i can do that

jedisct1:
Thanks for contributing this!

These changes look good to me as well. But as suggested by @awortman-fastly , maybe the optional translation from `wat` should be handled here.

jedisct1:
Looking good, thanks a lot @data-pup !

Travis is complaining because the indentation doesn't match what`rustfmt` would produce.

Not a big deal, that can done right after :)

PR #217, opened 2019-06-14
Update the shootout makefile to use the `fast` optimization level

### LINKED ISSUE #216: Benchmark OptLevel seems incorrect
When I attempt to run the `benchmarks/lucet-benchmarks` through, e.g., `make bench` the build fails. It seems that on my system clang expects the optimization level to be `fastest`, not `fast`:

```
$ clang-8 --version
clang version 8.0.0 (Fedora 8.0.0-1.fc30)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /bin

$ export WASI_SYSROOT=/opt/wasi-sysroot
$ export CLANG_ROOT=/usr/lib64/clang/8.0.0
$ export CLANG=/bin/clang-8
```

When I alter `compiler.rs` (diff below) then I can build and run the benchmarks:

```
diff --git a/lucetc/src/compiler.rs b/lucetc/src/compiler.rs
index 9cc82d8..dfc4925 100644
--- a/lucetc/src/compiler.rs
+++ b/lucetc/src/compiler.rs
@@ -37,7 +37,7 @@ impl Default for OptLevel {
 impl OptLevel {
     pub fn to_flag(&self) -> &str {
         match self {
-            OptLevel::None => "fast",
+            OptLevel::None => "fastest",
             OptLevel::Standard => "default",
             OptLevel::Fast => "best",
         }
```

I have several questions:
1. Does anyone know why my system thinks the opt-level should be `fastest`? Should I make a merge request to handle this type of difference?
2. For consistency, shouldn't `OptLevel::Fast` point to `"fast"` (or `"fastest"` in my case) and `OptLevel::None` to `"best"`? And should submit a merge request for this?

jedisct1:
`best` doesn't exist any more.

Fixes #216

### DIFF
--- a/benchmarks/shootout/Makefile
+++ b/benchmarks/shootout/Makefile
@@ -16,7 +16,7 @@ SHOOTOUT_SRCS:=$(shell ls $(SHOOTOUT)/*.c)
 SHOOTOUT_NATIVE_OBJS:=
 SHOOTOUT_LUCET_OBJS:=
 
-LUCETC_FLAGS:=--opt-level best --min-reserved-size 4294967296
+LUCETC_FLAGS:=--opt-level fast --min-reserved-size 4294967296
 COMMON_CFLAGS:=--std=c99 -Ofast -Wall -W -I$(SIGHTGLASS)/include
 
 SHOOTOUT_NATIVE_CFLAGS:=-march=native -fPIC \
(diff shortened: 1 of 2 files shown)


abrown:
This is exactly the change I make in my repo 👍

PR #218, opened 2019-06-17
Add flag to make wasm tables writable by the host

shravanrn:
Per https://github.com/WebAssembly/design/issues/1117, the wasm tables are modifiable by the host. However, lucet is currently emitting these tables as read only ELF sections. 

This PR makes these tables writable (This is hidden behind a flag "--writable-tables" for now)

*Use Case*
I am a PhD student working with Mozilla to look at the using WASM (with the WASI backend) to sandbox libraries used in the Firefox codebase. However, in our use case, the host can dynamically decide which host functions are callable by the wasm module. Thus the inability to modify the tables is blocking. 

I had a brief conversation with iximeow on the Mozilla IRC who suggested I open a PR to start a discussion. Please let me know if you have thoughts or require more detail etc.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/module.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/module.rs
@@ -20,6 +20,21 @@ pub struct TableElement {
     rf: u64,
 }
 
+impl TableElement {
+    pub fn set_element(&mut self, new_ty: u64, new_rf: u64) {
+        self.ty = new_ty;
+        self.rf = new_rf;
+    }
+
+    pub fn ty(&self) -> u64 {
+        self.ty
+    }
+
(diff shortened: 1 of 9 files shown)


acfoltzer:
I'm traveling right now, and will be able to provide a more detailed response later, but just to summarize what I said in IRC:

I think unfortunately that this will be a more complicated change than what is in your PR. Because one Lucet `DlModule` is potentially shared between many Lucet `Instance`s, we cannot allow the module to be mutable, because that would affect all instances. Due to upcoming additions to the wasm spec we will be adding more tables, as well as the ability to mutate the existing ones, so this is definitely in our roadmap. But it will involve some changes to how `MmapRegion`, `Slot`, `Alloc` and friends work, which makes it much trickier.

PR #219, opened 2019-06-17
Fix duplicate imports causing broken module data

awortman-fastly:
This addresses an issue with modules like
```
(module
  (type (func (param i32 i32 i32 i32) (result i32)))
  (type (func))
  (func (import "wasi_unstable" "fd_read") (type 0)) 
  (func (import "wasi_unstable" "fd_read") (type 0)) 
  (func $_setup (type 1) return)
  (export "_start" (func $_setup))
)
```
where the repeat import of `fd_read` causes module data and the function manifest to be out of sync, and function manifest entries to start having incorrect function pointers. That itself is a problem, but manifests in very strange behavior anywhere between crashes (trying to call a null pointer) and executing incorrect functions (if looking up an export after the two lists become misaligned).

This is addressed by a similar internal/external index split as with signatures. Internally, repeat declarations are mapped to a unique function index, and that index is used by lucetc in building module data, so nothing downstream has to care that modules can have these kind of declarations.

This fixes (kind of - other stuff came up there) #214


lucetc/tests/wasm.rs
```diff
@@ -78,6 +78,32 @@ mod module_data {
...
+        assert_eq!(mdata.import_functions().len(), 2);
+        assert_eq!(mdata.import_functions()[0].module, "env");
+        assert_eq!(mdata.import_functions()[0].name, "read");
+        assert_eq!(mdata.import_functions()[1].module, "env");
+        assert_eq!(mdata.import_functions()[1].name, "write");
+        assert_eq!(mdata.function_info().len(), 5);
+        assert_eq!(mdata.function_info()[0].name, Some("host_read"));
+        assert_eq!(mdata.function_info()[1].name, Some("host_write"));
+        assert_eq!(mdata.function_info()[2].name, Some("guest_func__start"));
+        assert_eq!(mdata.export_functions().len(), 1);
```
awortman-fastly: This test brushes up with exported imports not being fully supported yet (see #203) - this test will change once that PR is in as there should be three exported functions ("read", "write", and "_start")

lucetc/src/module.rs
```diff
@@ -10,6 +10,13 @@ use cranelift_wasm::{
 use lucet_module_data::UniqueSignatureIndex;
 use std::collections::{hash_map::Entry, HashMap};
 
+/// UniqueFuncIndex names a function after merging duplicate function declarations to a signle
```
JanFalkin: single not signle

awortman-fastly: good catch, thanks

lucetc/tests/wasm.rs
```diff
@@ -78,6 +78,32 @@ mod module_data {
         assert_eq!(mdata.function_info().len(), 3);
         assert_eq!(mdata.export_functions()[0].names, vec!["main"]);
     }
+
+    #[test]
+    fn duplicate_imports() {
+        let m = load_wat_module("duplicate_imports");
+        let b = Bindings::from_file(&PathBuf::from(
```
acfoltzer: should just be able to use `Path::new()` here to avoid allocation, not that it matters hugely in a test

lucetc/tests/wasm.rs
acfoltzer: So it looks like this test fail once #203 lands? Excellent

lucetc/tests/wasm/duplicate_imports.wat
```diff
@@ -0,0 +1,34 @@
+(module
```
acfoltzer: Could we add a version of this test to the `lucet-wasi` suite, so that we can observe it working correctly at runtime?

PR #221, opened 2019-06-21
End-to-end authentication and integrity verification of Lucet assets

jedisct1:
Lucetc can now verify the signature of a source file before compiling it.

The object resulting from the compilation of that source file can itself embed a signature, possibly created using a different key.

Lucetc can also create key pairs. These can be either password protected for interactive use, or passwordless, for automated deployment.

And the Lucet runtime gets new options to verify the embedded signature before loading a module.


lucetc/src/main.rs
```diff
@@ -19,14 +22,19 @@ fn main() {
         if !msg.ends_with('\n') {
             msg.push('\n');
         }
-        io::stderr().write(msg.as_bytes()).unwrap();
+        io::stderr().write_all(msg.as_bytes()).unwrap();
```
awortman-fastly: oh jeez, good catch

lucet-wasi/src/main.rs
```diff
@@ -158,7 +178,15 @@ fn run(config: Config) {
...
-        let module = DlModule::load(&config.lucet_module).expect("module can be loaded");
+        let pk = match (config.verify, config.pk_path) {
+            (false, _) => None,
+            (true, Some(pk_path)) => {
+                Some(PublicKey::from_file(pk_path).expect("public key can be loaded"))
+            }
+            (true, None) => panic!("signature verification requires a public key"),
+        };
+        let module =
+            DlModule::load_and_verify(&config.lucet_module, pk).expect("module can be loaded");
```
awortman-fastly: Usage/style nit: I'm pretty uncomfortable with `load_and_verify` possibly _not_ verifying (eg when pk is `None`) - I think this function makes sense as a helper internal to `DlModule`, but for a public function I think it's worth being a stickler about making the caller very aware they are (or are not) verifying at load.

lucet-module-data/src/signature.rs
```diff
@@ -0,0 +1,172 @@
...
+        Ok(())
+    }
+
+    // Retrieving the offset of a symbol is not supported by the object crate.
+    // In Mach-O, actual file offsets are encoded, whereas Elf encodes virtual
+    // addresses, requiring extra steps to retrieve the section, its base
+    // address as well as the section offset.
+
+    // Elf
+    #[cfg(not(target_os = "macos"))]
```
awortman-fastly: Can this be `#[cfg(all(target_family = "unix", not(target_os = "macos")))]` just to be a clear breakage point when building Lucet for a platform we know won't work? (Really just leaves Windows, but still!)

lucet-wasi/src/main.rs
jedisct1: Nod. I'm going to explicitly call `load` or `load_and_verify` instead.

lucet-module-data/src/signature.rs
jedisct1: Sure :)

lucet-wasi/src/main.rs
jedisct1: Done.

lucet-module-data/src/signature.rs
jedisct1: Done, we're now breaking on non-Unix systems. Although I hope we'll support WebAssembly at some point :)

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -65,6 +79,12 @@ impl DlModule {
             unsafe { slice::from_raw_parts(*module_data_ptr, *module_data_len) };
         let module_data = ModuleData::deserialize(module_data_slice)?;
 
+        // If a public key has been provided, verify the module signature
+        // The TOCTOU issue is unavoidable without reimplenting `dlopen(3)`
```
awortman-fastly: after spending some time thinking about this, my only conclusion is `"dlopen() not taking a file descriptor is a crime"`

lucet-module-data/src/signature.rs
```diff
@@ -0,0 +1,172 @@
...
+        Ok(RawModuleAndData {
+            obj_bin,
+            module_data_offset: module_data_symbol_data.offset,
+            module_data_len: module_data_symbol_data.len,
+        })
+    }
+
+    pub fn module_data_bin(&self) -> &[u8] {
+        &self.obj_bin[self.module_data_offset as usize
+            ..self.module_data_offset as usize + self.module_data_len]
```
awortman-fastly: I only _just_ realized this as a rust-y replacement for `[a:b]` style indices for a > 0:
```
&self.obj_bin[self.module_data_offset as usize..][..self.module_data_len]
```
just a thought, this reads fine as-is

awortman-fastly:
I've developed some light paranoia around signing/verification, so while these changes look good, and really most of the diff is rustfmt stuff, I'd like someone else to look as well

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
jedisct1: 🤷

helpers/lucet-toolchain-tests/signature.sh
```diff
@@ -0,0 +1,91 @@
...
+LUCET_DIR="."
+LUCET_TARGET_DIR="${LUCET_DIR}/target/debug"
+TMPDIR="$(mktemp -d)"
+
+if [ -d "${LUCET_DIR}/target/release" ]; then
+    LUCET_TARGET_DIR="${LUCET_DIR}/target/release"
+fi
+
+if ! command -v rsign >/dev/null; then
+    cargo install rsign2
```
pchickey: can we install this in the devenv image by default?

lucetc/src/signature.rs
```diff
@@ -0,0 +1,88 @@
...
+            Ok(kp)
+        }
+    }
+}
+
+pub fn verify(buf: &[u8], signature_box: &SignatureBox, pk: &PublicKey) -> Result<(), Error> {
+    minisign::verify(pk, signature_box, Cursor::new(buf), false, false)
+        .map_err(|_| format_err!("Unable to verify the signature"))
+}
+
```
pchickey: Can some of this machinery be moved into the lucet_module_data crate? I can see the argument for `keygen` and `signature_path` staying here, but can `sign` and `verify` be moved into `lucet_module_data` itself? It would make more sense to expose those primitives than `RawModuleAndData`, which doesn't seem like a very useful public api.

It's a bummer that we can't sign the code until the linker has turned it into a shared object :/. between that and the TOCTOU issue, this is a bit unsatisfying, but I don't see a good way to solve either of those without a massive undertaking.

jedisct1: `RawModuleAndData` is not public any more, and `sign` was moved to the `module_data` crate. But in Lucetc `verify` verifies the source code signature, which `module_data` has nothing to do with. It was renamed to `verify_source_code` in order to avoid confusion.

helpers/lucet-toolchain-tests/signature.sh
jedisct1: Done. But once the next stable Rust is out, we can run the webassembly version with Lucet instead.

lucetc/src/signature.rs
pchickey: Great, the name verify confused me and I got mixed up, so this rename and reorganization helps a lot. Thanks

PR #222, opened 2019-06-21
Dynamic traits without "dyn" will not be supported any more

jedisct1:
Also add explicit lifetimes, because implicit ones is so 2015.


jedisct1:
Done.

Unfortunately, I don't know of a way to enforce the rule for implicit lifetimes.

I was hoping that this  requirement would eventually get reverted, but apparently not :(

PR #224, opened 2019-06-22
Fix nix->wasi file timestamp conversion

jedisct1:
`st_*tim` value are in seconds, not nanoseconds.

### DIFF
--- a/lucet-wasi/src/host.rs
+++ b/lucet-wasi/src/host.rs
@@ -227,9 +227,9 @@ pub fn filestat_from_nix(filestat: nix::sys::stat::FileStat) -> __wasi_filestat_
         st_ino: filestat.st_ino as __wasi_inode_t,
         st_nlink: filestat.st_nlink as __wasi_linkcount_t,
         st_size: filestat.st_size as __wasi_filesize_t,
-        st_atim: filestat.st_atime as __wasi_timestamp_t,
-        st_ctim: filestat.st_ctime as __wasi_timestamp_t,
-        st_mtim: filestat.st_mtime as __wasi_timestamp_t,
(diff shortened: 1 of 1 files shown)


lucet-wasi/src/host.rs
```diff
@@ -227,9 +227,9 @@ pub fn filestat_from_nix(filestat: nix::sys::stat::FileStat) -> __wasi_filestat_
         st_ino: filestat.st_ino as __wasi_inode_t,
         st_nlink: filestat.st_nlink as __wasi_linkcount_t,
         st_size: filestat.st_size as __wasi_filesize_t,
-        st_atim: filestat.st_atime as __wasi_timestamp_t,
-        st_ctim: filestat.st_ctime as __wasi_timestamp_t,
-        st_mtim: filestat.st_mtime as __wasi_timestamp_t,
+        st_atim: filestat.st_atime as __wasi_timestamp_t * 1_000_000_000,
```
acfoltzer: seems like we're just asking for overflows here. should we look at adding seconds fields to the WASI type?

jedisct1: Adding seconds to the WASI type would indeed be a very good idea.

acfoltzer: oh, I forgot those were 64-bit fields. I think we'll be fine :slightly_smiling_face:

jedisct1: Yes, these are 64-bit fields, so overflows will not happen, but we constantly have to convert between representations that are either in `time_t`, `timespec` or `timeval`, so having a second field would make things more consistent. But it's probably not worth a breaking change either.

PR #228, opened 2019-06-24
Update lucetc to match cranelift upstream changes

awortman-fastly:
Byproduct of testing that changes I made to cranelift look good by our tests as well, most of this is that FuncEnvironment functions are now fallible, data alignment can now be specified, and changes I don't quite follow about how source locations are tracked.


lucetc/src/compiler.rs
```diff
@@ -72,14 +72,15 @@ impl<'a> Compiler<'a> {
         }
 
         translate_module(wasm_binary, &mut module_info).map_err(|e| match e {
+            WasmError::User(_) => e.context(LucetcErrorKind::Input),
```
awortman-fastly: @pchickey I saw you mentioned wanting this on the corresponding cranelift issue, were you thinking about doing anything in particular with this variant? I have it contextualized as `Input` because I imagine this would be "user provided invalid input/options"

pchickey: I don't think we need to do anything special for now. In the future we may customize error reporting for invalid user-provided input but the way you've done it here is sufficient.

pchickey:
just needs one last format fix

awortman-fastly:
that's what i get for fixing merge conflicts online

PR #230, opened 2019-06-25
Add a helper script to set the global package version

jedisct1:
Check that in their current state, subpackages can be published.

Only then, set the global version number to the same as the main Cargo.toml file

Then, try one last dry run with the new version number.

### DIFF
--- a/helpers/bump-global-version.sh
+++ b/helpers/bump-global-version.sh
@@ -1,0 +1,27 @@
+#! /bin/sh
+
+VERSION=$(grep '#*Lucet version ' Cargo.toml | sed 's/^ *# Lucet version *//')
+[ -z "$VERSION" ] && echo "Version header not found in the top Cargo.toml file" >&2 && exit 1
+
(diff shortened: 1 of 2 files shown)


helpers/bump-global-version.sh
```diff
@@ -0,0 +1,27 @@
...
+    done || exit 1
+    echo
+    echo "Done."
+}
+
+version_bump() {
+    echo
+    echo "Setting the global Lucet version number to: [$VERSION]"
+    find lucetc lucet-* benchmarks/lucet-* -type f -maxdepth 1 -name 'Cargo.toml' -print | while read -r file; do
+        sed -i '' "s/^ *version *=.*/version = \"${VERSION}\"/" "$file"
```
pchickey: the set of single quotes `''` after `-i` break it on my machine - sed then thinks the regular expression is the filename

jedisct1: Updated :)

PR #231, opened 2019-06-26
Move wasi-sdk tests from lucetc to wasi-sdk

jedisct1:


lucet-wasi-sdk/tests/wasi-sdk.rs
```diff
@@ -0,0 +1,205 @@
+#[cfg(test)]
+mod tests {
```
pchickey: there are really two modules of tests here - the original tests from the crate which check that the Compile and Link abstractions work, and the tests which use lucetc to examine the resulting module. Can you separate them so this distinction is clear? that will help us identify breakages more easily

jedisct1: Done.

PR #232, opened 2019-06-26
move Bindings from lucetc to lucet-module-data

pchickey:
bindings are an (implicit) contract between compiler and runtime, so
they belong there instead

more importantly this breaks the dependency between lucet-wasi and
lucetc.


lucetc/src/lib.rs
```diff
@@ -31,6 +29,7 @@ pub use crate::{
     patch::patch_module,
 };
 use failure::{format_err, Error, ResultExt};
+use lucet_module_data::bindings::Bindings;
```
acfoltzer: since Bindings are basically a requirement to compile a module, I would `pub use` this so that most clients can just depend on `lucetc`

PR #242, opened 2019-07-09
Reorganize lucet module binary format

awortman-fastly:
This (finally!) makes a change we discussed a few months ago and has compilation produce binaries where all the internal components we need are reachable from a single symbol (`lucet_module`). This is nice for organization reasons, but generally reduces our dependence on `dlsym` as well as removing a few failure modes that were never realistically hit, like `function_manifest_len` not being present in a lucetc-compiled `.so`

Because other structures are internal and reachable via `lucet_module` and the `NativeData` it points to, most symbols are no longer exported.

### DIFF
--- a/lucet-module-data/src/module.rs
+++ b/lucet-module-data/src/module.rs
@@ -1,6 +1,6 @@
+use crate::functions::FunctionSpec;
 use crate::module_data::ModuleData;
 use crate::tables::TableElement;
-use crate::functions::FunctionSpec;
 
 pub const LUCET_MODULE_SYM: &str = "lucet_module";
 
--- a/lucet-module-data/src/module.rs
+++ b/lucet-module-data/src/module.rs
@@ -4,6 +4,7 @@ use crate::tables::TableElement;
 
 pub const LUCET_MODULE_SYM: &str = "lucet_module";
 
+/// Module is the exposed structure that contains all the data backing a Lucet-compiled object.
 #[derive(Debug)]
 pub struct Module<'a> {
     pub module_data: ModuleData<'a>,
@@ -11,6 +12,10 @@ pub struct Module<'a> {
     pub function_manifest: &'a [FunctionSpec],
 }
 
+/// NativeData is a serialization-friendly form of Module, in that `ModuleData` is serialized, with
+/// the `module_data_*` fields here referring to that serialized data, while `tables_*` and
+/// `function_manifest_*` refer to the actual tables and function manifest data written in the
+/// binary.
 #[repr(C)]
 #[derive(Debug)]
 pub struct NativeData {
--- a/lucet-module-data/src/module.rs
+++ b/lucet-module-data/src/module.rs
@@ -1,0 +1,23 @@
+use crate::module_data::ModuleData;
+use crate::tables::TableElement;
+use crate::functions::FunctionSpec;
+
+pub const LUCET_MODULE_SYM: &str = "lucet_module";
+
+#[derive(Debug)]
+pub struct Module<'a> {
+    pub module_data: ModuleData<'a>,
+    pub tables: &'a [&'a [TableElement]],
+    pub function_manifest: &'a [FunctionSpec],
+}
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct NativeData {
+    pub module_data_ptr: u64,
+    pub module_data_len: u64,
+    pub tables_ptr: u64,
+    pub tables_len: u64,
+    pub function_manifest_ptr: u64,
+    pub function_manifest_len: u64,
+}
--- a/lucet-module-data/src/module.rs
+++ b/lucet-module-data/src/module.rs
@@ -12,13 +12,12 @@ pub struct Module<'a> {
     pub function_manifest: &'a [FunctionSpec],
 }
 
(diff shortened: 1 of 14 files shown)


awortman-fastly:
in addition to fixing the merge conflict, I think I neglected some error checking and want to document some of this a bit better. so this isn't _done_, but is pretty close to it

awortman-fastly:
Okay! figured out a few straggling issues and now this is ready for proper review!

lucet-module-data/src/module.rs
```diff
@@ -0,0 +1,28 @@
+use crate::functions::FunctionSpec;
+use crate::module_data::ModuleData;
+use crate::tables::TableElement;
+
+pub const LUCET_MODULE_SYM: &str = "lucet_module";
+
+/// Module is the exposed structure that contains all the data backing a Lucet-compiled object.
+#[derive(Debug)]
+pub struct Module<'a> {
```
acfoltzer: :bike: :derelict_house: I anticipate confusion calling this a `Module`, given how pervasively we use that trait name in the runtime. Naturally, I don't have any better ideas off the top of my head :slightly_frowning_face: Maybe something with Lucet in the name to match the symbol?

lucet-module-data/src/module.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub function_manifest: &'a [FunctionSpec],
+}
+
+/// NativeData is a serialization-friendly form of Module, in that `ModuleData` is serialized, with
+/// the `module_data_*` fields here referring to that serialized data, while `tables_*` and
+/// `function_manifest_*` refer to the actual tables and function manifest data written in the
+/// binary.
+#[repr(C)]
+#[derive(Debug)]
+pub struct NativeData {
```
acfoltzer: :bike: :derelict_house: Maybe call this `SerializedModule` or `FlattenedModule`, replacing `Module` potentially with another name, if we decide to change that? `Native` makes me think x86-64 or just "non-wasm" in the context of our project.

lucet-module-data/src/signature.rs
```diff
@@ -76,25 +79,28 @@ impl RawModuleAndData {
...
                 io::ErrorKind::InvalidInput,
-                "module data not found",
+                format!("`{}` symbol not present", LUCET_MODULE_SYM),
             ))?;
-        let module_data_len_symbol_data =
-            Self::symbol_data(&obj_bin, "lucet_module_data_len", true)?.ok_or(io::Error::new(
+
+        // While `module_data` is the first field of the `NativeData` that `lucet_module` points
+        // to, it is a virtual address, not a file offset. The translation is somewhat tricky at
+        // the moment, so just look at the corresponding `lucet_module_data` symbol for now.
```
acfoltzer: Can we add a TODO here or in an issue to resolve this, if there's a path forward?

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -85,63 +81,41 @@ impl DlModule {
...
-                    if is_undefined_symbol(&ptr_err) && is_undefined_symbol(&len_err) {
-                        &[]
-                    } else {
-                        // This is an unfortunate situation. Both attempts to look up symbols
-                        // failed, but at least one is not due to an undefined symbol.
-                        if !is_undefined_symbol(&ptr_err) {
-                            // This returns `ptr_err` (rather than `len_err` or some mix) because
-                            // of the following hunch: if both failed, and neither are undefined
-                            // symbols, they are probably the same error.
-                            return Err(Error::DlError(ptr_err));
```
acfoltzer: I'm so glad this is gone

lucet-module-data/src/module.rs
awortman-fastly: Thinking this through a bit:

On the `lucet-runtime` side, this is used only in `DlModule`, and not public, so it won't leak outside that file. There, I think explicitly naming `lucet_module_data::Module` seems ok to disambiguate.

Outside that use in `lucet-runtime`, I'm not sure it ends up being too confusing, because `lucet_module_data::Module` is only only in `lucet-analyze` at the moment and I don't really expect it to come up elsewhere.

Of course, maybe this reads very differently to someone new to the codebase :(

(One thought I did have and discard was renaming `lucet_runtime_internal::Module` to `lucet_runtime_internal::Executable`, with the name idea being that the runtime can turn an executable into an instance, and an executable is backed by a module or perhaps some other data. I'm not sold on that, but it's the best alternative I could think up)

lucet-module-data/src/module.rs
awortman-fastly: Definitely agree that `Serialized_` for whatever `Module` does or doesn't get renamed to is a better name, either way

lucet-module-data/src/module.rs
acfoltzer: I think "executable" too strongly implies "a program with `main`", but I do like the thought.

And yeah, my main concern is confusion for people who are new to the codebase, but I'm fine with leaving it for now if it's really used in such a limited fashion.

lucet-module-data/src/signature.rs
awortman-fastly: https://github.com/fastly/lucet/issues/245 for the adventurous spirit

PR #244, opened 2019-07-10
Rename "lucet-module-data" to "lucet-module"

awortman-fastly:
Does what it says on the tin; now that `lucetc` produces a `Module` structure that itself contains all the module information, including `ModuleData`, this crate now describes all the structure for a Lucet module, *in addition* to the part that `lucet_module_data` was named for.


lucet-module/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
-name = "lucet-module-data"
+name = "lucet-module"
 version = "0.1.1"
-description = "A structured interface for Lucet module data and metadata"
```
awortman-fastly: This and `lucet-module-data/src/lib.rs` are the only non-rename changes. Figured updating these to match the name is appropriate

PR #248, opened 2019-07-12
Install wasm-opt and wasm-reduce, use wasm-opt for benchmarking wasm code

jedisct1:


benchmarks/shootout/Makefile
```diff
@@ -60,10 +64,13 @@ build/lucet/shootout/%.o: $(SHOOTOUT)/%.c
...
 	$(WASI_CC) $(COMMON_CFLAGS) -c $^ -o $@
 
-build/lucet/module.wasm: $(patsubst %.c, %.o, $(addprefix build/lucet/shootout/, $(notdir $(SHOOTOUT_SRCS))))
+build/lucet/module.wasm.unoptimized: $(patsubst %.c, %.o, $(addprefix build/lucet/shootout/, $(notdir $(SHOOTOUT_SRCS))))
 	@mkdir -p $(@D)
 	$(WASI_CC) $^ -o $@ -nostartfiles -Wl,--no-entry -Wl,--export-all
 
+build/lucet/module.wasm: build/lucet/module.wasm.unoptimized
+	$(WASM_OPT) -mvp --disable-mutable-globals -O4 -o $@ $^
+
```
pchickey: I would prefer if we benchmarked a separate target that uses wasm-opt, so that we could see the difference.

PR #252, opened 2019-07-17
Supprt counting executed wasm operations

awortman-fastly:
This sketches out support for counting the number of wasm operations executed through an `Instance::run` call.

The new location that that count is tracked in also made a good excuse to refactor how we represent the sliver of data that lucetc and lucet-runtime must cooperate on, as well as adding docs about lucet-runtime's choices around memory organization, so that's here too :tada:


lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -173,6 +174,36 @@ impl Drop for InstanceHandle {
...
+///
+/// This layout allows modules to tightly couple to a handful of fields related to the instance,
+/// rather than possibly requiring compiler-side changes (and recompiles) whenever `Instance`
+/// changes.
+///
+/// It also obligates `Instance` to be immediately followed by the heap, but otherwise leaves the
+/// locations of the stack, globals, and any other data, to be implementation-defined by the
+/// `Region` that actually creates `Slot`s onto which `Instance` are mapped.
+/// For information about the layout of all instance-related memory, see the documentation of
+/// `./src/region/mmap.rs#MmapRegion`.
```
awortman-fastly: I need to look up how to do proper doc links, unless someone knows off the top of their head?

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -531,24 +573,38 @@ impl Instance {
...
-                as *mut *const i64) = globals_ptr;
+            let implicits_ptr = (self as *mut _ as *mut u8)
+                .offset((HOST_PAGE_SIZE_EXPECTED - mem::size_of::<InstanceRuntimeData>()) as isize)
+                as *mut InstanceRuntimeData;
+            mem::transmute::<*mut InstanceRuntimeData, &mut InstanceRuntimeData>(implicits_ptr)
         }
     }
 
+    #[inline]
+    pub fn get_globals_ptr(&self) -> *mut i64 {
```
awortman-fastly: These are `pub fn` in the `// Private API` part of `impl Instance { }` - are they lost stragglers or were they there intentionally? @acfoltzer ?

lucetc/src/function.rs
```diff
@@ -46,7 +48,9 @@ impl<'a> FuncInfo<'a> {
             let vmctx = self.get_vmctx(func);
             let global_base_value = func.create_global_value(ir::GlobalValueData::Load {
                 base: vmctx,
-                offset: GLOBAL_BASE_OFFSET.into(),
+                offset: (-(std::mem::size_of::<InstanceRuntimeData>() as i32)
```
awortman-fastly: I went back and forth a bit on `-size_of() + offset` vs `offset - size_of()` and I think the former reads a bit more nicely. Not super happy with how it reads either way.

lucetc/src/function.rs
```diff
@@ -302,4 +306,136 @@ impl<'a> FuncEnvironment for FuncInfo<'a> {
...
+        let op_cost = match op {
+            // Opening a scope is a syntactic operation, and free.
+            Operator::Block { .. } |
+            // These do not add counts, see above comment about return/unreachable
+            Operator::Unreachable |
+            Operator::Return => 0,
+            // Call is quick
+            Operator::Call { .. } => 1,
+            // but indirect calls take some extra work to validate at runtime
+            Operator::CallIndirect { .. } => 2,
```
awortman-fastly: Most wasm operations are either syntactic (like Block and End) or have some close correspondence with a single-ish machine instruction. `CallIndirect` and `BrTable` involve several operations when translated to clif and later native code, so I count them twice to represent that. Totally arbitrary otherwise.

lucetc/src/function.rs
```diff
@@ -302,4 +306,136 @@ impl<'a> FuncEnvironment for FuncInfo<'a> {
         let inst = pos.ins().call(mem_size_func, &[vmctx]);
         Ok(*pos.func.dfg.inst_results(inst).first().unwrap())
     }
+
+    fn before_translate_operator(
```
awortman-fastly: This entire function is one of the two parts of the `cranelift-wasm` side of this change - adding `before_translate_operator` and `after_translate_operator` functions to `FuncEnvironment` to support cranelift users prodding the instruction stream as it's translated. That's a separate PR and will have to land before this one.

lucetc/src/function.rs
```diff
@@ -78,6 +84,138 @@ impl<'a> FuncInfo<'a> {
...
+        fn flush_counter(environ: &mut FuncInfo, builder: &mut FunctionBuilder) {
+            if environ.op_offsets.last() == Some(&0) {
+                return;
+            }
+            let instr_count_offset: ir::immediates::Offset32 =
+                (-(std::mem::size_of::<InstanceRuntimeData>() as i32)
+                    + offset_of!(InstanceRuntimeData, instruction_count) as i32)
+                    .into();
+            let vmctx_gv = environ.get_vmctx(builder.func);
+            let addr = builder.ins().global_value(environ.pointer_type(), vmctx_gv);
```
awortman-fastly: This ends up inserting a clif sequence along the lines of
```
let instruction_count_ptr: &mut u64 = vmctx.instruction_count;
let mut instruction_count: u64 = *instruction_count_ptr;
instruction_count += 6;
*instruction_count_ptr = instruction_count;
```
which is a bit verbose, but should be straightforward for Cranelift to optimize down a bit.

this _actually_ turns into sequences like
```
mov rdx, qword [rbp + 8] ; load spilled vmctx
mov rdx, qword [rdx - 8] ; load instruction_count
add rdx, 6 ; update count
mov rbx, qword [rbp + 8]; load spilled vmctx
mov qword [rbx - 8], rdx ; store instruction count
```
which is again a bit worse.

To improve this we'll have to look forward to live-range splitting, count instructions using a pinned register, or use a different strategy to count (which is subject to the same live-range splitting needs).

live-range splitting helps because the vmctx loads shouldn't happen anymore (right now they happen because cranelift must spill `rdi`, and afterward the loads happen from the spilled location), and other long-lived variables will probably not spill in the same way without significant register pressure. That means in a function-local context we can probably store `instruction_count_ptr` in a local, and this should all reduce to `add [instruction_count_ptr], 6`.

The other option, counting into a pinned register, would probably be faster, but seems like more work to implement, and deeper changes Cranelift-side. I'd like to measure the impact of this instrumentation after live-range splitting has a chance to help, so we can figure out if we want to do that instead.

lucet-runtime/lucet-runtime-internals/src/instance.rs
acfoltzer: Ah, I actually missed this when we made the change away from explicit padding. These should be private.

lucetc/src/function.rs
acfoltzer: Live-range splitting should also help the register allocator do better for general-purpose code, right? Not just for our instruction counting special case

acfoltzer: > This ends up inserting a clif sequence along the lines of
> 
> ```
> let instruction_count_ptr: &mut u64 = vmctx.instruction_count;
> let mut instruction_count: u64 = *instruction_count_ptr;
> instruction_count += 6;
> *instruction_count_ptr = instruction_count;
> ```

It would be nice to have this as a comment

lucetc/src/function.rs
```diff
@@ -78,6 +84,138 @@ impl<'a> FuncInfo<'a> {
...
+            | Operator::BrTable { .. } => {
+                flush_counter(self, builder);
+            }
+            Operator::End => {
+                // We have to be really careful here to avoid violating a cranelift invariant:
+                // if the next operation is `End` as well, this end will have marked the block
+                // finished, and attempting to add instruction to update the instruction counter
+                // will cause a panic.
+                //
+                // We can avoid that case by ensuring instruction counts are flushed at the *entry*
```
acfoltzer: Is that what we're doing here, or is that something we need to do in the future? It looks like we're just calling `flush_counter` in the same way we do for the other basic block terminators

lucetc/src/function.rs
acfoltzer: I love how minimal the changes in cranelift are to accommodate this new feature

lucetc/src/lib.rs
```diff
@@ -98,6 +99,8 @@ pub trait LucetcOpts {
     fn with_verify(self) -> Self;
     fn sign(&mut self);
     fn with_sign(self) -> Self;
+    fn count_instructions(&mut self);
```
acfoltzer: These should take a `bool` (and now that I'm looking at it, the sign and verify methods should too)

lucetc/src/options.rs
```diff
@@ -244,6 +247,12 @@ impl Options {
                      .takes_value(true)
                      .help("Path to the secret key to sign the object file. The file can be prefixed with \"raw:\" in order to store a raw, unencrypted secret key")
             )
+            .arg(
+                Arg::with_name("count_instructions")
+                    .long("--count_instructions")
```
acfoltzer: the arg should have a hyphen rather than an underscore; I dunno why our convention in this file has underscores in the name, but it's not as fun to type

lucetc/src/function.rs
```diff
@@ -6,27 +6,31 @@ use cranelift_codegen::cursor::FuncCursor;
...
     runtime_funcs: HashMap<RuntimeFunc, ir::FuncRef>,
 }
 
 impl<'a> FuncInfo<'a> {
-    pub fn new(module_decls: &'a ModuleDecls<'a>) -> Self {
+    pub fn new(module_decls: &'a ModuleDecls<'a>, count_instructions: bool) -> Self {
         Self {
             module_decls,
+            count_instructions,
+            op_offsets: vec![0],
```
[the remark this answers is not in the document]
awortman-fastly: one better, how's `scope_costs`? it was initially `op_offsets` because it started counting offsets of wasm ops for any given scope, but now it's really a stack of counters for cost when we enter a new scope

I actually think because of that cost flushing behavior in the other comment this can just be a single integer, I'm gonna think through that..

lucetc/src/function.rs
awortman-fastly: Yup, live-range splitting will help generally, but because of how instruction counting is implemented here it'll be enough of a change I think it's worth noting

lucetc/src/function.rs
```diff
@@ -6,27 +6,31 @@ use cranelift_codegen::cursor::FuncCursor;
 use cranelift_codegen::entity::EntityRef;
 use cranelift_codegen::ir::{self, InstBuilder};
 use cranelift_codegen::isa::TargetFrontendConfig;
+use cranelift_frontend::FunctionBuilder;
 use cranelift_wasm::{
     FuncEnvironment, FuncIndex, GlobalIndex, GlobalVariable, MemoryIndex, SignatureIndex,
     TableIndex, WasmError, WasmResult,
 };
+use lucet_module::InstanceRuntimeData;
+use memoffset::offset_of;
```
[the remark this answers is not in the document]
awortman-fastly: no more or less than in other places we use it, but I think the answer is still yes. I don't think there's a non-UB implementation of this, so the alternative is associated consts for field offsets here. Personally, I'm banking on a well-behaved `offset_of` existing before this gets exploited in some way by rustc, and that being easier to work with than manually updating field constants, but that's shaky ground and I'm a small nudge away from making hte change.

lucetc/src/function.rs
awortman-fastly: oops, good eye. this is a relic from my initial attempt at this, and currently will call `flush_counter`, but the counter is always 0 (so no instructions are added, and no cranelift invariants are violated)

I'll reword this and change the call to an assert

lucetc/src/function.rs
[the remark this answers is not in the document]
acfoltzer: I think manual updates would definitely be brittle, while the UB threat would probably (?) only be an issue when updating Rust versions. So I'm inclined to stick with it for now

lucetc/src/function.rs
awortman-fastly: (several hours of thought later) I was wrong - the counter is _definitely_ not always 0 at this point, and I reworded this comment and added an example of what it's talking about for reference.

lucet-spectest/src/script.rs
```diff
@@ -67,8 +67,14 @@ impl ScriptEnv {
...
     pub fn instantiate(&mut self, module: &[u8], name: &Option<String>) -> Result<(), ScriptError> {
         let bindings = bindings::spec_test_bindings();
-        let compiler = Compiler::new(module, OptLevel::Fast, &bindings, HeapSettings::default())
-            .map_err(program_error)?;
+        let compiler = Compiler::new(
+            module,
+            OptLevel::Fast,
+            &bindings,
+            HeapSettings::default(),
+            false,
```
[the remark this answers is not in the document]
awortman-fastly: I flipped this to true because why not, and it turns out this as-is was causing a lot of spectest issues.

There's a lot of dead code in the spec tests that also cause the instrumentation logic to add code where Cranelift expects there to be no more. I've added some additional logic to figure out when wasm ops are dead code, and the spec tests all pass now, but it also pretty much duplicates the dead code logic out of Cranelift. Passing along the cranelift-wasm translation state exposes a lot of translation innards and really doesn't make for a great interface..

lucetc/src/function.rs
```diff
@@ -16,10 +16,117 @@ use memoffset::offset_of;
...
+    pub fn open_scope(&mut self) {
+        self.scopes.push(
+            ScopeControlFlow::Linear(
+                // the scope still has reachable code if the parent scope did not terminate
+                !self.scopes.last().unwrap().terminates()
+            )
+        )
+    }
+
+    pub fn open_conditional(&mut self) {
```
acfoltzer: this seems subtle enough to need a comment

lucetc/src/function.rs
```diff
@@ -16,10 +16,117 @@ use memoffset::offset_of;
...
+        if let Some(ScopeControlFlow::Conditional(_, ref mut alternate)) = self.scopes.last_mut() {
+            match alternate {
+                None => {
+                    *alternate = Some(true);
+                },
+                Some(false) => {
+                    // the parent scope was unreachable, we're walking through dead code anyway.
+                },
+                Some(true) => {
+                    // this is the second alternate branch added on an if? that's possible.
```
acfoltzer: typo?

lucetc/src/function.rs
```diff
@@ -119,98 +227,147 @@ impl<'a> FuncInfo<'a> {
...
+                | Operator::Else
+                | Operator::Br { .. }
+                | Operator::BrIf { .. }
+                | Operator::BrTable { .. } => {
+                    flush_counter(self, builder);
+                }
+                Operator::End => {
+                    // We have to be really careful here to avoid violating a cranelift invariant:
+                    // if the next operation is `End` as well, this end will have marked the block
+                    // finished, and attempting to add instruction to update the instruction counter
```
acfoltzer: typo

lucetc/src/function.rs
```diff
@@ -119,98 +227,147 @@ impl<'a> FuncInfo<'a> {
...
+            }
+            Operator::If { .. } => {
+                self.scope_costs.push(0);
+                self.scope_tracker.open_conditional();
             }
-            Operator::Block { .. } | Operator::Loop { .. } | Operator::If { .. } => {
-                // open a new scope
-                self.op_offsets.push(0);
+            Operator::Else => {
+                self.scope_tracker.add_alternate_branch();
```
acfoltzer: do we need to push onto `scope_costs` here as well? it looks odd that this is the only case that doesn't have a matching pair of `scope_costs` and `scope_tracker` manipulation

lucetc/src/function.rs
```diff
@@ -16,10 +16,117 @@ use memoffset::offset_of;
...
+            ScopeControlFlow::Conditional(fallthrough, Some(alternate)) => {
+                // an if terminuates if the fallthrough terminates and the alternate, when present,
+                // terminates
+                !fallthrough && !alternate
+            }
+        }
+    }
+}
+
+struct ScopeTracker {
```
acfoltzer: would appreciate some guide-level documentation of this machinery, including an explanation of the interrelationship between  "terminates" and "reachable" in this setting

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -36,18 +36,20 @@ use std::sync::{Arc, Mutex, Weak};
 /// 0x0XXX: |  .inst_count = 0x0000 |
 /// 0x1000: +-----------------------+ <-- Heap, and `lucet_vmctx`
 /// 0x1XXX: |                       |
-/// 0xXXXX: ~  .......heap.......   ~
+/// 0xXXXX: ~  .......heap.......   ~ // heap size is coverned by limits.heap_address_space_size
 /// 0xXXXX: |                       |
-/// 0xN000: +-----------------------| <-- Stack (at 0x1000 + max_heap_size)
+/// 0xN000: +-----------------------| <-- Stack (at limits.heap_address_space_size + PAGE_SIZE)
```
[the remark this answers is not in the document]
awortman-fastly: hum. I'll reword this. My intent was for this to count with respect to the beginning of the allocation, which puts the stack is at `limits.heap_address_space_size + <INSTANCE_DATA>`. That _is_ one page, and how it came to be as it is in the comment - didn't mean to suggest that there's a page between the heap and the stack. How does a similar note about the heap starting at `<INSTANCE_SIZE>`, and changing this to `(at heap_start + limits.heap_address_space_size)` sound?

lucetc/src/function.rs
awortman-fastly: missing `not`, yeah

lucetc/src/function.rs
awortman-fastly: we do not, but only because `Else` causes the counter to be flushed, so we know it will be 0 entering the body of `Else`'s block. I'll note that as why there's nothing with `scope_costs` here.

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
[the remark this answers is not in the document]
acfoltzer: That sounds good!

awortman-fastly:
when this last build passes (and this PR has a :heavy_check_mark: again) this is good to merge - I've given up any hope of parceling this into meaningful commits and plan on just squashing :(

PR #260, opened 2019-07-31
Refactor instance State and associated C APIs

acfoltzer:
These changes reduce the number of times that our types must be `Clone`able simply in order to keep a copy around in the `State` while also returning a result. This had cascading implications, like having to store termination and yield values in `Arc`s, which are all now gone. The only restrictions on these values now is `Any + 'static`, which actually reflects the semantics we care about.

The downside to this is there are now more complicated invariants about states, but those are now noted in docs on the struct.

Accommodating these changes in the C API meant ripping out the explicit `lucet_instance_state()` function and replacing it with an out-parameter on run/resume. `lucet_state` is now `lucet_result`, and is more analogous to `Result<RunResult, Error>`. Along the way, the pointers to strings for file and symbol names were replaced with fixed-length buffers so that we don't need a free function for `lucet_result`. This all makes the C API nicer, in addition to supporting the `State` refactoring.


lucet-runtime/lucet-runtime-internals/src/c_api.rs
```diff
@@ -218,117 +224,113 @@ pub struct CYieldedVal {
...
+        Terminated,
+        Errored,
     }
 
     #[repr(C)]
     #[derive(Clone, Copy)]
-    pub union lucet_state_val {
+    pub union lucet_result_val {
         pub returned: lucet_val::lucet_untyped_retval,
-        // no meaning to this boolean, it's just there so the type is FFI-safe
```
awortman-fastly: What was the deal with this, I'd have thought that the union would be FFI-safe by virtue of `#[repr(C)]`?

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -800,41 +800,58 @@ impl Instance {
...
-                        // If there is no C-style fatal handler, or if it (erroneously) returns,
-                        // call the Rust handler that we know will not return
-                        (self.fatal_handler)(self)
-                    } else {
-                        // leave the full fault details in the instance state, and return the
-                        // higher-level info to the user
-                        Err(Error::RuntimeFault(details.clone()))
-                    }
+                //
+                // FIXME after lucet-module is complete it should be possible to fill this in without
```
awortman-fastly: since the line is highlighted: lucet-module should have all the information for this, it's just a `lucet-runtime-internals` tweak at this point - `lookup_trapcode` looks up faulting function from `rip` purely from module data, but the `addr_details` replacement would just want the name of the faulting function. I dunno if we have the guest module's file name though?

I see this is just a comment that moved up a bit, so this isn't to detract from the changes at all

lucet-runtime/lucet-runtime-internals/src/c_api.rs
acfoltzer: "safe" is maybe the wrong word; this is because C doesn't allow for unions with empty variants

lucet-runtime/lucet-runtime-internals/src/instance.rs
acfoltzer: We should totally make that change once we have the guest filename

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -800,41 +800,58 @@ impl Instance {
...
+        // * returned: state should be `Running`; transition to Ready and return a RunResult
+        // * yielded: state should be `Yielded`; take the yielded value and return a RunResult
+        // * trapped: state should be `Fault`; populate details and return an error or call a hanlder as appropriate
+        // * terminated: state should be `Terminated`; take the termination details and return as an Err
         // * function body returned: set state back to `Ready` with return value
+        //
+        // The state should never be `Ready` at this point
 
-        match &self.state {
+        match &mut self.state {
```
awortman-fastly: (the `mut` here is what got me thinking about it, so the comment is here)

summarizing our conversation: especially now that the C API isn't fiddling with `State` directly, it's not an additional complexity for users to split `State` into more granular variants to distinguish between `Yielded` (returning a result) and `Expecting` (to restart with an indicated value) - similarly for `Terminated` -> `TerminatedWith` (some fault information) and `Terminated` (the "this instance must be restarted to execute"). This would let us avoid the `Option<_>` and `take()` by just taking the enum fields for an eventual `RunResult` and directly resetting `state` to the appropriate instance state.

Parceling `State` into `payload`, `can execute`, and `must be reset` states might be interesting conceptually or for documentation reasons but probably not interesting to work into the enum.

acfoltzer: Done in 56970cb; thanks for the good suggestion :+1:

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -851,12 +855,12 @@ impl Instance {
...
                     // higher-level info to the user
-                    Err(Error::RuntimeFault(details.clone()))
+                    Err(Error::RuntimeFault(details))
                 }
             }
-            State::Ready { .. } => {
-                panic!("instance in Ready state after returning from guest context")
-            }
+            State::Ready | State::Terminated | State::Yielded { .. } | State::Transitioning => Err(
+                lucet_format_err!("\"impossible\" state found in `swap_and_return()`: {}", st),
```
awortman-fastly: `"impossible"` :grin:

PR #269, opened 2019-08-20
Improve lucetc error messages

awortman-fastly:
Currently, given malformed wasm text like
```
(module
  (type (func (param i32 i32 i32 i32) (result i32)))
  (func $write (import "wasi_unstable" "fd_write") (type 0)) 
  (func $strlen (param i32) (result i32)
    (local i32 i32 i32)
    locaf l.get 0   ;; <----- this is the important line. "locaf" is very much not a wasm opcode!
    local.set 1
    local.get 1
    local.get 0
    i32.sub)
  (memory (;0;) 2)
  (export "memory" (memory 0)) 
  (data (i32.const 0) "0123456789\00a\00\00\00\00")
  (data (i32.const 16) "strlen(\"a\") is: \00\n")
  ;; iovec struct for eventually printing "strlen("a") is: ""
  (data (i32.const 64) "\10\00\00\00\12\00\00\00")
  (func $_start (export "_start")
    i32.const 32
    (call $strlen (i32.const 11))
    ;; fix up line to print by loading the digit at result of strlen (should be in [0,9])
    i32.load8_u
    i32.store8
    (call $write (i32.const 1) (i32.const 64) (i32.const 1) (i32.const 0)) 
    drop
  )
)
```
lucetc's error message is unhelpful:
```
ErrorMessage { msg: "Input is neither valid WASM nor WAT" }

Input
```

this adjusts lucetc to give nicer error messages and try to be more informative about the underlying cause:
```
Error: Input. invalid WAT input, parse error:
test.wast:6:5: error: unexpected token locaf, expected ).
    locaf l.get 0
    ^^^^^
test.wast:6:11: error: unexpected token l.get.
    locaf l.get 0
          ^^^^^
```

For comparison on non-wabt errors, the non-typo'd version of the above program would have
```
LucetcError { inner: ErrorMessage { msg: "Unknown module for symbol `wasi_unstable::fd_write`" }
                                                  
Translating module }
```
without a `bindings.json`. Now that error looks like
```
Error: Translating module. Unknown module for symbol `wasi_unstable::fd_write`
```

this is a draft because when I made `wabt-rs`'s `ErrorKind` public, I neglected to ensure the field in the reported error is accessible. So it's not. oops. Just to check the underlying error is what I wanted (it is!) I snuck in a transmute, but that will not stick around long

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1014,7 +1014,7 @@ dependencies = [
  "parity-wasm 0.38.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "target-lexicon 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "wabt 0.7.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wabt 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "wasmonkey 0.1.8",
  "wasmparser 0.39.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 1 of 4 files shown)


jedisct1:
Great! 👍

PR #270, opened 2019-08-21
Upgrade to cranelift 0.40.0

pchickey:


lucetc/src/module.rs
```diff
@@ -130,9 +130,10 @@ impl<'a> ModuleInfo<'a> {
         sig: ir::Signature,
     ) -> (UniqueFuncIndex, SignatureIndex) {
         let new_sigidx = SignatureIndex::from_u32(self.signature_mapping.len() as u32);
-        self.declare_signature(sig);
+        self.declare_signature(sig).expect("declaring signature");
```
acfoltzer: Could this return a `Result<_, LucetcError>` instead? I know we have the definition right below us that returns `Ok(())`, but it'd be nice to not have to worry about getting bitten in the future.

pchickey: good idea. done.

PR #271, opened 2019-08-22
Use philipc's `object::write` to emit object files

pchickey:


jedisct1:
This is producing an invalid object on MacOS:

```text
ld: warning: No version-min specified on command line\nld: illegal text-relocation to \'guest_table_0\' in /var/folders/gq/pbvv220x79156lw2f_f22r5w0000gn/T/lucetcfv14BM/tmp.o from \'guest_table_0\'
```

## How to reproduce

```c
int main(void)
{
    return 0;
}
```

```sh
clang --target=wasm32-wasi --sysroot /opt/wasi-libc -O2 -s -o a.wasm a.c

lucetc-wasi a.wasm
```

WASM file and resulting object files:

[a.wasm.gz](https://github.com/fastly/lucet/files/3531300/a.wasm.gz)
[a.obj.gz](https://github.com/fastly/lucet/files/3531298/a.obj.gz)

philipc:
Looks like I moved too much into `__TEXT/__const` (it used to all be in `__DATA/__data`). I've added https://github.com/gimli-rs/object/pull/120/commits/1f8a490163135eef2fd51293efc3bf04afee6940 which should let us use `__DATA/__const` instead for const data that has relocations. This will require a small change in cranelift-object too.

jedisct1:
Pretty ugly,  but the following diff over `object` rev a2758023ea001c5e9fba79ebc476bd81b0bda8fa seems to help `lucetc` build valid Mach-O files.

https://gist.github.com/4eccc9f79d8db408e1c1717a4d446d49

BTW, that file can come in handy: https://opensource.apple.com/source/ld64/ld64-128.2/src/ld/parsers/macho_relocatable_file.cpp.auto.html

philipc:
@jedisct1 Thanks for figuring out those fixes!

Can you explain more about the underscore prefix, or point to some documentation about it? I see that [faerie writes the underscore for every symbol](https://github.com/m4b/faerie/blob/f88634e0c4d192e6a3cddb1f1ec44cb3facfb1f8/src/mach.rs#L873), but you've only added for externs.

I might have to come up with a way of making the underscore an option in `add_symbol`, since always adding it during writing is going to break the `objcopy` round trip (unless I remove it again when reading?).

jedisct1:
Hi @philipc ,

The [Mach-O documentation](http://math-atlas.sourceforge.net/devel/assembly/MachORuntime.pdf) says:

> The name of a symbol representing a function that conforms to standard C calling conventions is the name of the function with an underscore prefix. Thus, the name of the symbol representing the function main would be _main.

So I guess it indeed should also be applied to non-exported functions.

COFF symbols also need leading underscores.

jedisct1:
I don't think any other calling convention has even been used on MacOS.

LLVM [skips leading underscores unconditionally](https://llvm.org/doxygen/SymbolizableObjectFile_8cpp_source.html) on Mach-O files.

So maybe it would make sense to heave the `read` interface remove these underscores, so we have symmetry between the `read` and `write` interfaces.

philipc:
I've opened https://github.com/gimli-rs/object/pull/127 to fix those issues, if you could review that.

I chosen to make the mangling optional, and disable this for objcopy. That LLVM link is only for the symbolizer, which is a bit higher level than the API that the object crate wants to provide. I've based this on [this LLVM code](https://github.com/llvm-mirror/llvm/blob/89c6befa0ad772ed5259436fd69405346ee97002/include/llvm/IR/DataLayout.h#L303-L315).

PR #286, opened 2019-09-06
Fixes #276 - Allow use of aligned heaps

### LINKED ISSUE #276: Allow use of 4GB aligned heaps to optimize performance of pointer swizzling
**Context**

**Problem**
Since our use case involves sandboxing libraries and not entire applications, data structures may be shared between application and the wasm sandboxed library. Such data structures may contain pointer fields. Given that the representation of pointers differ between the application (which uses simple 64-bit addresses) and the wasm sandboxed library (which uses 32-bit offsets from the sandbox memory base), we need to convert/swizzle pointers at the appropriate locations. 

One particularly efficient and elegant approach to swizzle pointers automatically using only C++ operator overloading (if you are curious, this is implemented [here](https://github.com/PLSysSec/rlbox_lucet_sandbox/blob/master/include/rlbox_lucet_sandbox.hpp#L533)) requires the heap to be aligned according to its size. For instance, a 4GB heap would need to be 4GB aligned, a 1GB heap would need to 1GB aligned etc.

**Actions**
@pchickey mentioned that I could submit a patch for this alignment. I already have a prototype for the same and can clean it up and submit this shortly. The plan is to implement this under an alternate constructor of MmapRegion.

Please let me know if you have any thoughts/concerns.

shravanrn:


shravanrn:
> Before we merge, we'll need to add some tests for the new interface. 

I had a look, but as you said, it was a bit tricky to figure out where and how to add the tests :sweat_smile: 

> I can give you guidance on navigating our test suite.

I'm happy to take a stab at it, if you give me some details... Could you let me know where the right place in the repo would be for these tests? Any additional guidelines would be appreciated as well.

> It might be a little bit before I can get to doing that, so if merging the PR quickly is important 

 If this is a matter of a couple of weeks, I'm happy to wait as well :smiley:

acfoltzer:
> If this is a matter of a couple of weeks, I'm happy to wait as well

A couple weeks sounds about right. I'd also like to use this as an opportunity to write some better docs about how our test suite is laid out. We have the start of docs on [the wiki](https://github.com/fastly/lucet/wiki/Tests-and-benchmarks) but that page is not sufficient to answer your questions.

acfoltzer:
@shravanrn sorry for the delay on this. I've added more detail to the [wiki page](https://github.com/fastly/lucet/wiki/Tests-and-benchmarks) that hopefully helps answer your questions.

To make the advice there specific to this use case, I would put unit tests in `/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs` or a new child module, depending on size; integration tests should probably go in `/lucet-runtime/tests`.

Another idea that could help get a lot of test coverage quickly without too much new typing would be to create a wrapper type, something like `AlignedMmapRegion`, that would create an underlying `MmapRegion` with a particular alignment. Then, where you see the test macros instantiated with `MmapRegion`, you could add another instantiation with `AlignedMmapRegion`. I think it'd still be worth writing some tests that are specific to the alignment APIs, though.

shravanrn:
That's great! Thanks for the info. This next week unfortunately is pretty busy, but I should be able to get some tests written up the following week.

shravanrn:
Hi, 
Sorry for the delay here. I got pulled away on some other stuff and have finally resumed the lucet changes. I have rebased and added some tests. Does this look ok?

shravanrn:
@acfoltzer @pchickey - just following up about this

.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
...
+            "setupCommands": [
+                {
+                    "description": "Enable pretty-printing for gdb",
+                    "text": "-enable-pretty-printing",
+                    "ignoreFailures": true
+                }
+            ]
+        }
+    ]
+}
```
pchickey: this shouldn't be checked in - try adding .vscode to your global gitignore

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -343,7 +398,137 @@ impl MmapRegion {
...
+        let kb: usize = 1024;
+        let mb: usize = 1024 * kb;
+
+        struct TestProps {
+            pub mem_size: usize,
+            pub mem_align: usize,
+            pub offset: usize,
+        };
+
+        let mut tests = Vec::new();
```
pchickey: Can you write this as `vec![ TestProps{ whatever }, TestProps { whatever }]?;

.vscode/launch.json
shravanrn: Oops... added this by mistake. Have removed it now.

shravanrn:
Have made requested changes.

PR #289, opened 2019-09-09
Fixes #288 - Remove cranelift-codegen from the dependencies of the lucet-runtime components such as lucet-module

### LINKED ISSUE #288: lucet-module, which is part of the runtime pulls in all of cranelift-codegen which is large
**Context**
This is part of a series of bugs that I spoke to @tyler @pchickey about. We are currently using Lucet to sandbox libraries in C++ applications. The idea behind this is that using a wasm sandboxed version of the library allows ensuring that a memory safety issue in the library does not automatically result in a memory safety vulnerability in the full application. One of the consumers of this work is the Firefox web browser.

**Problem**
lucet-module, which is part of the runtime pulls in all of cranelift-codegen. So, from code comments [here](https://github.com/fastly/lucet/blob/master/lucet-module/Cargo.toml#L14), this is already on people's radar. I am just opening this bug to express that we have an interest in solving this. 

**Actions**
Could you let me know if this is a change that is being planned anytime soon internally, or if it would be better if I work on a patch that can be accepted? (cc @froydnj)

shravanrn:


shravanrn:
wasn't that difficult after all  :smiley:

PR #290, opened 2019-09-10
Fixes #279 - Allow minisign dependency  to be optionally excluded at compile time

### LINKED ISSUE #279: Add a feature flag to allow excluding crypto dependencies in lucet-module-data
**Context**

**Problem**
lucet-module-data provides the ability check the signature of modules prior to loading. However for the library sandboxing case, the libraries are packaged and sent together and thus have their own signing/certificate mechanisms. Thus dependencies like `minisign`, `pbkdf2`, `subtle` etc. are all unused. 

In order to simplify the dependency story here, would it make sense to add a feature flag that allows us to exclude the signing functionality? This feature flag would be enable inclusion of signing code by default, but would provide to us the ability to remove the functionality (and dependencies).

**Actions**
Please let me know your thoughts on this. If this sounds good, please also let me know if this is a change that can happen internally or if it would be better if a work on a patch that can be accepted?

shravanrn:


lucetc/src/signature.rs
```diff
@@ -74,6 +75,12 @@ pub fn verify_source_code(
...
 
 // Sign the compiled code
+#[cfg(feature = "signature_checking")]
 pub fn sign_module<P: AsRef<Path>>(path: P, sk: &SecretKey) -> Result<(), Error> {
     ModuleSignature::sign(path, sk).map_err(|e| e.into())
 }
+
+#[cfg(not(feature = "signature_checking"))]
+pub fn sign_module<P: AsRef<Path>>(_path: P, _sk: &SecretKey) -> Result<(), Error> {
+    Ok(())
```
jedisct1: If that public function is called but the signature feature is not available, how about throwing an error instead?

shravanrn: Yup, makes sense. Will fix... Looks like I forgot to run the cargo fmt as well on the last submission as well.

shravanrn:
Just following up on this... Does this look good to merge? :smiley: 

It looks like the travis CI failure seems to be due to an unrelated issue, but I may be wrong? If so, is there a way to kick of a new build?

pchickey:
I kicked off a new build

shravanrn:
Looks like this needs some investigation. Will follow up shortly.

shravanrn:
Yay! CI checks are finally happy :smiley:

shravanrn:
Just following up on this. This is ready to merge now.

shravanrn:
Actually, on speaking to Mozilla folk, it looks like we won't be needing this going forward. So cancelling this PR

Issue #291, opened 2019-09-10
error with lucet-runtime


lucifer-cpu:
error with lucet runtime

```
let module = DlModule::load("output.so").unwrap();
let region = MmapRegion::create(1, &Limits::default()).unwrap();
```

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DlError(Custom { kind: Other, error: "output.so: undefined symbol: __wasi_fd_prestat_get" })'
```

pchickey:
It looks like the lucet-wasi host functions didn't get linked into the executable you made. There are two things required to ensure that:

1. Call the runtime and lucet-wasi `ensure_linked` functions somewhere in your executable. The calls don't actually do anything, so putting them at the beginning of `main` is fine. Just has to be in reachable code.
https://github.com/fastly/lucet/blob/6ae5a401bfd2f3f9d9160623fd7c1312ef312de9/lucet-wasi/src/main.rs#L37-L40

2. You need to configure the linker to make all symbols available to the dynamic linker. The usual way to do this is with a `.cargo/config` file with the following flags:
```
[build]
rustflags = ["-C", "link-args=-rdynamic"]
```

lucifer-cpu:
my executable is in c

lucifer-cpu:
```
rustflags = ["-C", "link-args=-rdynamic"]
``` 
allready

lucifer-cpu:
I am trying to run a project that is already written in c and compiled to wasm.
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ModuleError(IncorrectModule("error loading required symbol `lucet_module_data`: /home/aranajhonny/lab/wasm-quickjs/qjs.so: undefined symbol: lucet_module_data")
```

lucifer-cpu:
Do you have any examples of any complete project written in c that runs with this runtime?

jedisct1:
Can you run it with lucet-wasi?

lucifer-cpu:
![image](https://user-images.githubusercontent.com/55155340/64656195-4b47fa00-d3f4-11e9-80ee-5b9577348a33.png)

awortman-fastly:
are `output.so` and `qjs.so` above both the same file?

lucifer-cpu:
yes

lucifer-cpu:
Maybe I'm doing something wrong, I'm not very expert.

awortman-fastly:
[...]

PR #273, opened 2019-09-03
Pinned heap test

awortman-fastly:
This gets everything in place to use a pinned register for heap accesses via https://github.com/CraneStation/cranelift/pull/960

~There's a subset of this PR that's just updating to match changed Cranelift APIs - I'll pull that out into a subsequent PR~ the `master` I'd branched from was out of date, oops.

PR #294, opened 2019-09-11
And version info to modules

awortman-fastly:
also add a check in lucet-runtime that modules loaded have version
version present, and that the module's version matches the runtime,
rejecting modules that do not

edit: the PR does fewer things, so I fixed the title and message to match


awortman-fastly:
this is a sketch of what's been kicking around in my head, wrote it out partially motivated by #291 today - this also adds a good place to include further codegen quirks like "this module uses pinned heaps". The lack of a good place for that kind of setting is why #273 remains a draft, so I figure this is a good time.

lucet-module/src/version_info.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct VersionInfo {
+    major: u8,
+    minor: u8,
+    patch: u8,
+    reserved: u8,
+    pub capabilities: Capabilities,
+}
+
+impl VersionInfo {
+    pub fn current() -> Self {
```
awortman-fastly: My expectation is that `lucet-runtime`'s version check would always use `VersionInfo::current`, sets supported capabilities, and then does a comparison with whatever we loaded. That comparison function is conspicuously absent here; I think we tend to know what will be breaking change to a module's ABI, and can just bump these fields accordingly, but that's where I'd really like thoughts

lucet-module/src/version_info.rs
```diff
@@ -0,0 +1,92 @@
...
+/// how the module can be loaded, if so requested. The information here describes implementation
+/// details in runtime support for `lucetc`-produced modules, and nothing higher level.
+#[repr(C)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct VersionInfo {
+    major: u16,
+    minor: u16,
+    patch: u16,
+    reserved: u16,
+    version_hash: [u8; 8],
```
acfoltzer: Should note which portion of the hash this is

lucetc/src/output.rs
```diff
@@ -136,10 +136,15 @@ fn write_module(
     function_manifest_len: usize,
     obj: &mut Artifact,
 ) -> Result<(), Error> {
-    let mut native_data = Cursor::new(Vec::with_capacity(NATIVE_POINTER_SIZE * 4));
+    let mut native_data = Cursor::new(Vec::with_capacity(16 + NATIVE_POINTER_SIZE * 4));
```
acfoltzer: Does this have to be a magic number or can we use `size_of`?

lucet-module/src/version_info.rs
awortman-fastly: you mean that it's the first eight ascii characters of the git hash?

lucetc/src/output.rs
awortman-fastly: good point! the entire `with_capacity` argument can and should be `mem::size_of::<SerializedModule>` now that such a type exists.

lucet-module/src/version_info.rs
acfoltzer: Yep

lucet-runtime/tests/version_checks.rs
```diff
@@ -0,0 +1,16 @@
+use lucet_runtime::{DlModule, Error};
+
+#[test]
+pub fn reject_old_modules() {
+    let err = DlModule::load("./tests/version_checks/old_module.so")
```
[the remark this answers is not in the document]
iximeow: Note from future-me to past-me, _what source was this module built from?_

PR #295, opened 2019-09-12
Runtime switch to select between `faerie` and `object`

pchickey:
Taking a different approach to #271 to mitigate risk:

Rather than translating all use of Faerie to Object, this PR makes it a runtime selection between the two, defaulting to Object.

WIP- still trying to divide between the two backends on the right line


pchickey:
I need to set this PR aside for a while, other stuff is higher priority at the moment.

PR #298, opened 2019-09-13
Exit dev shell scripts on first command exiting with non-zero return

### LINKED ISSUE #279: Add a feature flag to allow excluding crypto dependencies in lucet-module-data

### LINKED ISSUE #297: CI allows failed compilation
Referencing this comment https://github.com/fastly/lucet/pull/275#issuecomment-531299667: The CI scripts don't exit when Lucet fails to compile.

ekilmer:
Fixes #297

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,33 @@
 dist: xenial
 
+# Same as version in Dockerfile(s)
+rust: 1.36.0
+
 env:
     - UNOPTIMIZED_BUILD=true
 
 services:
     - docker
 
-script:
-    - ./devenv_run.sh make indent-check test audit
-    - ./devenv_stop.sh
-    - git diff --exit-code
-    - ./devenv_build_toolchain_only.sh
+stages:
+  - lint
+  - build_and_test
+
+jobs:
+  include:
+    - stage: lint
+      language: rust
+      install:
+        - rustup component add rustfmt
+      script:
+        - make indent-check
+        - git diff --exit-code
+    - stage: build_and_test
+      script:
+        - ./devenv_run.sh make test || travis_terminate 1
+        - ./devenv_run.sh make audit || travis_terminate 1
+        - ./devenv_stop.sh
+        - ./devenv_build_toolchain_only.sh
 
 notifications:
     email: false
--- a/devenv_run.sh
+++ b/devenv_run.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-set -e
 
 . "$(dirname ${BASH_SOURCE:-$0})/config.inc"
 
--- a/devenv_run.sh
+++ b/devenv_run.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
 . "$(dirname ${BASH_SOURCE:-$0})/config.inc"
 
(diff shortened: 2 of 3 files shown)


jedisct1:
Hi,

This is causing scripts to immediately stop without providing much information about what happened.
 
In these scripts and depend scripts, we prefer to handle errors explicitly:

```sh
if docker ps -f name='lucet' --format '{{.Names}}' | grep -q '^lucet$' ; then
        echo "container is already running" >&2
        exit 1
fi
```

awortman-fastly:
Yay, it fails! I'm surprised it how it fails though, it seems to fail in `devenv_run.sh` (as it should, I'd say), but Travis proceeds to continue running the CI job? That looks to be a [Travis thing](https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase):

> When one of the build commands returns a non-zero exit code, the Travis CI build runs the subsequent commands as well, and accumulates the build result.

Should we have one script that does the whole build process and run that instead?

I think `set -e` on these scripts gives the failure semantics we want - there isn't error recovery to do if we fail to build the container in the first place, right? `devenv_stop.sh` afterward is messy in the build log, but it seems incorrect to run that when its dependent task, building the container, already failed.

ekilmer:
> Travis proceeds to continue running the CI job? That looks to be a Travis thing:

Ah. I did not realize Travis would do that.

> Should we have one script that does the whole build process and run that instead?

I think it's a matter of preference, but from my experience the CI should test and run whatever actions are most likely to be executed by users of the software. If we do that, then the installation procedure is almost entirely self-documented in the CI (in case the README or Wiki directions fall out of sync).

The `devenv_run.sh` script seems to be top-most script that takes care of creating the Docker image and running the commands in the container, so I think that should be used in CI.

In terms of contributor experience, it would be nice to know that a PR doesn't pass the `indent-check` Makefile target before the `lucet` image is built, since the image takes a long time to build. This indent-check could be added to travis in `before_script`.

Please review the changes in https://github.com/fastly/lucet/pull/298/commits/93e4e6e62c9ab49fd48a2095ff4a3f0f3b0cf6b2 for Travis reorganization. All of this, assuming this commit behaves in Travis how I think it will behave

jedisct1:
`|| travis_terminate 1` looks like the best way to handle this.

So we don't need `set -e` any more.

ekilmer:
@jedisct1 The CI is supposed to fail because of the compilation error, but it doesn't when I remove the `set -e`. Which means the `set -e` is still needed in order to propagate the error in the devenv start script to the devenv run script.

ekilmer:
Master is still failing https://travis-ci.org/fastly/lucet/jobs/585003700#L2256

jedisct1:
Error handling is missing in `devenv_build_container.sh`.

Issue #282, opened 2019-09-05
Allowing compilation of a static library in lucet


shravanrn:
**Context**
This is part of a series of bugs that I spoke to @tyler @pchickey about. We are currently using  Lucet to sandbox libraries in C++ applications. The idea behind this is that using a wasm sandboxed version of the library allows ensuring that a memory safety issue in the library does not automatically result in a memory safety vulnerability in the full application. One of the  consumers of this work is the Firefox web browser. 

**Problem**
Currently lucet applications are compiled as shared object files that must have a main function. However, these restrictions make static linking of libraries very challenging. I thus wanted to start a discussion about how to allow static linking of binaries produces by lucet. 

The most obvious changes that come to mind are
- Removing the requirement of having a main function
- Symbol name collisions - when linking multiple binaries generated via lucet, both modules would have a `lucet_module_data` symbol which would collide.

**Actions**
Any thoughts on the feasibility of this request or whether would be fantastic. Additionally, if we can land on a design, I would be happy to contribute a patch implementing the same.

acfoltzer:
> * Removing the requirement of having a main function

I don't believe this is currently a requirement, except that `lucet-wasi` defaults to running `_start`. You can use `lucet-wasi --entrypoint foo` to pick a different entrypoint.

> * Symbol name collisions - when linking multiple binaries generated via lucet, both modules would have a `lucet_module_data` symbol which would collide.

Yeah, we definitely would need to configure `lucetc` to emit different symbol names for these. There also will probably be some issues with the relocations for initial table values; @awortman-fastly knows much more about that than I do, though.

awortman-fastly:
> Symbol name collisions - when linking multiple binaries generated via lucet, both modules would have a lucet_module_data symbol which would collide.

Ooh, statically linking multiple lucet modules! That's a cool problem :)

> Yeah, we definitely would need to configure lucetc to emit different symbol names for these.

I think something like a user-supplied prefix (or use the name of the input .wasm) to namespace the symbols we add would work well. That should be pretty easy to teach lucetc. I think the harder part would be indicating a non-default symbol for the module's metadata to anything runtime-side - which table to look into to resolve exports, things like that.

shravanrn:
> I don't believe this is currently a requirement, except that `lucet-wasi` defaults to running `_start`. You can use `lucet-wasi --entrypoint foo` to pick a different entrypoint.

[...]

PR #301, opened 2019-09-16
 Partially address #282 - allow statically linked wasm modules

shravanrn:
This small change allows the first step in allowing use of statically linked wasm modules - allow use of symbols from the current binary. This works only if
1) There is only one wasm module statically linked
2) The application is compiled with "rdynamic" linker flag

I built this on another PR (for optionally excluding minisign) which modifies similar code so that there won't be merge conflicts. So please ignore the "combined diff of the 2 PRs/commits" and only see the diff of the one commit for this change.


pchickey:

PR #302, opened 2019-09-17
Fixes #299 - WasiCtx should open dev/null read write

### LINKED ISSUE #299: WasiCtxBuilder opens dev_null as read only and may use it for writing
WasiCtxBuilder opens `/dev/null` read only [here](https://github.com/fastly/lucet/blob/master/lucet-wasi/src/ctx.rs#L243). `/dev/null` is then used as the default output and error stream for this context [here](https://github.com/fastly/lucet/blob/master/lucet-wasi/src/ctx.rs#L30). This seems problematic for callers who decide they don't want to [inherit_stdio](https://github.com/fastly/lucet/blob/master/lucet-wasi/src/ctx.rs#L61)

shravanrn:


jedisct1:
Thank you!

PR #304, opened 2019-09-19
Refactor Context::init and document initial guest stacks more

awortman-fastly:
Does what it says on the tin

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -202,6 +202,49 @@ impl ContextHandle {
     }
 }
 
+struct CallStackBuilder<'a> {
+    offset: usize,
+    stack: &'a mut [u64],
+}
+
+impl<'a> CallStackBuilder<'a> {
+    pub fn new(stack: &'a mut [u64]) -> Self {
+        CallStackBuilder { offset: 0, stack }
+    }
+
+    fn push(&mut self, val: u64) {
+        self.offset += 1;
+        self.stack[self.stack.len() - self.offset] = val;
+    }
+
+    /// Stores `args` onto the stack such that when a return address is written after, the
+    /// complete unit will be 16-byte aligned, as the x86_64 ABI requires.
+    ///
+    /// That is to say, `args` will be padded such that the current top of stack is 8-byte
+    /// aligned.
+    fn store_args(&mut self, args: &[u64]) {
+        let items_end = args.len() + self.offset;
+
+        if items_end % 2 == 1 {
+            // we need to add one entry just before the arguments so that the arguments start on an
+            // aligned address.
+            self.push(0);
+        }
+
+        for arg in args.iter().rev() {
+            self.push(*arg);
+        }
+    }
+
+    fn offset(&self) -> usize {
+        self.offset
+    }
+
+    fn into_inner(self) -> (&'a mut [u64], usize) {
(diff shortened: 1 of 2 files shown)


lucet-runtime/lucet-runtime-internals/src/context/mod.rs
```diff
@@ -281,6 +324,41 @@ impl Context {
...
+    /// 0x0fc8: | lucet_context_backstop  | <-- This forms an ABI-matching call frame for fptr.
+    /// 0x0fc0: | fptr                    | <-- The actual guest code we want to run.
+    /// 0x0fb8: | lucet_context_bootstrap | <-- The guest stack pointer starts here.
+    /// 0x0fb0: |                         |
+    /// 0x0XXX: ~                         ~ // Rest of the stack needs no preparation.
+    /// 0x0000: |                         |
+    ///         +-------------------------+
+    /// ```
+    ///
+    /// This packing of data on the stack is interwoven with noteworthy constraints on what the
```
awortman-fastly: Explicitly writing out the constraints this deals with made an interesting option visible:
`lucet_context_bootstrap` can load initial volatile volatile registers (so, arguments from `fptr`) from the stack, with an initial thunk like
```
pop rdi
pop rsi
pop rcx
pop rdx
pop r8
pop r9
ret
```
This frees up all callee-saves for use however we'd want - and means we can keep the parent and child contexts in, for example, `r12` and `r13`. This also avoids forcing the [gymnastics](https://github.com/fastly/lucet/pull/273/files#diff-adb782ea0cfdba94fce93722c7e66267R98) I needed to do to thread the heap address through to r15 in the guest, since r15 becomes available in the initial `Context` as well.

acfoltzer: How would this interact with yield/resume? Would the new `Context::swap()` still be suitable even if we didn't have a function entrypoint?

lucet-runtime/lucet-runtime-internals/src/context/mod.rs
```diff
@@ -318,52 +396,52 @@ impl Context {
...
 
         // RSP, RBP, and sigset still remain to be initialized.
-        // Stack pointer: this has the return address of the first function to be run on the swap.
-        child.gpr.rsp = &mut stack[sp - stack_start] as *mut u64 as u64;
-        // Frame pointer: this is only used by the backstop code. It uses it to locate the ctx and
-        // parent arguments set above.
-        child.gpr.rbp = &mut stack[sp - 2] as *mut u64 as u64;
+        // Stack pointer: this points to the return address that will be used by `swap`, in place
+        // of the original (eg, in the host) return address. This is the address of the first
+        // function to be run on the swap.
```
acfoltzer: Could this specifically say `lucet_context_bootstrap` to disambiguate from `fptr`?

lucet-runtime/lucet-runtime-internals/src/context/mod.rs
awortman-fastly: Ah this wouldn't change what `Context::swap()` works with. As far as the guest call stack would be concerned, nothing changes. These are all caller-save registers, so they're not saved in `swap()` anyway, and there's no difference there. So the only change is that initial argument registers that make it to `fptr` would come from the stack, instead of through callee-saves we set up the initial Context with

acfoltzer: Got it, that sounds good!

Issue #152, opened 2019-04-24
lucetc no longer looks at wasm name sections


pchickey:
Before the refactor to use `cranelift-wasm` (#44), lucetc would use the name section in a WebAssembly module to give more useful object file symbol names to non-exported functions.

Currently, lucetc uses `cranelift-wasm`'s `ModuleEnvironment` trait to parse sections out of a WebAssembly module, but `ModuleEnvironment` does not expose any name sections it comes across. We could either patch `ModuleEnvironment` to support name sections, or we could stop using `ModuleEnvironment`/`cranelift_wasm::translate_module`, and just create a `ModuleInfo` by using `wasmparser` directly.

data-pup:
I would love to try and take a swing at this!  yurydelendik/wasmparser.rs#70 added a `NameSectionReader` to the wasmparser crate, so do you think that patching `ModuleEnvironment` would be worth trying first?

If you have advice on how to get this done, any help is appreciated! Otherwise, I can follow up with questions as they come up.

pchickey:
Cool, this would be a great one to do next!

You could patch `ModuleEnvironment` first if you want. However, having talked it through with @sunfishcode and others, I don't think there's a really compelling reason for the `ModuleEnvironment` trait to stick around at all - consumers should just inline `cranelift_wasm::translate_module`.

That would be a little bit of a bigger change, but I think it would leave us in a better place. We want to support more extensions to Wasm MVP soon and it will just create work to patch `ModuleEnvironment` to support each as we go, rather than just get them from `wasmparser`.

data-pup:
I'm just about done w/ CraneStation/cranelift#811, which will let us get name information out of the module. Once that's landed, I'll be able to continue work on this on the lucet side of things. :slightly_smiling_face:

PR #308, opened 2019-09-20
upgrade to cranelift 0.44.0, wasmparser 0.39.1

pchickey:
Brings in Cranelift 0.44.0 and wasmparser 0.39.1.

* Wasmparser 0.39.1 returns much better errors on a verifier error
* Cranelift-wasm uses the latest wasmparser, and exposes custom sections to module environment (relevant to #152)
* lots of other cranelift codegen updates. the known regression in 0.43 is fixed in this release

### DIFF
--- a/benchmarks/lucet-benchmarks/src/par.rs
+++ b/benchmarks/lucet-benchmarks/src/par.rs
@@ -42,7 +42,7 @@ fn par_instantiate<R: RegionCreate + 'static>(c: &mut Criterion) {
     let workdir = TempDir::new().expect("create working directory");
 
     let so_file = workdir.path().join("out.so");
-    compile_hello(&so_file, OptLevel::Fast);
+    compile_hello(&so_file, OptLevel::default());
 
     let module = DlModule::load(&so_file).unwrap();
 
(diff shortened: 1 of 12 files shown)


pchickey:
blocked on https://github.com/CraneStation/cranelift/issues/980 - correctness regression for cranelift 0.43.*

benchmarks/lucet-benchmarks/src/par.rs
```diff
@@ -42,7 +42,7 @@ fn par_instantiate<R: RegionCreate + 'static>(c: &mut Criterion) {
     let workdir = TempDir::new().expect("create working directory");
 
     let so_file = workdir.path().join("out.so");
-    compile_hello(&so_file, OptLevel::Fast);
+    compile_hello(&so_file, OptLevel::default());
```
acfoltzer: Could you make these be `SpeedAndSize` explicitly for the benchmark suite, in case the default changes?

pchickey: good idea! i put a `const BENCHMARK_OPT_LEVEL` in each source file that uses it in benchmarks.

PR #311, opened 2019-09-26
Signal handling tweaks

awortman-fastly:
This is a pair of changes to add some wiggle room in our signal handler and avoid overflowing signal stacks.

First, move a stack-heavy part of `handle_signal` to a thunk, so debug builds defer reserving that region's stack use until we reach it, freeing ~5kb of stack for other paths through the signal handler.

Second, add a guard page under the guest's signal stack, to catch sigstack overflow when handling guest faults.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
@@ -180,31 +180,48 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
                 true
             }
             SignalBehavior::Default => {
-                // safety: pointer is checked for null at the top of the function, and the
-                // manpage guarantees that a siginfo_t will be passed as the second argument
-                let siginfo = unsafe { *siginfo_ptr };
-                let rip_addr = rip as usize;
-                // If the trap table lookup returned unknown, it is a fatal error
-                let unknown_fault = trapcode.is_none();
+                /*
+                 * /!\ WARNING: LOAD-BEARING THUNK /!\
+                 *
+                 * This thunk, in debug builds, introduces multiple copies of UContext in the local
+                 * stack frame. This also includes a local `State`, which is quite large as well.
+                 * In total, this thunk accounts for roughly 5kb of stack use, where default signal
+                 * stack sizes are typically 8kb total.
+                 *
+                 * In code paths that do not pass through this (such as immediately reraising as a
+                 * host signal), the code in this thunk would force an exhaustion of more than half
+                 * the stack, significantly increasing the likelihood the Lucet signal handler may
+                 * overflow some other thread with a minimal stack size.
+                 */
+                let mut thunk = || {
+                    // safety: pointer is checked for null at the top of the function, and the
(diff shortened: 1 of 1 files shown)


lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
```diff
@@ -373,6 +373,7 @@ impl Limits {
             self.globals_size,
             host_page_size(),
```
awortman-fastly: I'm not sure if this guard page really guards anything particularly interesting (globals overflow? sigstack underflow should be impossible), but I'm not worried about the 4k here, so.

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -50,6 +50,7 @@ use std::sync::{Arc, Mutex, Weak};
 /// 0xXXXX  --- global guard page ---
 /// 0xS000: +-----------------------| <-- Sigstack (at globals_start + globals_size + PAGE_SIZE)
 /// 0xSXXX: |  ......sigstack....   | // sigstack is SIGSTKSZ bytes
+/// 0xXXXX  -- sigstack guard page --
```
acfoltzer: Doesn't the guard page between the globals and the sigstack handle overflows of the signal stack?

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -180,31 +180,48 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
...
-                        trapcode: trapcode,
-                        rip_addr,
-                        // Details set to `None` here: have to wait until `verify_trap_safety` to
-                        // fill in these details, because access may not be signal safe.
-                        rip_addr_details: None,
-                    },
-                    siginfo,
-                    context: ctx.into(),
+                /*
+                 * /!\ WARNING: LOAD-BEARING THUNK /!\
```
awortman-fastly: I'm at a loss of how to test paths through our signal handler, just putting that out here. A more specialized version of `sigsegv_handler_during_guest`? Setting the host thread's signal stack to `SIGSTKSZ + page`, testing that we don't fault with only `SIGSTKSZ`, and if we do fault, letting `panic!` do its thing with slightly more breathing room?

I think I just need to try writing such a test.

acfoltzer: That's a tricky one; maybe mmap yourself a `SIGSTK + page` as you mentioned, run the test, then assert that the lowermost page is all zeroes? Wouldn't be perfect, but unless we're explicitly covering our tracks after running the handler it should give us some assurance.

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
awortman-fastly: Yep! Reverted the sigstack guard page change.

awortman-fastly:
This had a conflict with #312, whoops. Resolved now, but I need a :heavy_check_mark: again.

PR #317, opened 2019-09-30
Package lucetc using Docker

pchickey:
The control plane group needed an example, so here is one that works straight from the repo.

Note: this is not tested as part of CI, because if we add anything else to Travis CI itll implode into a black hole of tech debt and take us all down with it.

You can test it by hand with something like:

```
docker build . -t lucetc -f Dockerfile.lucetc
docker run -it -v "$(pwd)":/devenv lucetc lucetc "/devenv/lucet-wasi/examples/build/hello.wasm" --bindings /usr/local/share/lucetc/lucet-wasi.bindings.json
```

### DIFF
--- a/Dockerfile.lucetc
+++ b/Dockerfile.lucetc
@@ -1,0 +1,34 @@
+
+FROM ubuntu:16.04 AS builder
+RUN apt-get update \
+	&& apt-get install -y --no-install-recommends \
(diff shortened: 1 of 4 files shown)


pchickey:
We're going to create and maintain this Dockerfile over in one of control plane's repos.

PR #341, opened 2019-10-21
Report Lucetc JSON errors as an array; add state accessors to Instances

acfoltzer:
- We don't have the capability to produce multiple errors yet, so this only returns an array of a single element. Returning multiple errors will require refactoring of our code as well as upstream dependencies.
- Previously a client would have to keep around the `RunResult` to know which state an instance is in.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance.rs
@@ -697,6 +697,22 @@ impl Instance {
         KillSwitch::new(Arc::downgrade(&self.kill_state))
     }
 
+    pub fn is_ready(&self) -> bool {
+        self.state.is_ready()
+    }
+
+    pub fn is_yielded(&self) -> bool {
+        self.state.is_yielded()
+    }
+
+    pub fn is_faulted(&self) -> bool {
+        self.state.is_faulted()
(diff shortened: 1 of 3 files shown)


lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -794,7 +810,7 @@ impl Instance {
 
     /// Run a function in guest context at the given entrypoint.
     fn run_func(&mut self, func: FunctionHandle, args: &[Val]) -> Result<RunResult, Error> {
-        if !(self.state.is_ready() || (self.state.is_fault() && !self.state.is_fatal())) {
+        if !(self.state.is_ready() || (self.state.is_faulted() && !self.state.is_fatal())) {
```
awortman-fastly: this ought to be `self.is_faulted()`, no?

acfoltzer: The new accessors are defined in terms of the methods on `State`, so in release mode it'd probably be equivalent.

awortman-fastly: Oh! I'd skimmed over the rename in `state.rs` and was just missing that.

PR #342, opened 2019-10-22
Make it safe to send yielded instances to other threads

acfoltzer:
Currently we have an implicit assumption that the TLS variable `HOST_CTX` will always point to the same location it pointed to when `Instance::run()` was first called. This assumption used to be fine, because we always used to return from `run()` with a stack that was either empty or unusable (termination or fault). Whenever a user would be in a position to call an `InstanceHandle` from another thread, we'd have to start with a fresh stack.

Not so with yielded instances, which have a stack that we do expect to resume later. If that "later" happens to be on a different thread `HOST_CTX` will point somewhere different from the saved pointer at the base of the guest stack, and things will go wrong when the guest eventually returns through the backstop.

This is an initial pass to add a test to exercise this behavior, and to fix the bug in the most quick-and-dirty way possible. It would be nice to clean this up to something less brittle before merging.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance.rs
@@ -911,6 +911,11 @@ impl Instance {
 
         self.with_signals_on(|i| {
             HOST_CTX.with(|host_ctx| {
+                unsafe {
+                    let new_host_ctx = &mut *host_ctx.get() as *const _ as u64;
+                    let stack = i.alloc.stack_u64_mut();
+                    stack[stack.len() - 3] = new_host_ctx;
+                };
                 // Save the current context into `host_ctx`, and jump to the guest context. The
                 // lucet context is linked to host_ctx, so it will return here after it finishes,
                 // successfully or otherwise.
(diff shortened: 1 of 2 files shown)


benchmarks/lucet-benchmarks/src/context.rs
```diff
@@ -31,14 +26,14 @@ fn context_swap_return(c: &mut Criterion) {
...
-                )
-                .unwrap();
+                let mut guest_data = GuestData::default();
+                let mut child =
+                    ContextHandle::create_and_init(&mut *stack, &mut guest_data, f as usize, &[])
+                        .unwrap();
+                let child_ref: &mut Context = &mut *child;
+                guest_data.child_ctx = child_ref as *mut Context;
+                let parent_ref: &mut Context = &mut *parent;
+                guest_data.parent_ctx = parent_ref as *mut Context;
```
acfoltzer: This benchmark is segfaulting :grimacing:

acfoltzer:
It might be nice to eventually have `Context::set_parent_ctx(&mut self, other: &Context)` to support weird multi-way context swap chains, but I don't think it's useful for now and would rather not have to write tests for it.

lucet-runtime/lucet-runtime-internals/src/context/tests/rust_child.rs
```diff
@@ -363,7 +359,7 @@ macro_rules! child_n_fp_args {
                 write!(out, $prefix).unwrap();
                 $(
                     write!(out, " {:.1}", $arg).unwrap();
-                );*
```
awortman-fastly: (thank you for fixing the macro warnings btw!)

PR #347, opened 2019-10-30
Add `lucetc --target-cpu=` option to control CPU features used in codegen

acfoltzer:
This will make it more feasible to compile on one machine and run on another when the target triple is the same, but CPU-specific features differ.


acfoltzer:
This introduces a couple new command-line arguments and the associated programmatic APIs for selecting optional CPU features at compile time.

Previously, we used `cranelift-native` to automatically detect and use the features available on the host, but this can cause problems when compiling for a different CPU, even if the target triple is constant.

Now, we can enable or disable individual features using `--target-feature=+sse3,+avx,-popcnt`, for example. For convenience, we can also specify `--target-cpu=haswell`, for example, which enables a fixed set of individual features.

The default behavior of auto-detection and use of features remains the same for now.

lucetc/Cargo.toml
```diff
@@ -9,22 +9,19 @@ categories = ["wasm"]
...
 [dependencies]
 bincode = "1.1.4"
-cranelift-codegen = { path = "../cranelift/cranelift-codegen", version = "0.44.0" }
-cranelift-entity = { path = "../cranelift/cranelift-entity", version = "0.44.0" }
-cranelift-native = { path = "../cranelift/cranelift-native", version = "0.44.0" }
-cranelift-frontend = { path = "../cranelift/cranelift-frontend", version = "0.44.0" }
-cranelift-module = { path = "../cranelift/cranelift-module", version = "0.44.0" }
-cranelift-faerie = { path = "../cranelift/cranelift-faerie", version = "0.44.0" }
-cranelift-wasm = { path = "../cranelift/cranelift-wasm", version = "0.44.0" }
+cranelift-codegen = { path = "../cranelift/cranelift-codegen", version = "0.46.1" }
```
acfoltzer: I ended up not actually changing cranelift to make this work, so this upgrade is just a nice little bonus :sparkles:

lucetc/Cargo.toml
```diff
@@ -9,22 +9,19 @@ categories = ["wasm"]
 authors = ["Lucet team <lucet@fastly.com>"]
 edition = "2018"
 
-[lib]
-crate-type=["rlib"]
```
acfoltzer: Doesn't actually change anything; this is the implicit default behavior for a manifest with a `src/lib.rs`

lucetc/src/compiler.rs
```diff
@@ -84,12 +89,15 @@ impl<'a> Compiler<'a> {
                 .context(LucetcErrorKind::Validation)?;
         }
 
-        translate_module(wasm_binary, &mut module_info).map_err(|e| match e {
-            WasmError::User(_) => e.context(LucetcErrorKind::Input),
-            WasmError::InvalidWebAssembly { .. } => e.context(LucetcErrorKind::Validation), // This will trigger once cranelift-wasm upgrades to a validating wasm parser.
-            WasmError::Unsupported { .. } => e.context(LucetcErrorKind::Unsupported),
-            WasmError::ImplLimitExceeded { .. } => e.context(LucetcErrorKind::TranslatingModule),
-        })?;
+        let module_translation_state =
```
acfoltzer: This bit is just for the cranelift update. I guess this new state is for multiple return value support

lucetc/lucetc/options.rs
```diff
@@ -188,6 +238,69 @@ impl Options {
...
+                    .possible_values(&[
+                        "+sse3", "-sse3",
+                        "+ssse3", "-ssse3",
+                        "+sse41", "-sse41",
+                        "+sse42", "-sse42",
+                        "+popcnt", "-popcnt",
+                        "+avx", "-avx",
+                        "+bmi1", "-bmi1",
+                        "+bmi2", "-bmi2",
+                        "+lzcnt", "-lzcnt",
```
acfoltzer: you: "this doesn't look like valid rustfmt"
me: "correct, for some reason rustfmt does nothing to the right-hand side of this `let m = ...` binding. maybe it exceeds some expression size threshold?"

lucetc/src/compiler/cpu_features.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub has_popcnt: bool,
+    pub has_avx: bool,
+    pub has_bmi1: bool,
+    pub has_bmi2: bool,
+    pub has_lzcnt: bool,
+}
+
+impl SpecificFeatures {
+    /// Return a `CpuFeatures` with all optional features disabled.
+    pub fn all_disabled() -> Self {
```
awortman-fastly: How about calling this `baseline`? Functionally this means "generate for anything NetBurst/San Diego or newer", but we still require 64-bit support, for example.

lucetc/src/compiler/cpu_features.rs
```diff
@@ -0,0 +1,159 @@
...
+            has_lzcnt: false,
+        }
+    }
+}
+
+impl From<TargetCpu> for SpecificFeatures {
+    fn from(cpu: TargetCpu) -> Self {
+        use TargetCpu::*;
+        match cpu {
+            Baseline => Self::all_disabled(),
```
awortman-fastly: The above comment was before I'd scrolled to this line, for the record :)

lucetc/src/compiler/cpu_features.rs
```diff
@@ -0,0 +1,159 @@
...
+            Haswell => Self {
+                // Note: this is not part of the Cranelift profile for Haswell, which only uses
+                // CPUID detection to enable AVX. If we want to bypass CPUID when compiling, we need
+                // to set it manually, and Haswell is the first of the CPUs with profiles to have
+                // AVX.
+                has_avx: true,
+                has_bmi1: true,
+                has_bmi2: true,
+                has_lzcnt: true,
+                ..Nehalem.into()
```
awortman-fastly: :clap: this is a very cool trick for reusing settings!

lucetc/src/compiler/cpu_features.rs
```diff
@@ -0,0 +1,159 @@
...
+                has_sse3: true,
+                has_ssse3: true,
+                has_sse41: true,
+                has_sse42: true,
+                has_popcnt: true,
+                // Note: similarly to the Haswell AVX flag, we don't want to rely on CPUID detection
+                // here, but Ryzen does support AVX.
+                has_avx: true,
+                has_bmi1: true,
+                has_bmi2: true,
```
awortman-fastly: It might be worth a note here since we're talking about specific architectures - we technically don't have a profile that properly describes Bulldozer (which has BMI1 and not BMI2). I'm not sure if Cranelift even uses those instructions, and the likelihood of it affecting people seems low, but worth at least being aware of.

lucetc/src/compiler/cpu_features.rs
```diff
@@ -0,0 +1,159 @@
...
+                has_sse42: true,
+                has_popcnt: true,
+                ..Baseline.into()
+            },
+            Haswell => Self {
+                // Note: this is not part of the Cranelift profile for Haswell, which only uses
+                // CPUID detection to enable AVX. If we want to bypass CPUID when compiling, we need
+                // to set it manually, and Haswell is the first of the CPUs with profiles to have
+                // AVX.
+                has_avx: true,
```
fst-crenshaw: I'm so shiny and new that I don't expect to contribute meaningfully to PR review.  That said, I do like to treat PR review as an opportunity for learning.  When I look at the CPU settings in crane lift, I don't see that it `has-avx`. 

https://github.com/CraneStation/cranelift/blob/c053a0bbe1fec8c23ad6f15148e0b26efe194f7c/cranelift-codegen/meta/src/isa/x86/settings.rs#L81-L83

 I wonder how I would learn that haswell has the AVX instruction, as you've indicated here, aside from glancing at a Wikipedia page (like I did over at: https://en.wikipedia.org/wiki/Advanced_Vector_Extensions).

lucetc/src/compiler/cpu_features.rs
acfoltzer: Cranelift at least detects them; I'm not sure if it does anything interesting with them, though. My thinking is that someone wanting to target profile we don't have baked-in can always specify the feature set manually. It's not too hard to add new ones though, so I may transcribe a bit more of the gcc manpage...

lucetc/src/compiler/cpu_features.rs
acfoltzer: It still is detecting and using AVX further up in that file: https://github.com/CraneStation/cranelift/blob/c053a0bbe1fec8c23ad6f15148e0b26efe194f7c/cranelift-codegen/meta/src/isa/x86/settings.rs#L12

It's strange, though, that it doesn't include it in the CPU family profiles.

Also, it took a little bit to find, but an x86_64 gcc's manpage gives a pretty good rundown on what these aliases entail: https://embarc.org/man-pages/gcc/x86-Options.html

fst-crenshaw: Thanks for this guidance!  ❤️   Thanks, too, for validating the strangeness that AVX isn't included in the CPU family profiles.  Please let me know if it's alright to ask these sorts of questions on PRs, or if slack is a better place for these kind of "slightly out of scope" conversations.

lucetc/src/compiler/cpu_features.rs
```diff
@@ -18,89 +21,49 @@ pub enum TargetCpu {
...
-            has_ssse3: false,
-            has_sse41: false,
-            has_sse42: false,
-            has_popcnt: false,
-            has_avx: false,
-            has_bmi1: false,
-            has_bmi2: false,
-            has_lzcnt: false,
+impl TargetCpu {
+    fn features(&self) -> Vec<SpecificFeature> {
```
awortman-fastly: :clap:

PR #350, opened 2019-10-31
Run more test cases per fuzz job

acfoltzer:
Run 100k fuzz tests per run of the fuzz job.

Github apparently gives us six hours per job, so we can probably turn this up once we know how long that number takes on their infra.


acfoltzer:
I assume that if it returns a non-zero exit code, it'll email us like it does for other Actions failures?

pchickey:
oh. i route all that my github email to the trash, so I didn't even know that.

PR #351, opened 2019-10-31
Include CPU features in module data

awortman-fastly:
In theory if a module is built with features the runtime doesn't support, the runtime will fail to load the module with a descriptive error.

"In theory" because I don't have a test for this, although it seems correct - I could be convinced to add a test-only feature in lucetc so we can write a test around missing features being rejected, but because each feature is hand-specified, even that's a little brittle. Comments and opinions encouraged, as it's very much not my favorite code.

Also fixes lucet-spectest bitrot and adds a build of it to `make test`.

### DIFF
--- a/lucet-module/src/module_data.rs
+++ b/lucet-module/src/module_data.rs
@@ -34,6 +34,38 @@ pub struct ModuleData<'a> {
     export_functions: Vec<ExportFunction<'a>>,
     signatures: Vec<Signature>,
     module_signature: Vec<u8>,
+    features: ModuleFeatures,
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct ModuleFeatures {
+    pub sse3: bool,
+    pub ssse3: bool,
+    pub sse41: bool,
+    pub sse42: bool,
+    pub avx: bool,
+    pub bmi1: bool,
+    pub bmi2: bool,
+    pub lzcnt: bool,
+    pub popcnt: bool,
+    _hidden: (),
+}
+
+impl ModuleFeatures {
+    pub fn none() -> Self {
+        Self {
+            sse3: false,
+            ssse3: false,
+            sse41: false,
+            sse42: false,
+            avx: false,
+            bmi1: false,
+            bmi2: false,
+            lzcnt: false,
+            popcnt: false,
+            _hidden: (),
+        }
+    }
 }
 
 impl<'a> ModuleData<'a> {
@@ -44,6 +76,7 @@ impl<'a> ModuleData<'a> {
         import_functions: Vec<ImportFunction<'a>>,
         export_functions: Vec<ExportFunction<'a>>,
         signatures: Vec<Signature>,
+        features: ModuleFeatures,
     ) -> Self {
         let module_signature = vec![0u8; SignatureBones::BYTES];
         Self {
@@ -54,6 +87,7 @@ impl<'a> ModuleData<'a> {
             export_functions,
             signatures,
             module_signature,
+            features,
         }
     }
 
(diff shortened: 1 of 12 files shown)


lucet-module/src/module_data.rs
```diff
@@ -34,6 +34,38 @@ pub struct ModuleData<'a> {
...
+    pub sse3: bool,
+    pub ssse3: bool,
+    pub sse41: bool,
+    pub sse42: bool,
+    pub avx: bool,
+    pub bmi1: bool,
+    pub bmi2: bool,
+    pub lzcnt: bool,
+    pub popcnt: bool,
+    _hidden: (),
```
acfoltzer: What's this for?

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -13,6 +13,65 @@ use std::slice;
...
+    fn missing_feature(feature: &str) -> Error {
+        Error::Unsupported(format!(
+            "Module requires feature host does not support: {}",
+            feature
+        ))
+    }
+
+    let info = cpuid
+        .get_feature_info()
+        .ok_or_else(|| Error::Unsupported("Unable to obtain host CPU feature info!".to_string()))?;
```
acfoltzer: I'm not sure how expensive this is, but we could put it in a `std::sync::Once` if it moves the needle.

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -13,6 +13,65 @@ use std::slice;
 use std::slice::from_raw_parts;
 use std::sync::Arc;
 
+use raw_cpuid::CpuId;
+
+fn check_feature_support(module_features: &ModuleFeatures) -> Result<(), Error> {
+    let cpuid = CpuId::new();
+
+    fn missing_feature(feature: &str) -> Error {
+        Error::Unsupported(format!(
```
acfoltzer: This should probably be a new variant on `ModuleError`, but we're already using `Unsupported` for similar things so I'm not gonna sweat it for now

lucet-spectest/src/script.rs
```diff
@@ -70,6 +70,7 @@ impl ScriptEnv {
         let compiler = Compiler::new(
             module,
             OptLevel::default(),
+            CpuFeatures::baseline(),
```
acfoltzer: Interesting choice to be made here (and also with optimization level): this will end up testing Cranelift with codegen settings we'll almost never actually use in practice

lucetc/src/compiler/cpu_features.rs
```diff
@@ -66,6 +69,86 @@ pub struct CpuFeatures {
...
+        features.bmi1 = info.has_bmi1();
+        features.bmi2 = info.has_bmi2();
+    }
+
+    if let Some(info) = cpuid.get_extended_function_info() {
+        features.lzcnt = info.has_lzcnt();
+    }
+}
+
+impl Into<ModuleFeatures> for CpuFeatures {
```
acfoltzer: I would prefer this as `impl From<&CpuFeatures> for ModuleFeatures`

lucet-spectest/src/script.rs
awortman-fastly: Definitely. Since we were talking about revisiting the spectest runner anyway, I'd _love_ to run spectests under at least both the minimum supported hardware, and most recent supported hardware.

lucet-module/src/module_data.rs
awortman-fastly: It's to prevent exhaustive matching on the struct by naming all fields. Since all the other fields of this struct are pub, you _could_ exhaustively match that way, meaning adding fields is a semver-breaking change. I'm particularly uncomfortable with that because this is in `lucet-module` which my mental model pairs with a pretty vigilant stance around versioning.

PR #354, opened 2019-10-31
Use app_from_crate!() instead of hardcoding clap base parameters

### LINKED ISSUE #353: lucetc --version doesn't work
```bash
$ lucetc --version
lucetc

$ lucetc-wasi --version
lucetc
```

jedisct1:
Fixes #353


jedisct1:
`app_from_crate!` depends on other internal macros. We can explicitly import them as well but there might be no guarantee that it's not going to break after a `clap` update.

acfoltzer:
D'oh, right. I guess they haven't upgraded to using `$crate` to fully qualify their macros yet

PR #355, opened 2019-10-31
Bump to 0.4.1 and refine some dependencies

acfoltzer:
There were some problems when trying to publish the crates due to parity-wasm versions differing in
our crates and Cranelift's. I went through and updated and/or fixed the versions of some crates that
look like they could've caused trouble.


acfoltzer:
Also, this changes the cranelift submodule to the 0.46.1 tag rather than the random `master` I had checked out before deciding not to add a new feature

PR #366, opened 2019-11-20
Look at wasm name section

### LINKED ISSUE #152: lucetc no longer looks at wasm name sections

data-pup:
Fixes #152.

# Parsing the custom name section

With this PR we can use the custom name section to generate more useful naming
information about functions, both local and exported.

## Local Functions

Previously, we were previously making up a name like `guest_func_n` for these
using the function index. This is still used as a fallback, in case a local
function does not have an entry in the name section.

One note, is that we need to assign unique names to functions, so we still
append the function index to the name found in the names section. Without that,
tests will fail, because of duplicate `dummy` functions in binaries. I don't
have quite enough context to know of a good solution to this, but it seems like
it would be nice if we could eventually remove that trailing index number, and
simply provide the name we found in the names section. Feedback on this is
welcome!

## Exported Functions

With this patch, we will use function names from the module names, and only use
the exported name as a fallback. Yay!

<details>
  <summary>What name is found in the names section for exported function?</summary>

  Having spent a little bit of time away from WasmWorld, I wanted to make sure
  what happens when the exported name doesn't match the function name in the
  source. For posterity, in case anybody is curious, here's a small proof of
  what that looked like, using the `names_export.wat` test file found in the
  `lucetc/tests/wasm` directory.

  ```
  (module
      (func $func_name (export "export_name")
      )
  )
  ```

  We can use tools from the `wabt` toolchain to check some of this manually.
  `wat2wasm` will compile WebAssembly text files into binaries. Note that debug
  names are not included by default, so we need to pass the `--debug-names`
  flag.

  `wat2wasm --debug-names names_export.wat -o names_export.wasm`

  We can use `wabt` to be sure that the value stored in the custom name section
  won't be the exported name by compining and disassembling this using
  `wasm-objdump`. The following command will print details from the name section.

  ```
  $: wasm-objdump --details --section=name names_export.wasm

  names_export.wasm:      file format wasm 0x1

  Section Details:

  Custom:
   - name: "name"
   - func[0] <func_name>
  ```

  Voila!

</details>


lucet-idl/lucet-idl-test/resources/rust_host/src/harness.rs
```diff
@@ -1,6 +1,281 @@
-// PLACEHOLDER FILE
-// this file is overwritten, and restored, by lucet-idl-test
-pub struct Ctx;
-pub fn ctx() -> Box<Ctx> {
-    Box::new(Ctx)
+use crate::idl::spec::*;
```
pchickey: changes to this file shouldnt be checked in - unfortunately its a side effect of exiting lucet-idl-test via ctrl-c. unfortunately, we need to just leave a placeholder file checked in.

data-pup: Ah! Thanks for confirming :slightly_smiling_face:

lucetc/tests/wasm.rs
```diff
@@ -508,14 +528,76 @@ mod module_data {
         );
         */
     }
+
+    #[test]
+    fn names_local() {
```
data-pup: I always like to point people towards the tests first, so _here_ there are! :microscope: :woman_scientist: :point_down:

lucetc/tests/wasm.rs
```diff
@@ -2,15 +2,35 @@ use lucet_module::bindings::Bindings;
...
+
+impl Default for LoadOptions {
+    fn default() -> Self {
+        LoadOptions {
+            write_debug_names: false,
+        }
+    }
+}
+
+fn load_wat_module(name: &str, options: LoadOptions) -> Vec<u8> {
```
data-pup: For our name section tests, we need to compile _with_ the names section included. The `wat2wasm` function does [not do this by default](https://github.com/pepyakin/wabt-rs/blob/master/src/lib.rs#L511), so I essentially vendored part of that structure here, since it's not exposed by `wabt-rs`.

This nice thing about this is that we can continue to share code for loading wat modules, and opt into debug information when we need to write tests about that. :slightly_smiling_face:

lucetc/tests/wasm.rs
```diff
@@ -2,15 +2,35 @@ use lucet_module::bindings::Bindings;
...
-    wat2wasm(contents).expect("convert module to wasm binary format")
+    let mut wat2wasm = Wat2Wasm::new();
+    if options.write_debug_names {
+        wat2wasm.write_debug_names(true);
+    }
+    wat2wasm
+        .convert(contents)
+        .expect("convert module to wasm binary format")
+        .as_ref()
+        .to_owned()
```
data-pup: :point_up: here, we're basically writing the equivalent to the `wat2wasm` function we were importing before. You can see the definition of that function [here](https://github.com/pepyakin/wabt-rs/blob/master/src/lib.rs#L1040).

lucetc/tests/wasm.rs
```diff
@@ -32,14 +52,14 @@ pub fn test_bindings() -> Bindings {
...
-    use super::load_wat_module;
+    use super::{load_wat_module, LoadOptions};
     use lucet_module::bindings::Bindings;
     use lucetc::{Compiler, CpuFeatures, HeapSettings, LucetcErrorKind, OptLevel};
     use std::path::PathBuf;
 
     #[test]
     fn exported_import() {
-        let m = load_wat_module("exported_import");
+        let m = load_wat_module("exported_import", LoadOptions::default());
```
data-pup: The rest of the changes in this function are noise from the new `load_wat_module` parameter, luckily via `Default` we don't need to be concerned with setting fields. :slightly_smiling_face:

lucetc/src/module.rs
```diff
@@ -401,4 +404,14 @@ impl<'a> ModuleEnvironment<'a> for ModuleInfo<'a> {
...
+
+    fn declare_func_name(&mut self, func_index: FuncIndex, name: &'a str) -> WasmResult<()> {
+        let unique_func_index = *self
+            .function_mapping
+            .get(func_index)
+            .expect("function indices are valid");
+        self.function_names
+            .insert(unique_func_index, name.to_owned());
+        Ok(())
+    }
```
data-pup: Over in bytecodealliance/cranelift#811, we added a method to the `ModuleEnvironment` trait for handling function names found in the name section. We define that here. The main point of complexity here is that we need to get the unique function id using `self.function_mapping`, or we won't have an easy way to get the name later.

data-pup: Note: Above I've defined `ModuleInfo::function_names` as `HashMap<UniqueFuncIndex, String>`. I _wanted_ a way to use a `PrimaryMap` but I'm not sure that's possible here. Let me know if I've missed something w/ respect to that.

lucetc/src/decls.rs
```diff
@@ -117,18 +117,31 @@ impl<'a> ModuleDecls<'a> {
...
+                    .cloned()
+                    .map(|s| format!("{}_{}", s, ix))
+            }
+
             fn export_name_for<'a>(
                 func_ix: UniqueFuncIndex,
                 decls: &mut ModuleDecls<'a>,
             ) -> Option<String> {
                 let export = decls.info.functions.get(func_ix).unwrap();
-
```
data-pup: This is an etiquette question tangential to the PR at hand, but because the _Opinions About Whitespace Haver_ has logged on, I figured it was worth asking so as not to step on toes: Should I stop myself from making _technically unrelated_ changes like this in future PR's?

lucetc/src/decls.rs
```diff
@@ -117,18 +117,31 @@ impl<'a> ModuleDecls<'a> {
...
+                ix: usize,
+                func_index: UniqueFuncIndex,
+                decls: &mut ModuleDecls<'a>,
+            ) -> Option<String> {
+                decls
+                    .info
+                    .function_names
+                    .get(&func_index)
+                    .cloned()
+                    .map(|s| format!("{}_{}", s, ix))
```
data-pup: :point_up:  this lil' friend here is a helper function for grabbing the name of a function we may have found while parsing. `UniqueFuncIndex` doesn't implement `fmt::Display` so I was using `func_index`, since I don't think we really need to add that derivation while we have a perfectly good `ix`.

lucetc/tests/wasm.rs
```diff
@@ -508,14 +528,76 @@ mod module_data {
...
+        assert_eq!(mdata.export_functions().len(), 0);
+        assert_eq!(mdata.function_info().len(), 3);
+        assert_eq!(
+            mdata.function_info().get(0).unwrap().name,
+            Some("func_name_0")
+        )
+    }
+
+    #[test]
+    fn names_export() {
```
data-pup: oh! also, happy to add a test here to make sure that the export name is used as a fallback if no names section exists, wasn't sure if that was overkill or not.

lucetc/src/decls.rs
```diff
@@ -155,32 +168,33 @@ impl<'a> ModuleDecls<'a> {
...
                     // if a function is an export and import, it will not have a real function body
                     // in this program, and we must not declare it with Linkage::Export (there will
                     // never be a define to satisfy the symbol!)
-
                     decls.declare_function(clif_module, import_sym, Linkage::Import, func_index)?;
                 }
                 (None, Some(export_sym)) => {
                     // This is a function that is only exported, so there will be a body in this
                     // artifact. We can declare the export.
+                    let export_sym = custom_info.unwrap_or(export_sym);
```
iximeow: I'm not sure we want to do this - if a function is declared an export then there's an explicit name someone's expecting to see it by, and at the least we're appending its index (an implementation detail from Lucet) to that name. We ought to be able to just use `export_sym` because duplicate exports would be rejected by wasm validation.

From there, the big issue is "what if a custom name collides with an export name?", which I think is only possible if there's an export `foo_N` for some function, and a custom function named `foo` at index N, where we'll have the export as `guest_func_foo_N` and the custom name will _also_ be `guest_func_foo_N`. I'm not sure if this is a novel error mode, because I _think_ symbols can be just numbers, and if so it would Technically Be Possible that we have this circumstance with an export `1234` with an anonymous function at index `1234`.

It probably makes sense to disambiguate `guest_func` from exported functions generally, so that we might have `guest_func_foo_1235` for an internal function, and `guest_export_foo` for an actual honest-to-goodness exported `foo`.  (To be clear, changing `guest_func` to `guest_export` would probably be best in a different PR if we can put together a test demonstrating the collision.)

lucetc/src/decls.rs
iximeow: I personally loosely prefer whitespace changes (in the form of trailing spaces or spaceful newlines) to be a distinct commit if not a different PR, but I'm curious about other opinions! (I tend to preserve whitespace except in situations of trailing whitespace, so I've never asked either)

lucetc/src/module.rs
iximeow: I think the thing you'd end up needing is in fact a `SecondaryMap`, but with a `SecondaryMap<UniqueFunctionIndex, &'a str>` I'm not sure how the lifetime of its `default` would work (maybe `SecondaryMap::with_default("")` does actually work? that would be super cool). We could probably do the same `HashMap` -> `SecondaryMap` treatment for `function_bodies` as well, but I'm not sure if it's very meaningful.

if memory serves, `PrimaryMap` is intended be kind of an authoritative source for the keyspace, so two `PrimaryMap` keyed on the same thing seem like they'd be in error. I see we have a few of those in `module.rs` and dunno if that was an oversight, intentional (and my understanding wrong), or `SecondaryMap` didn't exist yet?

data-pup: https://github.com/cranestation/cranelift/blob/3217e4e3253bd15a93f1a191345c5e3a0dab31de/upstream#L22-L23

> There should only be a single `PrimaryMap` instance for a given `EntityRef` type, otherwise
> conflicting references will be created. Using unknown keys for indexing will cause a panic.

It seems like you are right, re: `PrimaryMap` v. `SecondaryMap`. And it turns out the default lifetime worked out just fine! :tada:

lucetc/src/module.rs
```diff
@@ -72,6 +72,8 @@ pub struct ModuleInfo<'a> {
     pub function_mapping: PrimaryMap<FuncIndex, UniqueFuncIndex>,
     /// Function signatures: imported and local
     pub functions: PrimaryMap<UniqueFuncIndex, Exportable<'a, SignatureIndex>>,
+    /// Function names.
+    pub function_names: HashMap<UniqueFuncIndex, String>,
```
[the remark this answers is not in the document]
data-pup: Nice, yeah. As mentioned in the comment re: `PrimaryMap`, switched this over to a `SecondaryMap`, and added in the borrowing suggested here. Thanks for the tip :tada:

lucetc/src/decls.rs
data-pup: Ah yeah, good point! Undid this, the name collision stuff seems like it can be investigated outside of this PR :slightly_smiling_face:

lucetc/tests/wasm.rs
```diff
@@ -2,15 +2,35 @@ use lucet_module::bindings::Bindings;
...
-    use wabt::wat2wasm;
+    use wabt::Wat2Wasm;
     let watfile = PathBuf::from(&format!("tests/wasm/{}.wat", name));
     let mut contents = Vec::new();
     let mut file = File::open(&watfile).expect("open module file");
     file.read_to_end(&mut contents).expect("read module file");
-    wat2wasm(contents).expect("convert module to wasm binary format")
+    let mut wat2wasm = Wat2Wasm::new();
+    if options.write_debug_names {
+        wat2wasm.write_debug_names(true);
```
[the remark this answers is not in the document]
data-pup: I don't think we'll need the other options, so I went ahead and always write debug names. Didn't seem to slow things down really, and simplifies the function here a little :+1:

PR #368, opened 2019-11-25
Add `#[lucet_hostcall]` attribute macro and deprecate `lucet_hostcalls!`

acfoltzer:
The attribute macro is much more flexible syntactically, and in particular allows hostcall implementations to be properly `rustfmt`ed rather than being stuck in the invocation of a macro.

This PR also bumps the version to 0.4.2 due to a new API and a deprecation.

Unfortunately the size of the diff is pretty large on this patch, but it's mostly due to the indentation changes that are necessary when converting from a macro invocation block to an attribute.

### DIFF
--- a/benchmarks/lucet-benchmarks/src/modules.rs
+++ b/benchmarks/lucet-benchmarks/src/modules.rs
@@ -1,5 +1,5 @@
 use lucet_module::lucet_signature;
-use lucet_runtime::lucet_hostcalls;
+use lucet_runtime::lucet_hostcall;
 use lucet_runtime::vmctx::{lucet_vmctx, Vmctx};
 use lucet_runtime_internals::module::{
     FunctionPointer, HeapSpec, MockExportBuilder, MockModuleBuilder, Module,
@@ -215,33 +215,32 @@ pub fn many_args_mock() -> Arc<dyn Module> {
 }
 
 pub fn hostcalls_mock() -> Arc<dyn Module> {
-    lucet_hostcalls! {
-        #[inline(never)]
-        #[no_mangle]
-        pub unsafe extern "C" fn hostcall_wrapped(
-            &mut vmctx,
-            x1: u64,
-            x2: u64,
-            x3: u64,
-            x4: u64,
-            x5: u64,
-            x6: u64,
-            x7: u64,
-            x8: u64,
-            x9: u64,
-            x10: u64,
-            x11: u64,
-            x12: u64,
-            x13: u64,
-            x14: u64,
-            x15: u64,
-            x16: u64,
-        ) -> () {
-            vmctx.heap_mut()[0] =
-                (x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9 + x10 + x11 + x12 + x13 + x14 + x15 + x16)
-                    as u8;
-            assert_eq!(vmctx.heap()[0], 136);
-        }
+    #[lucet_hostcall]
+    #[inline(never)]
+    #[no_mangle]
+    pub unsafe extern "C" fn hostcall_wrapped(
+        vmctx: &mut Vmctx,
+        x1: u64,
+        x2: u64,
+        x3: u64,
+        x4: u64,
+        x5: u64,
+        x6: u64,
+        x7: u64,
+        x8: u64,
+        x9: u64,
+        x10: u64,
+        x11: u64,
+        x12: u64,
+        x13: u64,
+        x14: u64,
+        x15: u64,
+        x16: u64,
+    ) -> () {
+        vmctx.heap_mut()[0] =
+            (x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9 + x10 + x11 + x12 + x13 + x14 + x15 + x16)
+                as u8;
+        assert_eq!(vmctx.heap()[0], 136);
     }
 
     #[inline(never)]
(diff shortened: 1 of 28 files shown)


lucet-runtime/lucet-runtime-internals/src/hostcall_macros.rs
```diff
@@ -20,6 +22,7 @@
 /// }
 /// ```
 #[macro_export]
+#[deprecated(since = "0.4.2", note = "Use the #[lucet_hostcall] attribute instead")]
```
data-pup: nice attention to detail here :+1:

benchmarks/lucet-benchmarks/src/modules.rs
```diff
@@ -215,33 +215,32 @@ pub fn many_args_mock() -> Arc<dyn Module> {
 }
 
 pub fn hostcalls_mock() -> Arc<dyn Module> {
-    lucet_hostcalls! {
```
acfoltzer: Note that throughout this PR, whenever you see `lucet_hostcalls!` replaced by `#[lucet_hostcall]`, know that nothing in the body of the function has actually changed, just the syntax required to move from the old macro to the new one.

lucet-runtime/src/lib.rs
```diff
@@ -340,6 +333,10 @@
 
 #![deny(bare_trait_objects)]
 
+// This makes `lucet_runtime` in the expansion of `#[lucet_hostcall]` resolve to something
+// meaningful when used in this crate.
+extern crate self as lucet_runtime;
```
data-pup: nice job figuring this out, that's fairly wonky

lucet-runtime/src/lib.rs
```diff
@@ -369,6 +368,10 @@ pub mod vmctx {
     //! associated with a running instance. This should never occur if run in guest code on the
     //! pointer argument inserted by the compiler.
     pub use lucet_runtime_internals::vmctx::{lucet_vmctx, Vmctx};
+
+    // must be exported for `lucet_hostcall`, but we don't want to advertise it
+    #[doc(hidden)]
+    pub use lucet_runtime_internals::vmctx::VmctxInternal;
```
data-pup: as w/ the deprecated attributes, this is nice attention to detail :+1: great work

lucet-spectest/src/bindings.rs
```diff
@@ -1,27 +1,24 @@
...
 
-lucet_hostcalls! {
-    #[no_mangle]
-    pub unsafe extern "C" fn print_i32(&mut _vmctx, x: i32,) -> () {
-        println!("{}", x);
-    }
+#[lucet_hostcall]
+#[no_mangle]
+pub fn print_i32(_vmctx: &mut Vmctx, x: i32) {
+    println!("{}", x);
```
data-pup: :point_up: Big fan of the ergonomic improvements here.

jedisct1:
Good job! This is solving a long-standing annoying issue.

Should the `lucet_hostcall!` macro be kept around, even a the deprecation notice?

Maybe just bump the version to `0.5.0` and get rid of it, so that both versions don't need to be maintained and tested.

lucet-runtime/lucet-runtime-macros/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// Note that `lucet-runtime` must be a dependency of any crate where this attribute is used, and it
+/// may not be renamed (this restriction may be lifted once [this
+/// issue](https://github.com/rust-lang/rust/issues/54363) is resolved).
+#[proc_macro_attribute]
+pub fn lucet_hostcall(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    // determine whether we need to import from `lucet_runtime_internals`; this is useful if we want
+    // to define a hostcall for a target (or tests, more concretely) that doesn't depend on
+    // `lucet-runtime`
+    let in_internals = std::env::var("CARGO_PKG_NAME").unwrap() == "lucet-runtime-internals";
```
data-pup: very clever

lucet-runtime/lucet-runtime-macros/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    // the args after the first to provide to the hostcall impl
+    let impl_args = hostcall
+        .sig
+        .inputs
+        .iter()
+        .skip(1)
+        .map(|arg| match arg {
+            syn::FnArg::Receiver(_) => {
+                // this case is an error, but humor the token stream so it will produce a nicer
```
data-pup: This part it a little opaque to me, maybe elaborating on what you mean by 'humor' in the comment could help? This is a lightly held opinion, fwiw.

acfoltzer:
> Should the `lucet_hostcall!` macro be kept around, even a the deprecation notice?
> 
> Maybe just bump the version to `0.5.0` and get rid of it, so that both versions don't need to be maintained and tested.

@jedisct1 this is a great point, and I now realize the `lucet_hostcalls!` macro is now completely untested.

I'm going to try redefining it in terms of the attribute so that it's at least generating the same code, but I fear we have too many users at this point to make it disappear overnight.

lucet-runtime/src/lib.rs
acfoltzer: It's not great! Hopefully will be removable once there's a proc macro equivalent of `$crate`

lucet-runtime/lucet-runtime-macros/src/lib.rs
acfoltzer: Just rippin' off workarounds pasted into github issues, as one does :heart:

acfoltzer:
While we're still not actually testing `lucet_hostcalls!` anymore, 80a8c3871464a31a6d5b1a3a4ad85b1cf684fc85 redefines the macro in terms of `#[lucet_hostcall]`. This will at least prevent us from having semantic drift between the two as we phase out the former.

PR #369, opened 2019-11-25
specify signature size for module data

### LINKED ISSUE #349: Signatures don't need to change size
This isn't a bug but I'm liable to forget to do this so I'm writing it down:

The [`module_signature`](https://github.com/fastly/lucet/blob/0ad737c16c9d0cb54d12237503d6605eb674de09/lucet-module/src/module_data.rs#L48) we store in `ModuleData` is typed as a `Vec<u8>`, but the signature itself is a constant size we know ahead of time. The signature could instead be `[u8; BYTES]`, which has the upside of never being an incorrect size.

I was initially more concerned when re-reading this because we do some shenanigans to [patch in a signature](https://github.com/fastly/lucet/blob/0ad737c16c9d0cb54d12237503d6605eb674de09/lucet-module/src/module_data.rs#L116-L128) but we go through serde's deserializer, avoiding issues like building a Vec from a raw pointer into some other data. Making this an array would placate the part of my brain prone to reading code incorrectly. :grin:

data-pup:
Fixes #349.

Using the `serde-big-array` and `derivative` crates, we can change the type of `ModuleData::module_signature` from `Vec<u8>` to `[u8; minisign::SignatureBones::BYTES]`.


jedisct1:
LGTM so far :)

data-pup:
So one question I have is that [ModuleSignature::sign](https://github.com/bytecodealliance/lucet/blob/master/lucet-module/src/signature.rs#L57) means that we have to keep `ModuleData::patch_module_signature` as is, taking a `&[u8]` for the signature param. Is that alright? If so, I think this may be ready.

jedisct1:
`patch_module_signature()` taking a `&[u8]` is totally fine.

data-pup:
Awesome! Thanks for pointing me to `serde-big-array` and `derivable` @jedisct1, those are super neat :smiley_cat:

PR #370, opened 2019-11-26
delete lucet-idl. superceded by witx

pchickey:
Lucet-idl is a spiritual precursor to witx. It has no users in this tree or elsewhere. Now that witx is at the point where it can replace lucet-idl's intended functionality, its time to smash that delete button.

I think that lucet-idl-test in particular had some cool ideas and I'll be going back in the git history to port the proptest strategies to witx at some point.


jedisct1:
Fortunately, it will still live in the Git history forever. Parts of it can still be given a second life later.

jedisct1:
`helpers/bump-global-version.sh` still seems to reference it.

PR #371, opened 2019-11-26
upgrade cranelift, wasi-common, witx deps, and friends

pchickey:
* pull in latest Cranelift release 0.51
* upgrade to wasi-common 0.7, which fixes a lot of the WasiCtxBuilder
  stuff that required many extra unwraps
* pull in latest WASI repo. This moves the preview0 stuff to another
  directory. Biggest change here is that the lucetc deb, which includes
  the wasi witx files, has a slightly different file layout now.
* transitive upgrades to target-lexicon, faerie

### DIFF
--- a/lucet-validate/Cargo.toml
+++ b/lucet-validate/Cargo.toml
@@ -37,8 +37,8 @@ assets = [
     ["target/release/lucet-validate", "/opt/fst-lucet-validate/bin/lucet-validate", "755"],
     ["target/release/liblucet_validate.rlib", "/opt/fst-lucet-validate/lib/", "644"],
     ["LICENSE", "/opt/fst-lucet-validate/share/doc/lucet-validate/", "644"],
-    ["../wasi/phases/unstable/witx/typenames.witx",
-     "/opt/fst-lucet-validate/share/wasi/unstable/typenames.witx", "644"],
-    ["../wasi/phases/unstable/witx/wasi_unstable.witx",
-     "/opt/fst-lucet-validate/share/wasi/unstable/wasi_unstable_preview0.witx", "644"],
+    ["../wasi/phases/old/snapshot_0/witx/typenames.witx",
+     "/opt/fst-lucet-validate/share/wasi/snapshot_0/typenames.witx", "644"],
+    ["../wasi/phases/old/snapshot_0/witx/wasi_unstable.witx",
+     "/opt/fst-lucet-validate/share/wasi/snapshot_0/wasi_unstable.witx", "644"],
 ]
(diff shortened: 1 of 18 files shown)


lucet-validate/Cargo.toml
```diff
@@ -39,6 +39,6 @@ assets = [
     ["LICENSE", "/opt/fst-lucet-validate/share/doc/lucet-validate/", "644"],
     ["../wasi/phases/unstable/witx/typenames.witx",
      "/opt/fst-lucet-validate/share/wasi/unstable/typenames.witx", "644"],
-    ["../wasi/phases/unstable/witx/wasi_unstable_preview0.witx",
+    ["../wasi/phases/unstable/witx/wasi_unstable.witx",
      "/opt/fst-lucet-validate/share/wasi/unstable/wasi_unstable_preview0.witx", "644"],
```
acfoltzer: Do we also want to change the destination filename?

lucet-wasi-fuzz/src/main.rs
```diff
@@ -368,14 +368,11 @@ fn run_with_stdout<P: AsRef<Path>>(
     tmpdir: &TempDir,
     path: P,
 ) -> Result<(__wasi_exitcode_t, Vec<u8>), Error> {
-    let ctx = WasiCtxBuilder::new()
-        .map_err(|e| format_err!("WasiCtxBuilder: {}", e))?
-        .args(["gen"].iter())
-        .map_err(|e| format_err!("WasiCtxBuilder args: {}", e))?;
+    let ctx = WasiCtxBuilder::new().args(&["gen"]);
```
acfoltzer: :pray:

lucet-wasi/src/main.rs
```diff
@@ -235,13 +235,9 @@ fn run(config: Config<'_>) {
...
             .collect::<Vec<&str>>();
         let mut ctx = WasiCtxBuilder::new()
-            .expect("wasi context can be built")
             .args(args.iter())
-            .expect("arguments can be stored")
             .inherit_stdio()
-            .expect("stdio can be inherited")
-            .inherit_env()
-            .expect("environment can be inherited");
+            .inherit_env();
```
acfoltzer: :pray: :pray:

lucet-validate/Cargo.toml
pchickey: Ah, good catch

pchickey: These paths are wrong, we apparently never make the debs in CI. I'll add it to the makefile.

pchickey:
I added testing of the deb creation to `make test`. This will increase Dockerfile build times, and CI run times. Hopefully GH Actions gets docker image caching soon and it doesn't hurt quite as badly, but I'd rather have this stuff tested than find out downstream.

PR #374, opened 2019-12-05
Fixes #357:  Migrate from using the failure to thiserror + anyhow.

### LINKED ISSUE #357: Switch from failure to thiserror
For error handling, we currently use the `failure` crate everywhere.

`failure` was great when it was released and the standard `Error` trait didn't support backtraces yet. But things have changed since.

Backtraces are now supported in `std::error::Error`, so we should use it instead of a custom trait.

Cranelift [switched](https://github.com/CraneStation/cranelift/pull/1188) to `thiserror`, that is lightweight and does everything we need, so we may want to do the same.

fst-crenshaw:
This WIP PR fixes #357 .


lucet-runtime/lucet-runtime-internals/src/context/mod.rs
```diff
@@ -5,12 +5,12 @@ mod tests;
...
 use crate::instance::Instance;
 use crate::val::{val_to_reg, val_to_stack, RegVal, UntypedRetVal, Val};
-use failure::Fail;
 use nix;
 use nix::sys::signal;
 use std::arch::x86_64::{__m128, _mm_setzero_ps};
 use std::ptr::NonNull;
 use std::{mem, ptr};
+use thiserror::Error;
 use xfailure::xbail;
```
acfoltzer: We can now replace uses of `xfailure::xbail` with `anyhow::bail`, since it does the implicit type conversion of errors that I believe was @jedisct1's original motivation for `xfailure`.

lucet-module/src/bindings.rs
```diff
@@ -1,4 +1,4 @@
-use failure::{format_err, Error};
+use anyhow::Error;
```
acfoltzer: You've done a fine job of converting this module over, but I think it's probably time to convert these errors to new variants on the enum in lucet-module/src/error.rs. They're all caused by bad input, as opposed to "impossible" cases that would be bugs if we hit in normal use, which means it'd be nice to be able to get more than a string from them.

lucet-runtime/lucet-runtime-internals/src/error.rs
```diff
@@ -1,81 +1,76 @@
...
     StartYielded,
 
     /// A catch-all for internal errors that are likely unrecoverable by the runtime user.
     ///
     /// As the API matures, these will likely become rarer, replaced by new variants of this enum,
     /// or by panics for truly unrecoverable situations.
-    #[fail(display = "Internal error: {}", _0)]
-    InternalError(#[cause] failure::Error),
+    #[error("Internal error: {}", _0)]
+    InternalError(#[source] AnyError),
```
acfoltzer: Since it's not being used much in this module, I would recommend using the path to disambiguate rather than renaming it upon import; gives the reader fewer places to look to figure out what the type is.
```suggestion
    InternalError(#[source] anyhow::Error),
```

lucet-runtime/lucet-runtime-internals/src/error.rs
```diff
@@ -1,81 +1,76 @@
...
-    #[fail(display = "Unsupported feature: {}", _0)]
+    #[error("Unsupported feature: {}", _0)]
     Unsupported(String),
 }
 
-impl From<failure::Error> for Error {
-    fn from(e: failure::Error) -> Error {
-        Error::InternalError(e)
-    }
-}
```
acfoltzer: We might want to add a `#[from]` to the `InternalError` enum variant to replace this, but if leaving it out didn't cause any errors, we're probably not using `?` in ways that would need it.

lucet-validate/src/main.rs
```diff
@@ -87,14 +87,14 @@ fn run(module_path: &Path, witx_path: &Path, wasi_exe: bool) -> Result<(), Error
...
+#[derive(Debug, Error)]
 enum Error {
-    #[fail(display = "{}", _0)]
-    Witx(#[cause] witx::WitxError),
-    #[fail(display = "With file {:?}: {}", _0, _1)]
-    Io(PathBuf, #[cause] io::Error),
-    #[fail(display = "{}", _0)]
-    Validate(#[cause] lucet_validate::Error),
+    #[error("{}", _0)]
+    Witx(#[source] witx::WitxError),
```
acfoltzer: Some opportunities for `#[from]` on the witx and validate variants

lucet-module/src/error.rs
```diff
@@ -1,16 +1,16 @@
...
-    #[fail(display = "Deserialization error: {}", _0)]
-    DeserializationError(#[cause] bincode::Error),
-    #[fail(display = "Serialization error: {}", _0)]
-    SerializationError(#[cause] bincode::Error),
-    #[fail(display = "Module signature error: {}", _0)]
-    ModuleSignatureError(#[cause] minisign::PError),
-    #[fail(display = "I/O error: {}", _0)]
-    IOError(#[cause] std::io::Error),
+    #[error("Deserialization error: {}", _0)]
+    DeserializationError(#[source] bincode::Error),
```
dtolnay: Take a look at https://users.rust-lang.org/t/do-people-not-care-about-printable-error-chains-a-k-a-how-to-nicely-implement-display-for-an-error/35362 -- normally the inner error should not be included in both the Display impl of the outer error *and* the source, only one or the other, or else it will appear twice when printing out the full chain of causes.

The cause chain here currently would be:

```console
Error: Whatever top level thing failed

Caused by:
    - Deserialization error: whatever Bincode said failed
    - whatever Bincode said failed
```

lucet-runtime/lucet-runtime-internals/src/error.rs
```diff
@@ -1,81 +1,76 @@
...
-use failure::Fail;
+use anyhow::Error as AnyError;
+use thiserror::Error;
 
 /// Lucet runtime errors.
-#[derive(Debug, Fail)]
+#[derive(Debug, Error)]
 pub enum Error {
-    #[fail(display = "Invalid argument: {}", _0)]
+    #[error("Invalid argument: {}", _0)]
```
dtolnay: This is fine too, but consider using:

```rust
#[error("Invalid argument: {0}")]
```

which I think people generally find more readable.

lucet-validate/src/lib.rs
```diff
@@ -1,41 +1,46 @@
...
         module: String,
         field: String,
         expected: FuncSignature,
         got: FuncSignature,
     },
-    #[fail(
-        display = "Export type error: for {}, expected {:?}, got {:?}",
-        field, expected, got
+    #[error(
+        "Export type error: for {}, expected {:?}, got {:?}",
```
dtolnay: I would recommend:

```rust
#[error("Export type error: for {field}, expected {expected:?}, got {got:?}")]
```

rather than keeping the argument order straight.

fst-crenshaw:
My immense thanks to @acfoltzer and @dtolnay for their quick feedback on this WIP PR.  Lots of good thoughts on how to make this migration cleaner and more readable.  ❤️

lucet-runtime/lucet-runtime-internals/src/context/mod.rs
fst-crenshaw: Thanks @acfoltzer!  You and I talked through this in person and identified that the one moment that `bail!` is used in this file may not be an appropriate moment to bail.  If the function was supplied a misaligned stack, let's just return an error indicating as such.

lucet-module/src/bindings.rs
```diff
@@ -27,7 +27,7 @@ impl Bindings {
     pub fn from_json(v: &Value) -> Result<Bindings, Error> {
         match v.as_object() {
             Some(modules) => Self::parse_modules_json_obj(modules),
-            None => Err(format_err!("top level json expected to be object"))?,
+            None => Err(anyhow::format_err!("top level json expected to be object"))?,
```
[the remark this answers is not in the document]
fst-crenshaw: Thanks for this comment!  It helps me to see when using a fully-qualified path in Rust looks a bit strange and when it doesn't.

lucet-module/src/error.rs
fst-crenshaw: Thanks for citing this immensely helpful article!  I can take steps to avoid the annoying double-display.

lucet-validate/src/main.rs
fst-crenshaw: Hm.  Thanks so much for this comment, but I'm having a bit of trouble reading it.  Is this a comment on changes that could be made over in `witx` or changes that could be made in the scope of this PR?  I look forward to learning more.

lucet-module/src/bindings.rs
fst-crenshaw: Got it!  Will do.

lucet-validate/src/main.rs
acfoltzer: Just in this file. You should be able to replace the explicit `From` impls below this by adding the `#[from]` annotation next to the field that they'd otherwise be turned into. I don't believe `#[from]` was a feature available to us in `failure`; yay for less boilerplate!

lucet-module/src/bindings.rs
```diff
@@ -76,13 +75,15 @@ impl Bindings {
         match self.bindings.get(module) {
             Some(m) => match m.get(symbol) {
                 Some(s) => Ok(s),
-                None => Err(format_err!("Unknown symbol `{}::{}`", module, symbol)),
+                None => Err(Error::UnknownSymbol {
+                    module: module.to_string(),
```
fst-crenshaw: @acfoltzer -- I took an initial pass at converting all of the errors in bindings.rs to fields of the enum in errors.rs.  I'm unhappy with my use of to_string() and I wonder if you might have thoughts on a better way.

data-pup: Adam might have other opinions but `.to_string` looks okay to me :slightly_smiling_face:

acfoltzer: `format!("{}", x) ≃ x.to_string()`, so this looks fine to me!

dtolnay: IMO `module.to_owned()` is a better fit here, as explained in:
https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/6

fst-crenshaw: Thanks!

lucet-spectest/src/error.rs
```diff
@@ -1,6 +1,7 @@
-use failure::{Backtrace, Context, Fail};
-use std::fmt::{self, Display};
+//use std::fmt::{self, Display};
```
fst-crenshaw: There's still some cleanup to do here as I track down the handful of spec tests that aren't passing.

fst-crenshaw:
But otherwise, this is a decent complete first-draft that would benefit from your feedback.

lucet-runtime/lucet-runtime-internals/src/error.rs
```diff
@@ -1,81 +1,81 @@
...
 
-    #[fail(display = "Start function yielded")]
+    #[error("Start function yielded")]
     StartYielded,
 
+    // A test in instruction_counting.rs requires the conversion of a
+    // Lucetc error to the error type in this crate.
+    #[error("Lucetc error")]
+    LucetcError(#[from] LucetcError),
+
```
fst-crenshaw: I'm truly not keen on the hackery that I've done here and would love thoughts on another approach.

lucet-spectest/src/lib.rs
```diff
@@ -42,19 +41,15 @@ pub fn run_spec_test(spec_path: &PathBuf) -> Result<SpecScriptResult, Error> {
...
     if e.unsupported() {
-        Error::from(e)
-            .context(SpecTestErrorKind::UnsupportedLucetc)
-            .into()
+        Error::UnsupportedLucetc
     } else {
-        Error::from(e)
-            .context(SpecTestErrorKind::UnexpectedFailure)
-            .into()
+        Error::UnexpectedFailure(String::new())
```
data-pup: :thinking: :thought_balloon: idle thought: should this `String::new` be here? Wondering if either (a) `Error::UnexpectedFailure` doesn't need a payload, or if we could do something in the spirit of `e.to_string()`.

fst-crenshaw: Thanks for the comment!  I've made this change.

lucet-spectest/src/lib.rs
```diff
@@ -226,49 +231,54 @@ fn step(script: &mut ScriptEnv, cmd: &CommandKind) -> Result<(), SpecTestError>
...
-                    Err(format_err!("expected {}, got {}", expected, got.as_f64()))
-                        .context(SpecTestErrorKind::IncorrectResult)?
+                    let message = format!("expected {}, got {}", expected, got.as_f64());
+                    Err(Error::IncorrectResult(message))?
                 }
             }
             Value::V128(_) => {
-                Err(format_err!("got unsupported SIMD V128 value"))
-                    .context(SpecTestErrorKind::UnsupportedCommand)?;
+                let message = format!("got unsupported SIMD V128 value");
```
data-pup: I love how you've outlined the `message` values in this file :heart_decoration: :woman_technologist:

Minor nit: We create a `String` using `format!`, even though the literal doesn't need to interpolate a value with `{}`. We can use `String::from` instead, [like this](https://doc.rust-lang.org/std/string/struct.String.html#examples-97)

```suggestion
                let message = String::from("got unsupported SIMD V128 value");
```

lucet-spectest/src/script.rs
```diff
@@ -169,7 +168,7 @@ impl ScriptEnv {
         let (_, ref mut inst) = self.instance_named_mut(name)?;
         inst.run(field, &args)
             .and_then(|rr| rr.returned())
-            .map_err(|e| ScriptError::RuntimeError(e))
+            .map_err(ScriptError::RuntimeError)
```
data-pup: :sparkle: Stylish! :sparkle:

lucet-spectest/src/lib.rs
fst-crenshaw: Thanks.  I agonized a bit over this.  Was it better to consistently use `format!` for error messages, or use `String::new()` when interpolation isn't needed?  It seems I chose incorrectly.  I will replace all the places where I made this choice.

data-pup: For what it's worth, this is _totally_ a small style nit. I think incorrect is a strong word :slightly_smiling_face:

fst-crenshaw:
There were a few things that were making me cringe on this PR.  I've refactored them all out and I'm removing the `WIP` from the title.  It's ready for your review.

lucetc/src/decls.rs
```diff
@@ -418,10 +407,11 @@ impl<'a> ModuleDecls<'a> {
     }
 
     pub fn get_func(&self, func_index: UniqueFuncIndex) -> Result<FunctionDecl<'_>, Error> {
```
pchickey: The only Error this method throws is FunctionIndexError, which should only happen if the caller provides a bogus func_index. I'd rather it just return an Option<FunctionDecl<'_>> and make it the caller's problem. Then you can delete the FunctionIndexError variant

lucetc/src/decls.rs
```diff
@@ -528,7 +521,10 @@ impl<'a> ModuleDecls<'a> {
             let name = self
                 .function_names
                 .get(fn_index)
-                .ok_or_else(|| format_err!("func index out of bounds: {:?}", fn_index))
+                .ok_or_else(|| {
+                    let message = format!("{:?}", fn_index);
+                    Error::FunctionIndexError(message)
+                })
```
pchickey: This is trying to handle an error that should be impossible - fn_index should always be valid, so I'd just change it to `let name = self.function_names.get(fn_index).expect("fn_index is key into function_names");`. Thats the only other use of the FunctionIndexError variant, so you can delete it after this is gone

lucetc/src/error.rs
```diff
@@ -1,70 +1,123 @@
...
+    #[error("Memory specs: {0}")]
+    MemorySpecs(String),
+    #[error("Metadata serializer; start index point to a non-function")]
+    MetadataSerializer(#[source] ClifModuleError),
+    #[error("Output: {0}")]
+    Output(String),
+    #[error("Output function: error writing function {1}")]
+    OutputFunction(#[source] std::fmt::Error, String),
+    #[error("Path error: {0}")]
+    PathError(String),
```
pchickey: I think these are used for paths provided by the user, so you can fold this into Error::Input

lucetc/src/error.rs
```diff
@@ -1,70 +1,123 @@
...
+    #[error("Read error: {0}")]
+    ReadError(String),
```
pchickey: I'd name this Wat2Wasm error. You could even fold this into Error::Input, just using `format!` to say that its a wat2wasm error

lucetc/src/error.rs
fst-crenshaw: Excellent suggestion.  Thanks!

lucetc/src/error.rs
fst-crenshaw: Thanks for this suggestion!

lucetc/src/compiler.rs
```diff
@@ -65,54 +64,43 @@ impl<'a> Compiler<'a> {
...
-                WasmError::Unsupported { .. } => e.context(LucetcErrorKind::Unsupported),
-                WasmError::ImplLimitExceeded { .. } => {
-                    e.context(LucetcErrorKind::TranslatingModule)
+                WasmError::User(u) => Error::Input(u.to_string()),
+                WasmError::InvalidWebAssembly { message, offset } => {
+                    let message = format!("invalid input at offset {}: {}", offset, message);
+                    Error::Validation(message)
                 }
+                WasmError::Unsupported(s) => Error::Unsupported(s.to_owned()),
+                WasmError::ImplLimitExceeded { .. } => Error::TranslatingModule,
```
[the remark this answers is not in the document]
fst-crenshaw: In the spirit of, _"Hey Tanya, can we please have fewer variants?"_ --  Which I 100% support -- I could replace `Error::TranslatingModule` with the already existing `Error::TranslatingClifWasm` which sources the `WasmError`.  

See: https://github.com/bytecodealliance/lucet/pull/374/files#diff-22e27732dfa181d4f2f2750e1d87e18bR109-R110

Thoughts?

lucetc/src/error.rs
```diff
@@ -1,70 +1,123 @@
...
+    #[error("I/O error")]
+    IOError(#[from] std::io::Error),
+    #[error("Wasm validating parser error")]
+    WasmValidation(#[from] wasmparser::BinaryReaderError),
+    #[error("Wat input")]
+    WatInput(#[from] wabt::Error),
+    /*
+     * Cannot apply #[from] or #[source] to these error types due to missing traits. */
+    #[error("Artifact error: {0:?}")]
+    ArtifactError(failure::Error),
```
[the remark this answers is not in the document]
fst-crenshaw: I agree with you about taking some time to get faerie updated to the standard Error trait and return to this repo for a bit of subsequent cleanup.  I love your suggestion about the one ArtifactError.  Thanks!

lucetc/src/decls.rs
fst-crenshaw: I took your suggestion to have `get_func` return an Option.  Unfortunately, the way I interpreted your suggestion only moves the use of `FunctionIndexError`.  Let me know how I might improve this so I can follow your suggestion 100%.

lucetc/src/decls.rs
fst-crenshaw: Thanks!

fst-crenshaw:
My genuine thanks to @pchickey for the feedback to reduce the total number of variants in `lucetc/src/error.rs.` I implemented all suggestions.  The file went from 123 lines to 85.  That's a savings!  💵 💰 💵 

I look forward to further conversation on how this PR might improve. ❤️

fst-crenshaw:
I've addressed all of @pchickey 's feedback to the best of my ability and merged this PR with the recent updates to lucet.  I would love to hear any further feedback on this PR that folks might have.

PR #375, opened 2019-12-12
add a virtual directory for lucet-wasi tests to use

iximeow:
this is a matched draft with https://github.com/bytecodealliance/wasmtime/pull/701 that I expect to clip down to just
```
ctx = ctx.preopened_virt(Box::new(fs), "/sandbox");
```
or so whenever this is made a real PR. So far as lucet-wasi is concerned, I just want to exercise file APIs and ensure that things are still usable as I poke at wasi-common.


iximeow:
Closing this to be revisited at a later point, now that virtfs has landed in wasi-common

PR #378, opened 2019-12-13
replace empty constructor w/ default impl

data-pup:
This is another "tiny lint fix I found while looking around at things" PR.

Rather than using a `CtxMap::new()` constructor, we can implement the `Default` trait, and use that instead.


acfoltzer:
How about both? We could define one in terms of the other, too. But the standard library has so many instances of `Thing::new()` that I usually reach for that rather than `Thing::default()`

data-pup:
> How about both? We could define one in terms of the other, too.

That's a good point! Will do.

iximeow:
🧹 the merge-approved-but-forgotten-PRs fairy strikes 🧹

PR #391, opened 2020-01-15
ModuleDecls Cleanup

data-pup:
For reasons unrelated to this PR, I ended up looking through `lucetc/src/decls.rs` this morning. I noticed a few warts that would be easy fixes, so this is a small gardening PR.

* Addresses two lint warnings about redundant field names, using the shorthand notation instead.
* Moves `*_name_for` helper functions nested inside of `declare_funcs` out of a loop body. This is just a small readability fix for our loop.

### DIFF
--- a/lucetc/src/decls.rs
+++ b/lucetc/src/decls.rs
@@ -114,58 +114,58 @@ impl<'a> ModuleDecls<'a> {
         clif_module: &mut ClifModule<B>,
         bindings: &'a Bindings,
     ) -> Result<(), LucetcError> {
+        // Get the name for this function from the module names section, if it exists.
+        // Because names have to be unique, we append the index value (ix) to the name.
+        fn custom_name_for<'a>(
+            ix: usize,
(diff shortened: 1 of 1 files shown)


lucetc/src/decls.rs
```diff
@@ -114,58 +114,58 @@ impl<'a> ModuleDecls<'a> {
         clif_module: &mut ClifModule<B>,
         bindings: &'a Bindings,
     ) -> Result<(), LucetcError> {
-        for ix in 0..decls.info.functions.len() {
-            let func_index = UniqueFuncIndex::new(ix);
+        // Get the name for this function from the module names section, if it exists.
```
data-pup: This is all pure code motion, no edits.

iximeow: an extremely cool trick to help review that it is in fact only indentation change: [?w=1](https://github.com/bytecodealliance/lucet/pull/391/files?w=1)

PR #398, opened 2020-01-21
add triple target option and extend cpu features

maxmcd:


iximeow:
Hi! Do you have a particular motivation for this change? There's a similar-but-not-quite-the-same change in progress over at #390, which threads a target triple through to select linker flags but also target architecture as you're doing here.

If you're just trying to pick a target architecture for codegen, I suspect #390 might suit your needs?

You might want to be aware that targets other than x86_64/i686 are fairly untested, beware :)

maxmcd:
Whoops! I was aware of #390 but had done this work and was just intending to push it up to my fork. Closing.

PR #400, opened 2020-01-22
Use exact dependencies for crates within the repo

acfoltzer:
Most of our thinking about semver for this project is concerned with the more "public-facing" crates like `lucetc`, `lucet-runtime`, and `lucet-wasi`, as opposed to "implementation detail" crates like `lucet-module`, `lucet-runtime-internals`, etc. As the project matures, we are paying more attention to semver for all of the crates, but this change makes the released sets of packages more consistent.


acfoltzer:
I'm inclined to say no; the only reason I can think of doing so is if we decouple the versions between the crates, which we've decided not to do.

PR #404, opened 2020-01-29
Start an mdbook for developer documentation and add release instructions

acfoltzer:
I want to get a sense of how folks feel about moving away from the Github wiki to this form of documentation. If y'all like it, I will use this PR to also port over the existing wiki content, and set up github actions to publish at https://bytecodealliance.github.io/lucet. If not, the release guide I wrote up should serve as a fine wiki page.


acfoltzer:
To see what the output looks like for yourself:

```shell
$ cargo install mdbook
$ mdbook serve docs/
```

Then visit http://localhost:3000

docs/src/versioning_releasing.md
```diff
@@ -0,0 +1,103 @@
...
+   release. Between releases, the versions will end in `-dev`; if this is still the case, you'll
+   need to replace this version with the appropriate version according to the guidelines above,
+   likely through a PR.
+
+1. Edit `lucet-validate/Cargo.toml` and make the following change:
+
+   ```diff
+    [dev-dependencies]
+   -lucet-wasi-sdk = { path = "../lucet-wasi-sdk", version = "=0.5.2-dev" }
+   +#lucet-wasi-sdk = { path = "../lucet-wasi-sdk", version = "=0.5.2-dev" }
```
fst-crenshaw: Should these be different?

fst-crenshaw: Ohhhh.  One is commented out.

Timeout for Tanya.  I'll go sit in a corner for two minutes.

acfoltzer: Ah, the difference is easy to miss; it's the leading `#`. I've called it out in the prose

docs/src/versioning_releasing.md
```diff
@@ -0,0 +1,162 @@
...
+   1. `lucet-runtime-internals`
+   1. `lucet-runtime-tests`
+   1. `lucet-runtime`
+   1. `lucet-wasi`
+
+   It is unlikely but not impossible that a publish will fail in the middle of this process, leaving
+   some of the crates published but not others. What to do next will depend on the situation; please
+   consult with the Lucet team.
+
+Congratulations, the new crates are now on crates.io! 🎉
```
fst-crenshaw: I'm not familiar with publishing crates on crates.io, so please forgive this naive question.  Would it make sense to add some kind of sanity check step to make sure the lucet crates that were just published are actually available at crates.io?  I don't want to break other people's toys.

fst-crenshaw:
Let me second @iximeow 's declaration of how [RAD](https://www.youtube.com/watch?v=7MCgFYzbom0) this PR is. I love when docs can be vetted in the PR review process.  I will happily lend a hand with porting docs over to this new approach.

docs/src/versioning_releasing.md
acfoltzer: Added a step for this, thanks!

docs/src/Compiling-on-Linux.md
```diff
@@ -0,0 +1,115 @@
...
+## Installation on any recent Linux system, using the base compiler
+
+Support for WebAssembly was introduced in LLVM 8, released in March 2019.
+
+As a result, Lucet can be compiled with an existing LLVM installation, provided that it is up to
+date. Most distributions now include LLVM 8 or LLVM 9, so that an additional installation is not
+required to compile to WebAssembly .
+
+On distributions such as Ubuntu (> 18.04) and Debian, the following command installs the
+prerequisites:
```
katie-martin-fastly: Maybe replace `> 18.04` with `(versions from 19.04 and onward)`, or something along those lines?

Which Ubuntu versions should use the base compiler, and which should follow the version above, is slightly unclear. This might have been a preexisting issue with the docs, sorry if that's the case :grimacing:

katie-martin-fastly:
I second all of @fst-crenshaw 's points above. This is wonderful work.

docs/src/Compiling-on-Linux.md
acfoltzer: Done!

PR #408, opened 2020-01-31
Make lucetc and lucet-validate errors more verbose

acfoltzer:
The full fix for these issues is tracked in #409, but this can merge first because it is already useful.

### DIFF
--- a/lucetc/src/error.rs
+++ b/lucetc/src/error.rs
@@ -9,31 +9,31 @@ use thiserror::Error;
 pub enum Error {
     //
     // General #[from] implementations.
-    #[error("Builtins error")]
+    #[error("Builtins: {0}")]
     Builtins(#[from] parity_wasm::elements::Error),
-    #[error("Clif module error")]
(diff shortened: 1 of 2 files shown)


lucetc/src/error.rs
```diff
@@ -9,31 +9,31 @@ use thiserror::Error;
...
+    #[error("Builtins: {0}")]
     Builtins(#[from] parity_wasm::elements::Error),
-    #[error("Clif module error")]
+    #[error("Clif module: {0}")]
     ClifModuleError(#[from] ClifModuleError),
-    #[error("Translating error")]
+    #[error("Translating: {0}")]
     ClifWasmError(#[from] ClifWasmError),
-    #[error("Lucet Module error")]
+    #[error("Lucet Module: {0}")]
```
iximeow: this is definitely a "when we actually spec this out with tests" thought, but "Error: Lucet Module error: Unknown module for ..." ends up repeating that there's an error a bit more than we probably should :sweat_smile:

Issue #414, opened 2020-02-05
Argument struct for `lucetc::Compiler::new()`


acfoltzer:
Inspired by https://github.com/bytecodealliance/lucet/pull/405

The number of arguments here is getting out of hand, and it's easy to forget all the places that must be updated when adding a new one. For example, the new functionality in #405 is only available via the API, not the command line. Cleaning this up, and perhaps even sharing a data structure with the command-line parser would improve this a lot.

pchickey:
Closed #418

PR #418, opened 2020-02-09
414 Argument struct for `lucetc::Compiler::new()`

michael-groble:
Approach for  #414

First time working in this code base and not a huge amount of rust experience so comments much appreciated!

This is an approach that attempts to make construction simpler without changing the signature for `Compiler::new`.  `CompilerBuilder` could easily be something more like `CompilerOptions` that gets passed in as an argument to either a new `Compiler` method or `Compiler::new` if that is really the desire.

There are many more places using `Compiler::new` that I can update once I get some agreement on approach.  Also can add some testing.  Just throwing something out there fore feedback first.


lucetc/src/compiler.rs
```diff
@@ -47,6 +47,91 @@ impl OptLevel {
     }
 }
 
+pub struct CompilerBuilder {
+    pub target: Triple,
```
michael-groble: I built this struct before seeing the `with_` conventions in `LucetOpts`.  There is one place (`Lucetc.shared_object_file`) that wants access to this.  Making it `pub` was just a quick way to get something out for review.  Can adopt a different naming convention here to differentiate "accessors" vs. "setters".

lucet-wasi-sdk/tests/lucetc.rs
```diff
@@ -43,20 +42,12 @@ mod lucetc_tests {
...
-            false,
-            &Some(v),
-            false,
-        )
-        .expect("compile empty");
+        let mut builder = Compiler::builder();
+        let c = builder
+            .validator(&Some(v))
+            .create(&m, &b)
+            .expect("compile empty");
```
katie-martin-fastly: This looks much nicer now :tada:

lucetc/src/compiler.rs
```diff
@@ -47,6 +47,91 @@ impl OptLevel {
...
+        self
+    }
+
+    pub fn canonicalize_nans(&mut self, canonicalize_nans: bool) -> &mut Self {
+        self.canonicalize_nans = canonicalize_nans;
+        self
+    }
+
+    pub fn validator(&mut self, validator: &Option<Validator>) -> &mut Self {
+        self.validator = validator.clone();
```
katie-martin-fastly: Do you think that this could accept `Option<Validator>` instead? This will just move the `.clone()`'s to the places that this method is called.

lucetc/src/lib.rs
```diff
@@ -263,18 +284,12 @@ impl Lucetc {
         Self {
             input: LucetcInput::Path(input.to_owned()),
             bindings: vec![],
-            target: Triple::host(),
-            opt_level: OptLevel::default(),
-            cpu_features: CpuFeatures::default(),
-            heap: HeapSettings::default(),
+            builder: Compiler::builder(),
```
katie-martin-fastly: This too, looks much nicer now. :smile_cat:

lucetc/src/compiler.rs
katie-martin-fastly: * How about using `pub(crate)` instead of `pub`? I don't think exposing this to `Lucetc.shared_object_file`, but that would keep `target` from being exposed outside of the crate as well.
* You already noticed the `with_` conventions, following these here would be nice :slightly_smiling_face: I'll tag one of the methods below with an example of the signature change I would hope to see.

lucetc/src/compiler.rs
```diff
@@ -47,6 +47,91 @@ impl OptLevel {
...
+            opt_level: OptLevel::default(),
+            cpu_features: CpuFeatures::default(),
+            heap_settings: HeapSettings::default(),
+            count_instructions: false,
+            canonicalize_nans: false,
+            validator: None,
+        }
+    }
+
+    pub fn target(&mut self, target: Triple) -> &mut Self {
```
katie-martin-fastly: As you mentioned above, this could follow the `LucetcOpts` conventions :slightly_smiling_face: 

Instead of having this borrow the struct, we can have it consume `self`, and return an owned type, via the signature below. This just means passing ownership of the object along, instead of passing along the ability to mutate the object. Call-sites shouldn't change _too_ much. :crossed_fingers: 

```rust
pub fn with_target(self, target: Triple) -> Self
```

lucetc/src/compiler.rs
```diff
@@ -47,6 +47,91 @@ impl OptLevel {
...
+    }
+
+    pub fn cpu_features(&mut self, cpu_features: CpuFeatures) -> &mut Self {
+        self.cpu_features = cpu_features;
+        self
+    }
+
+    pub fn cpu_features_mut(&mut self) -> &mut CpuFeatures {
+        &mut self.cpu_features
+    }
```
katie-martin-fastly: Building off of the `with_` changes, I would change the signature of these mutator methods to look like

```rust
cpu_features(&mut self, cpu_features: CpuFeatures)
```

This means code that changes the value of a member, `CpuFeatures` in this case, does so by passing the new value _to_ our builder rather than borrowing and mutating the existing value.

lucetc/src/lib.rs
```diff
@@ -248,13 +256,26 @@ impl<T: AsLucetc> LucetcOpts for T {
...
+    fn canonicalize_nans(&mut self, canonicalize_nans: bool) {
+        self.as_lucetc()
+            .builder
+            .canonicalize_nans(canonicalize_nans);
+    }
+
+    fn with_canonicalize_nans(mut self, canonicalize_nans: bool) -> Self {
+        self.canonicalize_nans(canonicalize_nans);
+        self
+    }
```
katie-martin-fastly: Wonderful touch adding these as well :heart:  Thank you!

data-pup: Heads up that you will see a merge conflict here<sup>[1](https://github.com/bytecodealliance/lucet/pull/423/files#diff-a5ee6b9e6496674851a8b1023f579e5eR261)</sup>, now that #423 has landed.

1: https://github.com/bytecodealliance/lucet/pull/423/files#diff-a5ee6b9e6496674851a8b1023f579e5eR261

lucetc/src/compiler.rs
michael-groble: if I try this signature, then within `LucetOpts` impl, it complains that `CompilerBuilder` needs to implement `Copy`.  Lots of bits in there don't implement `Copy` (`Triple`, `CpuFeatures`, `HeapSettings`) so I'm wondering if I'm missing something about how to do this.  I see how the impl of `LucetOpts` uses the following signature

 ```rust
fn with_target(mut self, target: Triple) -> Self
```
but not sure why it doesn't need the `Copy`. Can dig more.

File: lucetc/src/compiler.rs (hunk 6 above)
Comment (michael-groble) [reply lucetc/src/compiler.rs]: so if you look at the example of how we expose `heap_settings` to enable `min_reserved_size`, `LucetOpts` wants to be able to manipulate _parts_ of `heap_settings`, not wholly overwrite it, which is what a signature like that would provide.  Alternately, I could not expose `heap_settings` at all (like `LucetOpts` doesn't) and instead expose each of the individual modifiable parts (although that is opposite of the pattern it uses with `cpu_features`).  I was trying to find a consistent way to handle both `cpu_features` and `heap_settings` where they don't seem to be handled consistently in `LucetOpts`.

File: lucetc/src/compiler.rs (hunk 5 above)
Comment (michael-groble) [reply lucetc/src/compiler.rs]: ok, i think i have this how you're asking @katie-martin-fastly.  Note there is still the `_mut` versions as below.  as before, there are a lot more places to clean up if this looks good.

File: lucetc/src/compiler.rs (hunk 6 above)
Comment (katie-martin-fastly) [reply lucetc/src/compiler.rs]: Ah! Thank you for noticing this, it is very thoughtful.

I think the way you have this currently is great! Sometimes exceptions are needed, so an additional `cpu_features_mut` method because of implementation details seems perfectly reasonable to me. The  `min_reserved_size` methods feel more like concerns of the public API, so I don't think this struct needs to mirror that.

File: lucetc/src/lib.rs [hunk 8]
```diff
@@ -380,7 +357,7 @@ impl Lucetc {
         let dir = tempfile::Builder::new().prefix("lucetc").tempdir()?;
         let objpath = dir.path().join("tmp.o");
         self.object_file(objpath.clone())?;
-        link_so(objpath, &self.target, &output)?;
+        link_so(objpath, self.builder.target_ref(), &output)?;
```
katie-martin-fastly: You mentioned a similar thing with `heap_settings_mut`, I think the same thing applies here. As well. Having an extra method doesn't hurt us, especially if it's scoped with `pub(crate)`

lucetc/src/compiler.rs
katie-martin-fastly: I commented elsewhere about this, but the `target_ref` method is another case like this. I think it's fine to have these, and the rest of the methods are there to provide a nice API surface for general-use.

michael-groble:
>  Do you mean that there is more you'd like to do here? If you are satisfied, I think this is already a great step in the right direction.

I just meant the last commit I just added where I clean up uses of `Compiler::new` in the tests, mostly in `lucetc/test/wasm.rs`.  With that, I think this is complete.  Really appreciate the guidance on this @katie-martin-fastly!

PR #421, opened 2020-02-12
Issue #417: Retain whether or not a module was compiled with --count-instructions.

### LINKED ISSUE #417: Modules should record if instruction count instrumentation is present
Back when we merged support for instruction counts we didn't have an extensible way to record what optional features were enabled, and so the only way to know if the runtime can rely on instruction counting is seeing if the count is zero or not.

These days, we have exactly such a mechanism with [ModuleFeatures](https://github.com/bytecodealliance/lucet/blob/5b1b33bbe03e4f0405e6803b86f34309ee0afbe7/lucet-module/src/module_data.rs#L50-L62), so we should pass the existence (or lack of) a `--count-instructions` flag from lucetc to this struct for later reference.

fst-crenshaw:
This PR uses the `--count-instructions` option to set a new field in ModuleFeatures called `icnt`.  

**Notes:**
Leveraging the new field, the lucet-runtime call `get_instruction_count()` now returns an Option<u64>.  If the module was compiled with `--count-instructions`, the function returns the instruction count.  Otherwise, it returns None.

An extended set of tests affirm that get_instruction_count() appropriately returns Some(_) or None.

**See original issue at:**
https://github.com/bytecodealliance/lucet/issues/417

Fixes #417

### DIFF
--- a/lucet-module/src/module_data.rs
+++ b/lucet-module/src/module_data.rs
@@ -74,7 +74,7 @@ impl ModuleFeatures {
             bmi2: false,
             lzcnt: false,
             popcnt: false,
-	    icnt: false,  // TLC Instruction Count
+            icnt: false,
             _hidden: (),
         }
     }
--- a/lucet-module/src/module_data.rs
+++ b/lucet-module/src/module_data.rs
@@ -58,6 +58,7 @@ pub struct ModuleFeatures {
     pub bmi2: bool,
     pub lzcnt: bool,
     pub popcnt: bool,
+    pub icnt: bool,
     _hidden: (),
 }
 
@@ -73,6 +74,7 @@ impl ModuleFeatures {
             bmi2: false,
             lzcnt: false,
             popcnt: false,
+	    icnt: false,  // TLC Instruction Count
             _hidden: (),
         }
     }
--- a/lucet-module/src/module_data.rs
+++ b/lucet-module/src/module_data.rs
@@ -58,7 +58,7 @@ pub struct ModuleFeatures {
     pub bmi2: bool,
     pub lzcnt: bool,
     pub popcnt: bool,
-    pub icnt: bool,
+    pub instruction_count: bool,
     _hidden: (),
 }
 
@@ -74,7 +74,7 @@ impl ModuleFeatures {
             bmi2: false,
             lzcnt: false,
             popcnt: false,
-            icnt: false,
+            instruction_count: false,
             _hidden: (),
         }
     }
--- a/lucet-runtime/tests/instruction_counting.rs
+++ b/lucet-runtime/tests/instruction_counting.rs
@@ -7,10 +7,10 @@ use std::path::Path;
 use std::sync::Arc;
 use tempfile::TempDir;
 
-pub fn wasm_test<P: AsRef<Path>>(wasm_file: P, icnt_option: bool) -> Result<Arc<DlModule>, Error> {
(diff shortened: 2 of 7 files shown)


lucet-module/src/module_data.rs
```diff
@@ -58,6 +58,7 @@ pub struct ModuleFeatures {
     pub bmi2: bool,
     pub lzcnt: bool,
     pub popcnt: bool,
+    pub icnt: bool,
```
iximeow: `lzcnt` and `popcnt` only have the abbreviated names to match [CPUID feature bits](https://en.wikipedia.org/wiki/Bit_Manipulation_Instruction_Sets#ABM) by the same name. I think you're entirely justified in using a more human friendly name like `instruction_count` here.

lucet-runtime/tests/instruction_counting.rs
```diff
@@ -33,7 +32,40 @@ pub fn check_instruction_counts() {
             );
             ent
         })
-        .collect();
+        .collect()
+}
+
+#[test]
+pub fn check_icnt_off() {
```
iximeow: :tada: effective test reuse!

lucet-module/src/module_data.rs
fst-crenshaw: Thanks for this feedback!  I'll change the name.

lucet-runtime/tests/instruction_counting.rs
fst-crenshaw: My thanks to you for writing a easy-to-extend test!

lucet-runtime/tests/instruction_counting.rs
```diff
@@ -53,9 +90,12 @@ pub fn check_instruction_counts() {
         inst.run("test_function", &[]).expect("instance runs");
 
         let instruction_count = inst.get_instruction_count();
+        if let None = instruction_count {
+            panic!("instruction count expected from instance");
```
iximeow: just a style thing: this `if` could be folded in with the `unwap` below into `instruction_count.expect("has instruction counting")`

fst-crenshaw: Hm.  I attempted something like that but got a compiler message like "`expect` not available for u64`.  Possible problem between me and the keyboard.  I may be missing something.

lucet-runtime/tests/instruction_counting.rs
```diff
@@ -33,7 +35,41 @@ pub fn check_instruction_counts() {
...
+        let region = MmapRegion::create(1, &Limits::default()).expect("region can be created");
+
+        let mut inst = region
+            .new_instance(module)
+            .expect("instance can be created");
+
+        inst.run("test_function", &[]).expect("instance runs");
+
+        let instruction_count = inst.get_instruction_count();
+        if let Some(_) = instruction_count {
```
katie-martin-fastly: For situations where you are disregarding the inner value, (i.e. `Some(_)` or `Ok(_)`) there are also `is_some` and `is_none` methods! You can see more about `is_some` here: https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some

So, this condition could instead look like:

```suggestion
        if instruction_count.is_some() {
```

katie-martin-fastly: this is getting into _Real Pedantic Hours_™ however, so this isn't something I'd consider blocking by any means.

katie-martin-fastly: OOPS

This is an outdated diff, you've already fixed this.

PR #426, opened 2020-02-16
Check extended CPU features only when necessary as querying flags fail on some AMD CPUs.

shravanrn:
Some AMD CPUs fail on querying CPUID extended flags. See https://bugzilla.mozilla.org/show_bug.cgi?id=1615786. Performing this check only when actually required, allows wasm modules that don't rely on extended features to be run.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/module/dl.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/module/dl.rs
@@ -48,24 +48,28 @@ fn check_feature_support(module_features: &ModuleFeatures) -> Result<(), Error> 
         return Err(missing_feature("POPCNT"));
     }
 
-    let info = cpuid.get_extended_feature_info().ok_or_else(|| {
-        Error::Unsupported("Unable to obtain host CPU extended feature info!".to_string())
-    })?;
+    if module_features.bmi1 || module_features.bmi2 {
+        let info = cpuid.get_extended_feature_info().ok_or_else(|| {
+            Error::Unsupported("Unable to obtain host CPU extended feature info!".to_string())
+        })?;
 
-    if module_features.bmi1 && !info.has_bmi1() {
-        return Err(missing_feature("BMI1"));
+        if module_features.bmi1 && !info.has_bmi1() {
+            return Err(missing_feature("BMI1"));
+        }
+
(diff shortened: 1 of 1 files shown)


lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -48,24 +48,28 @@ fn check_feature_support(module_features: &ModuleFeatures) -> Result<(), Error>
         return Err(missing_feature("POPCNT"));
     }
 
-    let info = cpuid.get_extended_feature_info().ok_or_else(|| {
-        Error::Unsupported("Unable to obtain host CPU extended feature info!".to_string())
-    })?;
+    if module_features.bmi1 || module_features.bmi2 {
```
pchickey:
```suggestion
   // Some AMD CPUs fail on querying CPUID extended flags. See https://bugzilla.mozilla.org/show_bug.cgi?id=1615786. 
    if module_features.bmi1 || module_features.bmi2 {
```

lucet-runtime/lucet-runtime-internals/src/module/dl.rs
```diff
@@ -48,24 +48,28 @@ fn check_feature_support(module_features: &ModuleFeatures) -> Result<(), Error>
...
-        return Err(missing_feature("BMI2"));
+        if module_features.bmi2 && !info.has_bmi2() {
+            return Err(missing_feature("BMI2"));
+        }
     }
 
-    let info = cpuid.get_extended_function_info().ok_or_else(|| {
-        Error::Unsupported("Unable to obtain host CPU extended function info!".to_string())
-    })?;
+    if module_features.lzcnt {
```
pchickey:
```suggestion
    // Same situation - see above
    if module_features.lzcnt {
```

pchickey:
I messed up the review thing - looks good, if you can approve adding those suggested comments we'll get it merged.

iximeow:
pt2: on account of not being AMD-specific, I'm going to squash merge and rewrite the commit message to reference the bug report and having made a stronger assumption than we should about CPUID leaf availability.

shravanrn:
Ah sorry, was away all day so just saw the messages now. That makes sense. Thanks @pchickey and @iximeow!

PR #428, opened 2020-02-19
Fix Sightglass' Lucet flags

iximeow:
specifically, pass correct form of "as optimized as possible please" flag to lucet

### DIFF
--- a/benchmarks/shootout/Makefile
+++ b/benchmarks/shootout/Makefile
@@ -18,7 +18,7 @@ SHOOTOUT_SRCS:=$(shell ls $(SHOOTOUT)/*.c)
 SHOOTOUT_NATIVE_OBJS:=
 SHOOTOUT_LUCET_OBJS:=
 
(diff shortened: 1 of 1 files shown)


fst-crenshaw:
Curious: Why not just `speed`?

awortman-fastly:
The simple answer: because once upon a time this was ["best"](https://github.com/bytecodealliance/lucet/commit/d45a2aa7eb3eef03af1dac265ce6cef88cefda51#diff-5d226338f361546cf8921b0c019df026) and "speed_and_size" these days maps to "ok cranelift, do the best you can do".

As for why we'd pick `speed_and_size` rather than `speed` when we're measuring... speed: so far as I know, `speed_and_size` won't do anything that would regress speed in favor of size, but does get cranelift to run through instruction squeezing that can do things like reduce offset sizes from four to one (or no!) byte, remove prefixes that aren't necessary, and other optimizations of that nature. This _probably_ doesn't make a measurable impact, but if we're doing the best we can vs the best gcc can, it doesn't hurt to try.

PR #429, opened 2020-02-19
KillSwitch terminates pending instance

data-pup:
### Overview

This updates the `lucet_runtime_internals::instance::execution` module so that `KillSwitch::terminate` will also terminate pending instances.

_Before_ if an instance was remotely terminated _before_ running, the result would be a `KillError::NotTerminable` error. Moreover, the instance would still run a guest context if `run` was called afterwards.

_Now_ if an instance is remotely terminated _before_ running, the result will now be `Ok(KillSuccess::Cancelled)` rather than a `KillError::NotTerminable` error. The instance will then enter a `Domain::Cancelled` state, and will not enter a guest region when `run` is called.

### Example

This is best summarized in a small snippet, assuming an initialized instance `inst`:

```rust
let inst: Instance = // ...
let kill_switch = inst.kill_switch();
assert_eq!(kill_switch.terminate(), Ok(KillSuccess::Cancelled));

// if terminated before running, the guest should not start at all
match inst.run("onetwothree", &[]) {
    Err(Error::RuntimeTerminated(TerminationDetails::Remote)) => {}
    res => panic!("unexpected result: {:?}", res),
}
```

### Implementation

In order for this to happen, it means that an instance is _now_ considered terminable when it is not running. In the original implementation, `terminable` was _really_ a flag that meant that the instance could be safely signalled. (_Recall: Depending on the current execution domain, it may or may not be safe to signal the instance to terminate_)

If the instance is now terminable when the guest context is not active (i.e. it has not yet started), we now need to indicate whether or not the instance can be safely signalled using the execution domain. This means touching some of the :ghost: fun :ghost: parts of `lucet`.

The **TLDR** of the changes here is:
* `Domain` now has two additional variants: `Pending`, and `Cancelled`. A new instance will be in the `Pending` state. If a kill switch calls its `terminate` method before the instance runs, that instance will be left in the `Cancelled` domain.
* Signal behavior now depends on `KillState::execution_domain`.
  * The `terminability` member is no longer semantically useful. (_This_ part, is the :hot_pepper: spicy part.) 
  * `lucet_context_activate` should call a Rust function to lock the domain `Mutex`, and update the execution domain before jumping to guest code. This means we now have a `enter_guest_region` that corresponds to the `exit_guest_region` that we use as a backstop callback.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -121,8 +121,8 @@ pub struct Context {
     retval_fp: __m128,
     parent_ctx: *mut Context,
     // TODO ACF 2019-10-23: make Instance into a generic parameter?
-    backstop_callback: *const unsafe extern "C" fn(*mut Instance),
-    callback_data: *mut Instance,
+    backstop_callback: *const unsafe extern "C" fn(*const Instance),
+    callback_data: *const Instance,
     sigset: signal::SigSet,
 }
 
@@ -142,7 +142,7 @@ impl Context {
     }
 
     /// Get a raw pointer to the instance's callback data.
-    pub(crate) fn callback_data_ptr(&self) -> *mut Instance {
+    pub(crate) fn callback_data_ptr(&self) -> *const Instance {
         self.callback_data
     }
 }
@@ -382,7 +382,7 @@ impl Context {
     }
 
     /// The default backstop callback does nothing, and is just a marker.
-    extern "C" fn default_backstop_callback(_: *mut Instance) {}
+    extern "C" fn default_backstop_callback(_: *const Instance) {}
 
     /// Similar to `Context::init()`, but allows setting a callback function to be run when the
     /// guest entrypoint returns.
@@ -392,8 +392,8 @@ impl Context {
     pub fn init_with_callback(
         stack: &mut [u64],
         child: &mut Context,
-        backstop_callback: unsafe extern "C" fn(*mut Instance),
-        callback_data: *mut Instance,
+        backstop_callback: unsafe extern "C" fn(*const Instance),
+        callback_data: *const Instance,
         fptr: usize,
         args: &[Val],
     ) -> Result<(), Error> {
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -728,9 +728,25 @@ extern "C" {
     /// Never returns because the current context is discarded.
     fn lucet_context_set(to: *const Context) -> !;
 
-    /// Enables termination for the instance, after performing a context switch.
+    /// Runs an activation function after performing a context switch. Implemented in assembly.
     ///
-    /// Takes the guest return address as an argument as a consequence of implementation details,
-    /// see `Instance::swap_and_return` for more.
+    /// In practice, this is used with `enter_guest_region` so that the guest will appropriately
+    /// set itself to be terminable upon entry before continuing to any guest code.
+    ///
+    /// `lucet_context_activate` is essentially a function with three arguments:
+    ///   * rdi: the data for the entry function.
+    ///   * rsi: the address of the entry function.
+    ///   * rbx: the address of the guest code to resume at.
+    ///
+    /// We do not actually define `lucet_context_activate` as having these arguments because we
+    /// manually load these arguments, as well as a pointer to this function, into the context's
+    /// registers. See `Instance::with_activation_routine` for more information.
+    ///
+    /// Note that `rbx` is used to store the address of the guest code because it is a callee-saved
+    /// register in the System V calling convention. It is also a non-violatile register on
+    /// Windows, which is a nice additional benefit.
+    ///
+    /// For more information, see `Instance::swap_and_return`, `Instance::with_activation_routine`,
+    /// `enter_guest_region`, and `lucet_context_activate`'s assembly implementation.
     pub(crate) fn lucet_context_activate();
 }
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -26,7 +26,7 @@ use thiserror::Error;
 /// `u64`, this should be fine?
 #[repr(C)]
 pub(crate) struct GpRegs {
-    rbx: u64,
+    pub(crate) rbx: u64,
     pub(crate) rsp: u64,
     rbp: u64,
     pub(crate) rdi: u64,
@@ -122,7 +122,8 @@ pub struct Context {
     parent_ctx: *mut Context,
     // TODO ACF 2019-10-23: make Instance into a generic parameter?
     backstop_callback: *const unsafe extern "C" fn(*mut Instance),
-    backstop_data: *mut Instance,
+    // FIXME KTM 2020-02-14: this should NOT be pub(crate)
+    pub(crate) backstop_data: *mut Instance,
     sigset: signal::SigSet,
 }
 
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -142,8 +142,6 @@ impl Context {
     }
 
     /// Get a raw pointer to the instance's callback data.
-    ///
-    /// REVIEW QUESTION: Should this be `unsafe fn` ? Can aliasing `*mut` lead to UB?
     pub(crate) fn callback_data_ptr(&self) -> *mut Instance {
         self.callback_data
     }
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -728,15 +728,15 @@ extern "C" {
     /// Never returns because the current context is discarded.
     fn lucet_context_set(to: *const Context) -> !;
 
-    /// Runs an activation function after performing a context switch. Implemented in assembly.
+    /// Runs an entry callback after performing a context switch. Implemented in assembly.
     ///
     /// In practice, this is used with `enter_guest_region` so that the guest will appropriately
     /// set itself to be terminable upon entry before continuing to any guest code.
     ///
     /// `lucet_context_activate` is essentially a function with three arguments:
-    ///   * rdi: the data for the entry function.
-    ///   * rsi: the address of the entry function.
-    ///   * rbx: the address of the guest code to resume at.
+    ///   * rdi: the data for the entry callback.
+    ///   * rsi: the address of the entry callback.
+    ///   * rbx: the address of the guest code to execute.
     ///
     /// We do not actually define `lucet_context_activate` as having these arguments because we
     /// manually load these arguments, as well as a pointer to this function, into the context's
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -121,8 +121,8 @@ pub struct Context {
     retval_fp: __m128,
     parent_ctx: *mut Context,
     // TODO ACF 2019-10-23: make Instance into a generic parameter?
-    backstop_callback: *const unsafe extern "C" fn(*const Instance),
-    callback_data: *const Instance,
+    backstop_callback: *const unsafe extern "C" fn(*mut Instance),
+    callback_data: *mut Instance,
     sigset: signal::SigSet,
 }
 
@@ -142,7 +142,7 @@ impl Context {
     }
 
     /// Get a raw pointer to the instance's callback data.
-    pub(crate) fn callback_data_ptr(&self) -> *const Instance {
+    pub(crate) fn callback_data_ptr(&self) -> *mut Instance {
         self.callback_data
     }
 }
@@ -382,7 +382,7 @@ impl Context {
     }
 
     /// The default backstop callback does nothing, and is just a marker.
-    extern "C" fn default_backstop_callback(_: *const Instance) {}
+    extern "C" fn default_backstop_callback(_: *mut Instance) {}
 
     /// Similar to `Context::init()`, but allows setting a callback function to be run when the
     /// guest entrypoint returns.
@@ -392,8 +392,8 @@ impl Context {
     pub fn init_with_callback(
         stack: &mut [u64],
         child: &mut Context,
-        backstop_callback: unsafe extern "C" fn(*const Instance),
-        callback_data: *const Instance,
+        backstop_callback: unsafe extern "C" fn(*mut Instance),
+        callback_data: *mut Instance,
         fptr: usize,
         args: &[Val],
     ) -> Result<(), Error> {
--- a/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/context/mod.rs
@@ -122,8 +122,7 @@ pub struct Context {
     parent_ctx: *mut Context,
     // TODO ACF 2019-10-23: make Instance into a generic parameter?
     backstop_callback: *const unsafe extern "C" fn(*mut Instance),
-    // FIXME KTM 2020-02-14: this should NOT be pub(crate)
-    pub(crate) backstop_data: *mut Instance,
+    callback_data: *mut Instance,
     sigset: signal::SigSet,
 }
 
@@ -137,9 +136,16 @@ impl Context {
             retval_fp: unsafe { _mm_setzero_ps() },
             parent_ctx: ptr::null_mut(),
             backstop_callback: Context::default_backstop_callback as *const _,
-            backstop_data: ptr::null_mut(),
+            callback_data: ptr::null_mut(),
             sigset: signal::SigSet::empty(),
         }
+    }
+
+    /// Get a raw pointer to the instance's callback data.
+    ///
+    /// REVIEW QUESTION: Should this be `unsafe fn` ? Can aliasing `*mut` lead to UB?
+    pub(crate) fn callback_data_ptr(&self) -> *mut Instance {
+        self.callback_data
     }
 }
 
@@ -384,12 +390,12 @@ impl Context {
     /// guest entrypoint returns.
     ///
     /// After the entrypoint function returns, but before swapping back to the parent context,
-    /// `backstop_callback` will be run with the single argument `backstop_data`.
+    /// `backstop_callback` will be run with the single argument `callback_data`.
     pub fn init_with_callback(
         stack: &mut [u64],
         child: &mut Context,
         backstop_callback: unsafe extern "C" fn(*mut Instance),
-        backstop_data: *mut Instance,
+        callback_data: *mut Instance,
         fptr: usize,
         args: &[Val],
     ) -> Result<(), Error> {
@@ -399,7 +405,7 @@ impl Context {
 
         if backstop_callback != Context::default_backstop_callback {
             child.backstop_callback = backstop_callback as *const _;
-            child.backstop_data = backstop_data;
+            child.callback_data = callback_data;
         }
 
         let mut gp_args_ix = 0;
--- a/lucet-runtime/lucet-runtime-internals/src/instance.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance.rs
@@ -870,9 +870,9 @@ impl Instance {
     /// terminable before continuing to whatever guest code we want to run.
     ///
     /// `lucet_context_activate` takes three arguments in the following registers:
-    ///   * rdi: the data for the entry function
-    ///   * rsi: the address of the entry function
-    ///   * rbx: the address of guest code to execute
+    ///   * rdi: the data for the entry callback.
+    ///   * rsi: the address of the entry callback.
+    ///   * rbx: the address of the guest code to execute.
     ///
     /// The appropriate value for `rbx` is the top of the guest stack, which we would otherwise
     /// return to and start executing immediately. For `rdi`, we want to pass our callback data
--- a/lucet-runtime/lucet-runtime-internals/src/instance.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance.rs
@@ -533,10 +533,15 @@ impl Instance {
     /// modified by this call; it is the embedder's responsibility to clear or reset their state if
     /// necessary.
     ///
+    /// This will also reinitialize the kill state, which means that any outstanding
+    /// [`KillSwitch`](struct.KillSwitch.html) objects will be unable to terminate this instance.
+    /// It is the embedder's responsibility to initialize new killswitches after resetting an
+    /// instance.
+    ///
     /// # Safety
     ///
-    /// This function runs the guest code for the WebAssembly `start` section, and running any guest
-    /// code is potentially unsafe; see [`Instance::run()`](struct.Instance.html#method.run).
+    /// This function runs the guest code for the WebAssembly `start` section, and running any
+    /// guest code is potentially unsafe; see [`Instance::run()`](struct.Instance.html#method.run).
     pub fn reset(&mut self) -> Result<(), Error> {
         self.alloc.reset_heap(self.module.as_ref())?;
         let globals = unsafe { self.alloc.globals_mut() };
@@ -554,9 +559,8 @@ impl Instance {
         }
 
         self.state = State::Ready;
-
+        self.kill_state = Arc::new(KillState::new());
         self.run_start()?;
-
         Ok(())
     }
 
@@ -869,14 +873,26 @@ impl Instance {
         //
         // once we've set up arguments, swap out the guest return address with
         // `lucet_context_activate` so we start execution there.
+        //
+        // DEV KTM:
+        // in rdi, the data for the setup/frontstop
+        // in rsi, the address of the steup/"frontstop" callback
+        // in rbx, the address of the guest code to resume at.
         unsafe {
             let top_of_stack = self.ctx.gpr.rsp as *mut u64;
-            // move the guest code address to rsi
-            self.ctx.gpr.rsi = *top_of_stack;
+            // move the guest code address to rbx
+            self.ctx.gpr.rbx = *top_of_stack;
             // replace it with the activation thunk
             *top_of_stack = crate::context::lucet_context_activate as u64;
             // and store a pointer to indicate we're active
-            self.ctx.gpr.rdi = self.kill_state.terminable_ptr() as u64;
+            self.ctx.gpr.rdi = self.ctx.backstop_data as u64;
+            // DEV KTM: Let's try and pass a function pointer!
+            self.ctx.gpr.rsi = execution::enter_guest_region as u64;
+            // self.ctx.gpr.rbx = 0xFFFF;
+
+            // DEV KTM:
+            // Let's make `Domain` a u64, explicitly tag the pending and guest invariants.
+            // Then, we can change which invariant it is?
         }
 
         self.swap_and_return()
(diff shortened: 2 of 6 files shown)


lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -125,6 +127,41 @@ pub struct KillState {
...
+pub unsafe extern "C" fn enter_guest_region(instance: *mut Instance) {
+    let mut current_domain = (*instance).kill_state.execution_domain.lock().unwrap();
+    match *current_domain {
+        // We are about to start executing a guest. Set the execution domain accordingly.
+        Domain::Pending => *current_domain = Domain::Guest,
+        // We are about to enter guest code for an instance that has already been terminated.
+        // Instead, we should forgo entering the guest region entirely, terminate the instance, and
+        // swap back to the host without any unwinding.
+        Domain::Cancelled => (*instance).terminate(TerminationDetails::Remote),
+        // If we are in some other domain, do nothing. These variants are handled elsewhere.
```
iximeow: I think it would in fact be an error to reach this point in any other domain:
* `Guest` would be a no-no because that suggests we might send a `SIGALRM` before we've indicated it safe to do so by `enable_termination`
* `Terminated` would be a no-no because this would be a zombie ("this has been stopped, but we're just starting again?")
* `Hostcall` would suggest that a host has called into guest code, which isn't bad on its own, but that should _not_ involve going through `enter_guest_region` a second time

so I think you're perfectly justified in making this arm a `panic`, since reaching this would be some form of implementation error.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -125,6 +127,41 @@ pub struct KillState {
...
+/// not return, and we will swap back to the host context without unwinding.
+pub unsafe extern "C" fn enter_guest_region(instance: *mut Instance) {
+    let mut current_domain = (*instance).kill_state.execution_domain.lock().unwrap();
+    match *current_domain {
+        // We are about to start executing a guest. Set the execution domain accordingly.
+        Domain::Pending => *current_domain = Domain::Guest,
+        // We are about to enter guest code for an instance that has already been terminated.
+        // Instead, we should forgo entering the guest region entirely, terminate the instance, and
+        // swap back to the host without any unwinding.
+        Domain::Cancelled => (*instance).terminate(TerminationDetails::Remote),
```
data-pup: This function has some strong warnings in its comments ("_This is almost certainly not what you want to use to terminate from a hostcall_") so I wanted to make sure this wasn't a total error. In this case though, we don't want to unwind because there isn't anything _wound_!

https://github.com/bytecodealliance/lucet/blob/ee003cd27643a4ad5a745cd3a4d02d5315d3a8b2/lucet-runtime/lucet-runtime-internals/src/vmctx.rs#L425-L429

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -273,6 +275,21 @@ macro_rules! timeout_tests {
             t.join().unwrap();
         }
 
+        // after resetting, existing kill swtiches will not work. we can still run a function.
+        // FIXME KTM 2020-02-19: I did not see a way around this behavior. Bring this up in review.
```
iximeow: I think it would end up surprising if a `KillSwitch` from some prior `reset` were to still work, especially since they have a `operates only once` notion when they _do_ work. Being able to reuse an old `KillSwitch` to reportedly `NotTerminable` an instance is also not what I imagine would be expected either, so maybe it makes sense to have a `KillError::Invalid` to communicate having used a stale resource?

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -533,10 +533,15 @@ impl Instance {
     /// modified by this call; it is the embedder's responsibility to clear or reset their state if
     /// necessary.
     ///
+    /// This will also reinitialize the kill state, which means that any outstanding
+    /// [`KillSwitch`](struct.KillSwitch.html) objects will be unable to terminate this instance.
+    /// It is the embedder's responsibility to initialize new killswitches after resetting an
+    /// instance.
```
data-pup: [This change](https://github.com/bytecodealliance/lucet/pull/429/files#diff-a99c9cb71117942c648b057de7d9a29aR562) is responsible for this, but I wanted to highlight this as a change that I feel _weird_ about. I think there's a better way forward that I'm just not seeing.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -29,8 +29,8 @@
 //! possible points of interest in an instance's lifecycle:
 //!
 //! * `Instance created`
-//!   - terminable: `false`
-//!   - execution_domain: `Guest`
+//!   - terminable: `true`
+//!   - execution_domain: `Pending`
```
data-pup: If you wanted a two line summary of this change, this is it. It would be very useful if an instance is terminable from creation onward. Before running, it will be in the `Pending` domain.

And as mentioned in the `FIXME` below, I will make a pass at updating this before merging.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
iximeow: I think "_almost certainly not what you want to use_" is more a warning to hostcall implementors because resources like Mutex guards wouldn't get dropped, possibly leading to deadlocks in the host after exiting. Because no custom user code has run at this point, guest or host, this should be fine.

data-pup: Figured that was the case, but those were strong enough words that I wanted to be sure :slightly_smiling_face: Sounds good!

lucet-runtime/lucet-runtime-internals/src/instance.rs
data-pup: https://github.com/bytecodealliance/lucet/pull/429#discussion_r381623306

Discussion about this continues over here. Resolving this.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
data-pup: We talked about this out-of-band, but for posterity: Yes! This totally makes sense. All of these cases would mean that something has gone terribly wrong. I added panic statements here.

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -245,12 +247,29 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
...
+                    // If we are here, we should still be in the guest domain.
+                    debug_assert_eq!(*current_domain, execution::Domain::Guest);
+                    *current_domain = execution::Domain::Terminated;
+                }
+                Err(poison_error) => {
+                    // Because a signal must never be sent while the host context is active, we
+                    // will update the domain even if the mutex has been poisoned.
+                    let mut current_domain = poison_error.into_inner();
+                    *current_domain = execution::Domain::Terminated;
+                }
```
data-pup: So :point_up:  this whole block is one of the places where pulling `terminable` out _does_ make things more complicated. Given that this is inside of a signal handler, what should we do here?

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -245,12 +247,29 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
...
+        // If we are switching to the host context and we are not processing a SIGALRM from a
+        // KillSwitch, then we were either externally signalled in some other manner, or we faulted
+        // during execution in a guest region.
+        //
+        // In either case, we will need to unlock and update the execution domain ourselves.
         if switch_to_host {
-            // we must disable termination so no KillSwitch may fire in host code.
-            inst.kill_state.disable_termination();
+            // This debug assert exists to make sure that this is not a blocking operation.
+            debug_assert_eq!(inst.kill_state.domain_is_locked(), false);
```
iximeow: A `KillSwitch` could fire exactly after this assert, but before caring about `execution_domain`, so it's not exactly correct :grimacing:

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
iximeow: We don't need it! :grin: There's a comment above re. `domain_is_locked`, but there's another concern this introduces: we could be handling a to-be-fatal `SIGSEGV` right at this point having grabbed the lock when the guest times out. Then, the `SIGALRM` causes us to re-enter this function, would also `switch_to_host`, reach here, attempt to grab the lock again, and deadlock.

I think the signal handler will be fine in not updating `execution_domain`, and if we really want to set it to `Terminated`, we should do so in `KillSwitch` once the `tid_change_notifier` has done its thing. Even if we don't, a second `KillSwitch` wouldn't signal, because the shared `thread_id` would have been updated to `None`, but I think reiterating that the referenced instance is terminated is good for clarity.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -856,27 +860,28 @@ impl Instance {
...
+        //   * rdi: pointer to the data for the entry function
+        //   * rsi: the address of the entry function
+        //   * rbx: the address of guest code to execute
         //
-        // The appropriate value for `rsi` is the top of the guest stack, which
-        // we would otherwise return to and start executing immediately. For
-        // `rdi`, we want to pass a pointer to the instance's `terminable` flag.
+        // The appropriate value for `rbx` is the top of the guest stack, which we would otherwise
+        // return to and start executing immediately. For `rdi`, we want to pass a raw pointer to
+        // the instance, which will be pass
```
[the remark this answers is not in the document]
katie-martin-fastly: My apologies, there are a few places in comments (like this) that I need to do a second pass over.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -217,12 +308,52 @@ impl KillState {
...
+    }
 }
 
+/// Instance execution domains.
+///
+/// This enum allow us to distinguish what an appropriate mechanism to signal termination is.
+#[derive(Debug, PartialEq)]
 pub enum Domain {
+    /// Represents an instance that is not currently running.
+    /// FIXME KTM 2020-02-20: Maybe we should call this `Ready` or `Paused`.
```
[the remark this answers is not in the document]
data-pup: Agh, I didn't see this comment here! I responded to this elsewhere, but I think pending is deceptively named because we return to this state after running.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -263,14 +395,8 @@ impl KillSwitch {
         // discarded, so we can't terminate.
         let state = self.state.upgrade().ok_or(KillError::NotTerminable)?;
 
-        // Attempt to take the flag indicating the instance may terminate
-        let terminable = state.terminable.swap(false, Ordering::SeqCst);
-        if !terminable {
-            return Err(KillError::NotTerminable);
-        }
-
         // we got it! we can signal the instance.
```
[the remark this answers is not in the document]
data-pup: A-ha. Fixed!

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -166,19 +167,20 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
...
-            debug_assert!(!inst.kill_state.is_terminable());
-
+            debug_assert!(inst.kill_state.domain_is_locked());
+            // set the state before jumping back to the host context
             inst.state = State::Terminating {
                 details: TerminationDetails::Remote,
             };
             return true;
-        }
+        };
```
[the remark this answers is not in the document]
data-pup: eagle eyes! :bird: 

This is definitely vestigial, and should be removed. I'm... actually shocked it is still here to be honest?

lucet-runtime/lucet-runtime-tests/src/timeout.rs
data-pup: Yeah, this totally makes sense! Because `Invalid` might be vague to a caller, I also added some doc comments to `KillError` to explain what we mean here.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -298,6 +424,7 @@ impl KillSwitch {
                     }
                     Ok(KillSuccess::Signalled)
                 } else {
+                    // If we are here, we probably faulted without updating the execution domain.
```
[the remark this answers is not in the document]
data-pup: We talked about this directly, and over [here](https://github.com/bytecodealliance/lucet/pull/429#discussion_r382876336), but we can avoid trying to lock the domain in the signal handler. As you mentioned, that invites a lot of trouble, and we can update the domain here.

The catch is that this arm _is_ then reachable, if we have faulted in guest code, as we won't have updated the domain from `Guest`.

data-pup:
This is ready for another round of review! After addressing your notes, this feels much more concise.  The documentation for the `lucet_runtime_internals::instance::execution` module is still stale, but I can take a pass updating that once the changes here look good :slightly_smiling_face:

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -244,14 +239,7 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
                 thunk();
                 true
             }
-        };
-
-        if switch_to_host {
-            // we must disable termination so no KillSwitch may fire in host code.
-            inst.kill_state.disable_termination();
```
data-pup: Talked about this here: https://github.com/bytecodealliance/lucet/pull/429#discussion_r382876336

Noting for posterity, since that discussion got orphaned via new changes.

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -100,10 +100,13 @@ macro_rules! timeout_tests {
 
             let kill_switch = inst.kill_switch();
 
-            thread::spawn(move || {
-                thread::sleep(Duration::from_millis(100));
-                assert_eq!(kill_switch.terminate(), Ok(KillSuccess::Signalled));
-            });
+            thread::Builder::new()
+                .name("helper".to_owned())
```
data-pup: Updated these tests to use [thread::Builder](https://doc.rust-lang.org/std/thread/struct.Builder.html) rather than the free `thread::spawn` function so that we can name the threads.

This makes for large quality of life improvements when debugging problems with these tests.

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -252,25 +269,42 @@ macro_rules! timeout_tests {
...
-            });
+            let helper = thread::Builder::new()
+                .name("helper".to_owned())
+                .spawn(move || {
+                    // This fires first, and will work properly, terminating the instance.
+                    assert_eq!(kill_switch.terminate(), Ok(KillSuccess::Cancelled));
+                    // Next, we will delay for 100ms and try to terminate the instance again.
+                    // This should fail, because we have already terminated the instance.
+                    thread::sleep(Duration::from_millis(100));
+                    assert_eq!(kill_switch.terminate(), Err(KillError::NotTerminable));
```
data-pup: specifying the `KillError` and `KillSuccess` variants for the sake of consistency.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -856,30 +860,41 @@ impl Instance {
...
+    ///   * rbx: the address of the guest code to execute.
+    ///
+    /// The appropriate value for `rbx` is the top of the guest stack, which we would otherwise
+    /// return to and start executing immediately. For `rdi`, we want to pass our callback data
+    /// (a raw pointer to the instance). This will be passed as the first argument to the entry
+    /// function, which is responsible for updating the kill state's execution domain.
+    ///
+    /// See `lucet_runtime_internals::context::lucet_context_activate`, and
+    /// `execution::enter_guest_region` for more info.
+    fn with_terminability(&mut self) -> &mut Self {
```
data-pup: :hot_pepper: So, _this_ is a spicy change, but it's a no-op in the name of documentation. :hot_pepper:

We have our activation thunk, our entry callback, and data to pass to the entry callback. Notably, the activation thunk does not _actually_ care what the entry callback does, or what data is passed to it! The activation thunk exists purely to call some callback, and then jump to an address.

_This_ block, which was previously inlined in `run_func`, is how that thunk is used to make the guest terminable upon entry. Putting this into a separate private method means we can document this a little more cleanly.

I am open to bikeshedding name here if you'd like, I know returning `&mut Self` means this isn't quite what we would canonically call `with_something`. I just liked that this means we can place this into `run_func` above inside of a line that looks like:

```rust
self.with_terminability().swap_and_return()
```

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -212,65 +265,61 @@ macro_rules! timeout_tests {
...
+        #[test]
+        fn timeout_after_guest_drop() {
+            let module = mock_timeout_module();
+            let region = TestRegion::create(1, &Limits::default()).expect("region can be created");
+            let mut inst = region
+                .new_instance(module)
+                .expect("instance can be created");
+            let kill_switch = inst.kill_switch();
+            std::mem::drop(inst);
+            assert_eq!(kill_switch.terminate(), Err(KillError::Invalid));
```
data-pup: Added a few corner cases while I was here.

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -212,65 +265,61 @@ macro_rules! timeout_tests {
...
+            run_onetwothree(&mut inst);
+        }
 
-            let t = thread::spawn(move || {
-                assert!(kill_switch.terminate().is_err()); // fails too soon
-                thread::sleep(Duration::from_millis(100));
-                assert!(kill_switch.terminate().is_ok()); // works
-                thread::sleep(Duration::from_millis(100));
-                assert!(kill_switch.terminate().is_err()); // fails too soon
```
data-pup: This case is covered by the other tests now :slightly_smiling_face:

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -1,20 +1,23 @@
 //! The `execution` module contains state for an instance's execution, and exposes functions
```
data-pup: All of the changes in the top-level module documentation are worth proofreading carefully. I tried to make sure this was all updated properly, but would love a second set of eyes here.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -25,58 +28,68 @@
...
+//!   - `execution_domain` may be `Terminated` before returning, in which case `terminate` will
+//!     return `NotTerminable`, but the hostcall would then exit. If a hostcall successfully
+//!     returns to its caller it was not terminated, so the only state an instance will have after
+//!     returning from a hostcall will be that it is executing terminable guest code.
+//! * `Instance::reset` or `Instance dropped`
+//!   - termination result: `Err(KillError::Invalid)`
+//!   - execution_domain: `Pending`
+//!   - If an instance is reset, outstanding `KillSwitch` objects will no longer hold a valid
+//!     reference to the instance. In that case, calls to `terminate` will return `Invalid`, and
+//!     have no bearing on further execution.
```
data-pup: Added this because it's worth highlighting, and it's part of why we have the `Invalid` variant of our kill error.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -125,54 +131,129 @@ pub struct KillState {
...
-    pub fn enable_termination(&self) {
-        self.terminable.store(true, Ordering::SeqCst);
-    }
-
-    pub fn disable_termination(&self) {
-        self.terminable.store(false, Ordering::SeqCst);
-    }
-
-    pub fn terminable_ptr(&self) -> *const AtomicBool {
-        &self.terminable as *const AtomicBool
```
data-pup: :tada: still very pumped about this

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -219,26 +312,62 @@ impl KillState {
     }
 }
 
+/// Instance execution domains.
+///
+/// This enum allow us to distinguish how to appropriately terminate an instance.
+///
+/// We can signal in guest code, but must avoid signalling in host code lest we interrupt some
+/// function operating on guest/host shared memory, and invalidate invariants. For example,
+/// interrupting in the middle of a resize operation on a `Vec` could be extremely dangerous.
```
data-pup: I added some documentation for the domain, and the hostcall methods on the kill state, since domain is now much more important w.r.t. terminability. Most of this was borrowed from existing comments etc., but I would also greatly appreciate eyes here :slightly_smiling_face:

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -97,61 +110,90 @@ macro_rules! timeout_tests {
...
                 .new_instance(module)
                 .expect("instance can be created");
+            let kill_switch = inst.kill_switch();
 
+            // The killswitch will fail if the instance has already finished running.
             match inst.run("do_nothing", &[]) {
                 Ok(_) => {}
                 res => panic!("unexpected result: {:?}", res),
             }
+            assert_eq!(kill_switch.terminate(), Ok(KillSuccess::Cancelled)); // FIXME -> Invalid
```
iximeow:
```suggestion
            assert_eq!(kill_switch.terminate(), Ok(KillSuccess::Cancelled));
```
`Cancelled` seems correct here, since you could try to run another function on the same instance, but if it timed out in the mean time the instance should refuse to.

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -281,32 +330,41 @@ macro_rules! timeout_tests {
                 .new_instance(module)
                 .expect("instance can be created");
 
+            // Spawn a thread to terminate the instance after waiting for 100ms.
             let kill_switch = inst.kill_switch();
+            let helper_1 = thread::Builder::new()
+                .name("helper-1".to_owned())
```
iximeow:
```suggestion
                .name("killswitch-1".to_owned())
```
(and the same for `helper-2`)? Mostly so the `expect` below are clearer about what did/didn't panic.

lucet-runtime/lucet-runtime-internals/src/context/mod.rs
```diff
@@ -723,9 +728,25 @@ extern "C" {
...
+    ///   * rsi: the address of the entry callback.
+    ///   * rbx: the address of the guest code to execute.
+    ///
+    /// We do not actually define `lucet_context_activate` as having these arguments because we
+    /// manually load these arguments, as well as a pointer to this function, into the context's
+    /// registers. See `Instance::with_activation_routine` for more information.
+    ///
+    /// Note that `rbx` is used to store the address of the guest code because it is a callee-saved
+    /// register in the System V calling convention. It is also a non-violatile register on
+    /// Windows, which is a nice additional benefit.
```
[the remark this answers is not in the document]
katie-martin-fastly: Oh that's a good question! The reference says: https://doc.rust-lang.org/reference/types/never.html

> NB. The never type was expected to be stabilized in 1.41, but due to some last minute regressions detected the stabilization was temporarily reverted. The ! type can only appear in function return types presently. See the tracking issue for more details.

So not on stable at the moment, unfortunately. (see https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9cec971186966ad0b1a4657e48ea0fc9 for a playground example)

I like this idea though :crying_cat_face:

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -25,58 +28,68 @@
...
-//!   was not terminated, so the only state an instance will have after returning from a hostcall
-//!   will be that it's executing terminable guest code.
-
+//!   - `execution_domain` may be `Terminated` before returning, in which case `terminate` will
+//!     return `NotTerminable`, but the hostcall would then exit. If a hostcall successfully
+//!     returns to its caller it was not terminated, so the only state an instance will have after
+//!     returning from a hostcall will be that it is executing terminable guest code.
+//! * `Instance::reset` or `Instance dropped`
+//!   - termination result: `Err(KillError::Invalid)`
+//!   - execution_domain: `Pending`
```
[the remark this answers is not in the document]
data-pup: Really good catch here. :slightly_smiling_face:

lucet-runtime/lucet-runtime-tests/src/timeout.rs
data-pup: Yep, you're exactly right :slightly_smiling_face:

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -23,61 +14,100 @@
...
+//!     executing a hostcall, but may also be `Terminated` while in a hostcall to indicate that it
+//!     should exit when the hostcall completes.
 //!   - `terminable` will be false if and only if `execution_domain` is `Terminated`.
+//!   - `KillSwitch::terminate` will succeed with `Signalled` when terminated while executing guest
+//!     code, `Pending` when terminated while executing a hostcall, and will fail with
+//!     `NotTerminable` when the instance has already been terminated.
 //! * `Instance::run returns`
-//!   - terminable: `false`
+//!   - terminable: `true`
 //!   - execution_domain: `Guest, Hostcall, or Terminated`
```
iximeow:
```suggestion
//!   - execution_domain: `Pending, Hostcall, or Terminated`
```

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -125,33 +155,131 @@ pub struct KillState {
...
         loop {}
+    } else {
+        let current_domain = (*instance).kill_state.execution_domain.lock().unwrap();
+        match *current_domain {
+            Domain::Guest => {
+                // We finished executing the code in our guest region normally! We should reset
+                // the kill state, invalidating any existing killswitches' weak references. We
+                // forget the mutex guard so that we don't invoke its destructor and segfault.
+                (*instance).kill_state = Arc::new(KillState::default());
+                mem::forget(current_domain);
```
iximeow: I think you mean `mem::drop` here? I'm pretty sure `forget`ing the lock guard would result in old `KillState` accumulating over time.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -533,10 +533,15 @@ impl Instance {
     /// modified by this call; it is the embedder's responsibility to clear or reset their state if
     /// necessary.
     ///
+    /// This will also reinitialize the kill state, which means that any outstanding
+    /// [`KillSwitch`](struct.KillSwitch.html) objects will be unable to terminate this instance.
+    /// It is the embedder's responsibility to initialize new killswitches after resetting an
```
iximeow:
```suggestion
    /// It is the embedder's responsibility to initialize new `KillSwitch`es after resetting an
```
just to be clear to the reader that we do mean this library item, `KillSwitch`

katie-martin-fastly:
> This PR has been a real journey. sunglasses

Yep!! Thanks for reviewing along the way :heart:

PR #432, opened 2020-02-20
Make clippy happy!

jedisct1:


lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -441,27 +441,25 @@ unsafe fn mmap_aligned(
...
     {
         let unused_back = (unaligned + (padded_length - 1)) - (aligned + (requested_length - 1));
-        if unused_back != 0 {
-            if munmap((aligned + requested_length) as *mut c_void, unused_back).is_err() {
-                // explicitly ignore failures now, as this is just a best-effort clean up after the last fail
-                let _ = munmap(unaligned as *mut c_void, padded_length);
-                return Err(Error::Unsupported("Could not align memory".to_string()));
-            }
+        if unused_back != 0
+            && munmap((aligned + requested_length) as *mut c_void, unused_back).is_err()
```
data-pup: I love that clippy catches things like this :slightly_smiling_face:

lucet-validate/tests/wasitests.rs
```diff
@@ -57,7 +56,7 @@ mod lucet_validate_tests {
             };
             validator
                 .validate(&entry_wasm)
-                .expect(&format!("validate {:?}", entry_path));
+                .unwrap_or_else(|_| panic!("validate {:?}", entry_path));
```
awortman-fastly: I didn't realize `expect` doesn't take a format-like series of arguments. huh! This undoubtedly would have been formatting the error string even on non-errors so :+1:

lucetc/src/decls.rs
```diff
@@ -383,7 +383,7 @@ impl<'a> ModuleDecls<'a> {
             }
             _ => Err(Error::Unsupported(
                 "lucetc only supports memory 0".to_string(),
-            ))?,
```
awortman-fastly: I'm also surprised to learn that `Err(..)?` where a returned expression would be isn't an error! but I guess rustc figures `Err(..)?` won't continue, so for the purposes of inference the entire thing gets a type of `!` or something.

PR #437, opened 2020-02-29
Multiplatform signals

iximeow:
This _mostly_ gets things in place for the Lucet signal handler to work properly on non-linux posix-y platforms, most interestingly: macos and flavors of bsd.

The gist of this change is to not context swap in the signal handler itself, but replace the address the guest continues with by the context swap function. So the signal handler can return on all platforms, passing through `sigreturn`-like mechanisms if present, and we have fewer cross-platform compatibility issues.


iximeow:
the remaining unhandled piece of work here is turning on macos CI, which is, I think, straightforward? I just haven't looked at it yet.

lucet-runtime/lucet-runtime-internals/src/context/mod.rs
```diff
@@ -628,10 +628,12 @@ impl Context {
     ///
     /// TODO: the return type of this function should really be `Result<!, nix::Error>`, but using
     /// `!` as a type like that is currently experimental.
-    #[inline]
```
iximeow: the `inline` goes away because at first I thought "to do this it needs to not be inlined into the signal handler!!" but then scrapped the first very bad idea for only using a function pointer to `set_from_signal`. And if it's only used as a function pointer in the code, there's nowhere to inline, so even wanting to inline it would be misleading to the reader.

lucet-runtime/lucet-runtime-internals/src/sysdeps/macos.rs
```diff
@@ -118,54 +118,55 @@ pub struct ucontext_t {
...
-    pub fn new(ptr: *const c_void) -> Self {
+    pub fn new(ptr: *mut c_void) -> Self {
         assert!(!ptr.is_null(), "non-null context");
-        UContextPtr(ptr as *const ucontext_t)
+        UContextPtr(ptr as *mut ucontext_t)
     }
 
     #[inline]
     pub fn get_ip(self) -> *const c_void {
-        let mcontext = &unsafe { *(*self.0).uc_mcontext };
```
iximeow: this did a deref when we'd wanted a pointer coercion :grimacing:

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
```diff
@@ -435,6 +435,7 @@ macro_rules! guest_fault_tests {
                     SigSet::all(),
                 );
                 unsafe { sigaction(Signal::SIGSEGV, &sa).expect("sigaction succeeds") };
+                unsafe { sigaction(Signal::SIGBUS, &sa).expect("sigaction succeeds") };
```
iximeow: so this is why the test's `SIGBUS` handler never appeared to be restored :grimacing:

`guest_fault.rs` becomes even more ominous on macos.

lucet-runtime/lucet-runtime-internals/src/sysdeps/macos.rs
```diff
@@ -118,54 +118,55 @@ pub struct ucontext_t {
...
+    pub fn new(ptr: *mut c_void) -> Self {
+        let context = unsafe { (ptr as *mut ucontext_t).as_mut().expect("non-null context") };
+        UContext { context }
     }
 
     pub fn as_ptr(&mut self) -> UContextPtr {
-        self.context.uc_mcontext = &self.mcontext;
-        UContextPtr::new(&self.context as *const _ as *const _)
+        // TODO: load-bearing?
+        // self.context.uc_mcontext = &self.mcontext;
```
iximeow: I don't know why we ever saved `mcontext` independently of `context`/`ucontext_t`. @jedisct1 would you happen to know these bits?

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -255,11 +254,7 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
     });
 
     if switch_to_host {
-        HOST_CTX.with(|host_ctx| unsafe {
-            Context::set_from_signal(&*host_ctx.get())
-                .expect("can successfully switch back to the host context");
-        });
-        unreachable!()
+        ctx.set_ip(Context::set_from_signal as *const c_void);
```
data-pup: This comes from `lucet-runtime/lucet-runtime-internals/src/sysdeps/linux.rs` I believe? As a reader, it might be nice to have a comment explaining why this is? (i.e., "_we can use this on other platforms_", or "_this function is linux specific to begin with_"?)

lucet-runtime/lucet-runtime-internals/src/context/mod.rs
data-pup: Thanks for explaining this, makes sense to me. Nice catch!

lucet-runtime/lucet-runtime-internals/src/sysdeps/macos.rs
jedisct1: On Apple devices, `mcontext` is a pointer, not the actual data.

jedisct1: ... and it's not guaranteed that the pointer leads to something that's part of the `ucontext` structure. The structure size doesn't even account for that data unless `_XOPEN_SOURCE` is defined.
But even then, if my memory serves me correctly, MacOS didn't store the data after the pointer.

iximeow: Oh I see, because this struct copied the whole `ucontext_t`, we have to fix up the `mcontext` pointer before handing out a pointer to `self`. That makes sense, and I'll delete this comment/line since `struct UContext` is just a wrapper around the `ucontext_t*` from the signal handler now.

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
katie-martin-fastly: We talked about this one directly, this was a misunderstanding on my part. Resolved :+1:

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
acfoltzer: Aw come on. Good catch

lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
```diff
@@ -468,6 +492,8 @@ impl Limits {
             self.globals_size,
             host_page_size(),
             self.signal_stack_size,
+            host_page_size(),
+            Slot::emergency_stack_size(),
```
acfoltzer: Can you update the comment above?

lucet-runtime/lucet-runtime-internals/src/context/mod.rs
```diff
@@ -628,10 +628,12 @@ impl Context {
...
-    pub unsafe fn set_from_signal(to: &Context) -> Result<(), nix::Error> {
-        signal::pthread_sigmask(signal::SigmaskHow::SIG_SETMASK, Some(&to.sigset), None)?;
-        Context::set(to)
+    pub unsafe extern "C" fn set_from_signal() -> Result<(), nix::Error> {
```
acfoltzer: Can we make an `Instance`-specific version of this, and leave the original, as this does something rather different? This is mostly to satisfy my own itch to keep `context` and its submodules cleanly separated from the rest of the codebase.

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -218,8 +217,12 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
                     // If the trap table lookup returned unknown, it is a fatal error
                     let unknown_fault = trapcode.is_none();
 
-                    // If the trap was a segv or bus fault and the addressed memory was in the
-                    // signal stack guard page or outside the alloc entirely, the fault is fatal
+                    // The stack of where we want to resume, a Lucet guest, may have been
+                    // exhausted. So, swap to an emergency stack to ensure we'll have at least a
+                    // few words of stack space as may be necessary for `Context::set_from_signal`
+                    // in debug builds.
+                    ctx.set_sp(inst.alloc.slot().emergency_stack_base());
```
acfoltzer: Will we ever end up at this point with something already on the emergency stack? My thoughts is "no" because any signals that arrive while doing the emergency work will be native code, and so we'd want to bring down the process anyway. But I want to make sure, and note the reasoning in the code so folks aren't worried about clobbering.

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -255,11 +258,7 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
     });
 
     if switch_to_host {
-        HOST_CTX.with(|host_ctx| unsafe {
-            Context::set_from_signal(&*host_ctx.get())
-                .expect("can successfully switch back to the host context");
-        });
-        unreachable!()
+        ctx.set_ip(Context::set_from_signal as *const c_void);
```
acfoltzer: Could you add a little prose comment for our future selves about what's being done here, and why we're not calling `set_from_signal` directly?

lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
```diff
@@ -166,6 +183,13 @@ impl Alloc {
             return AddrLocation::Stack;
         }
 
+        let stack_emergency_start = self.slot().emergency_stack as usize;
+        let stack_emergency_end = stack_emergency_start + Slot::emergency_stack_size();
+
+        if (addr >= stack_emergency_start) && (addr < stack_emergency_end) {
+            return AddrLocation::StackEmergencyGuard;
+        }
+
```
acfoltzer: Can you move this to the end of the function, so that it appears in the same order that it does in the laid out memory?

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -376,16 +380,35 @@ impl MmapRegion {
...
         let globals = stack + region.limits.stack_size;
         let sigstack = globals + region.limits.globals_size + host_page_size();
+        let stack_emergency_guard = sigstack + region.limits.signal_stack_size;
+        let stack_emergency_space = stack_emergency_guard + host_page_size();
+
+        // ensure we've accounted for all space
+        assert_eq!(
+            stack_emergency_space + Slot::emergency_stack_size() - mem as usize,
+            region.limits.total_memory_size()
+        );
```
acfoltzer: :heart_eyes_cat:

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -376,16 +380,35 @@ impl MmapRegion {
...
+
+        // set the emergency stack to read/write, so it's ready in advance in case we need it.
```
acfoltzer: Is there a scenario where this would be needed before the sigaltstack is unprotected? If not I'd rather treat them uniformly.

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
```diff
@@ -305,8 +305,8 @@ macro_rules! guest_fault_tests {
                 _siginfo_ptr: *const siginfo_t,
                 _ucontext_ptr: *const c_void,
             ) -> SignalBehavior {
-                // Triggered by a SIGSEGV writing to protected page
-                assert!(signum == SIGSEGV);
+                // Triggered by a SIGSEGV writing to protected page, or SIGBUS on macos
+                assert!(signum == SIGSEGV || signum == SIGBUS);
```
acfoltzer: Could we make this more precise with something like:
```rust
#[cfg(not(target_os="macos"))]
const PROTECTED_PAGE_SIG: libc::c_int = SIGSEGV;
#[cfg(target_os="macos")]
const PROTECTED_PAGE_SIG: libc::c_int = SIGBUS;
[...]
assert!(signum == PROTECTED_PAGE_SIG);
```

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
acfoltzer: And likewise, let's try to only modify the handler that is relevant for the current OS.

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
iximeow: While all the emergency stack stuff is gone now, I kept this assert around just to be safe in the future.

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
iximeow: In the test or in `lucet-runtime`? Since in the runtime we're unconditionally replacing `SIGBUS` already, I think you mean the test.

acfoltzer: Yes, sorry. My context for this was recommending platform-specific constants earlier in the test file, rather than checking for both.

awortman-fastly:
:tada: All checks have passed :tada:

This branch has picked up a few non-signal-handling-related changes now, so given a :+1: I'll fix up the commit history here into the ~three commits I'd like to land ("change signal handling around", "fix C API signal stack sizes to the largest minimum we know we have to work with -- 32kb on macos:, and "enable mac CI")

iximeow:
:tada: All checks have passed :tada:

This branch has picked up a few non-signal-handling-related changes now, so given a +1 I'll fix up the commit history here into the ~three commits I'd like to land ("change signal handling around", "fix C API signal stack sizes to the largest minimum we know we have to work with -- 32kb on macos", and "enable mac CI")

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
iximeow: Done - there are instances we need the `libc::c_int`, and instances we need the `nix::Signal`, so this turned into two `const`s

lucet-runtime/lucet-runtime-internals/src/instance/signals.rs
```diff
@@ -333,11 +332,23 @@ extern "C" fn handle_signal(signum: c_int, siginfo_ptr: *mut siginfo_t, ucontext
...
-                .expect("can successfully switch back to the host context");
-        });
-        unreachable!()
+        // Switch to host by preparing the context to switch when we return from the signal andler.
+        // We must return from the signal handler for POSIX reasons, so instead prepare the context
+        // that the signal handler will resume the program as if a call were made. First, by
+        // pointing the instruction pointer at `lucet_context_set`, then by preparing the argument
+        // that `lucet_context_set` should read from `rdi` - the context to switch to.
+        //
+        // NOTE: it is absolutely critical that `lucet_context_set` does not use the guest stack!
```
iximeow: full disclosure: I left most of the branch unchanged while adjusting the commits, but this `NOTE:` is new. I'd realized that there wasn't anything discussing the specific constraints on the function we continue into after handling a signal, and this addresses that.

iximeow:
Promise I'm done fiddling with this branch now, it's ready for a final +1! :crossed_fingers:

PR #438, opened 2020-03-02
lucetc: report all unresolved bindings, not just the first one

jedisct1:


fst-crenshaw:
Thanks for this quick work to improve the error output that lucet provides!  It will help users to know the full list of unresolved bindings.  

One of the things that bit me a little bit over in https://github.com/bytecodealliance/lucet/pull/374 was the lack of tests on error output.  I ended up squashing too much error output.  😿 

With that in mind, I wonder if you might have plans to add a test for this work?

jedisct1:
Is `anyhow` guaranteed to always print errors the same way, byte for byte?

lucet-module/src/error.rs
```diff
@@ -31,4 +32,6 @@ pub enum Error {
     UnknownModule { module: String, symbol: String },
     #[error("Unknown symbol `{module}::{symbol}`")]
     UnknownSymbol { module: String, symbol: String },
+    #[error("Unknown bindings {unknown_bindings}")]
+    UnknownBindings { unknown_bindings: UnknownBindings },
```
acfoltzer: If you add a `#[from]` here, decls.rs:201 could be a bit simpler

jedisct1: Done, thanks!

pchickey:
@jedisct1 for testing I'd reuse the existing `import_many.wat` test and match on the `lucetc::Error`, rather than coerce it to an anyhow::Error.

pchickey:

PR #440, opened 2020-03-04
Have curl follow redirects in compiling on linux docs

torch2424:
Was following these docs, and the URL redirects quite a bit (as the project moved under the Wasm org, and then points to an asset). Thus, Curl needs to follow these redirects to get the final asset :smile:

### No Redirects

Screenshot from 2020-03-04 11-39-02

### Redirects (this change)

Screenshot from 2020-03-04 11-39-24


torch2424:
@iximeow Made offline requested changes :smile: :+1:

jedisct1:
Thanks!

PR #447, opened 2020-03-10
update cranelift to bytecodealliance/wasmtime cranelift

froydnj:
This stack of commits (which is not bisectable in any way!) updates lucet to use cranelift from the wasmtime repository.  It does not implement @pchickey's suggestion of making everything use released versions of cranelift; I can do that here or in a separate PR.

The commits around `ModuleEnvironment` and `FunctionEnvironment` take a "make it build" mentality as regards the new methods on said traits.


froydnj:
I'm also not 100% sure about the changes to make the TLS bits build properly.  I don't think passing `false` for TLS-ness everywhere is necessarily correct, but it's also not obvious to me that things would deal correctly with TLS inside the rest of Lucet anyway.

pchickey:
Switching the submodule rather than switching to the crates.io versions is fine, I can do that in a follow-up PR. This is the much more important work of upgrading all the places where interfaces have changed.

It looks like GH is having trouble checking out the PR because its expecting cranelift to still have a url in .gitmodules. Submodules are always a pain when making big changes like this.

froydnj:
> It looks like GH is having trouble checking out the PR because its expecting cranelift to still have a url in .gitmodules. Submodules are always a pain when making big changes like this.

Clearly I am missing something, because I removed the cranelift definition from `.gitmodules`!  I have pushed a new branch which keeps it there, and tests seem to be somewhat happier (though e.g. packaging probably won't be).  Does the submodule need to be kept around and then removed in a separate PR when it's no longer needed?

pchickey:
It should be possible to delete in this PR, but it may need its own commit. I believe that `git rm <submodulepath>` will do the right thing in newer versions of git.

froydnj:
The `start` failures seem like the easiest thing to start tackling.  The `global_init` test winds up with a bad pointer for `guest_start`; readelf on the binary says:

```
    24: 000000000020085e     0 OBJECT  GLOBAL DEFAULT    9 guest_start
```

whereas the pre-PR binary says:

```
    24: 000000000000058d     8 OBJECT  GLOBAL DEFAULT    6 guest_start
```

Post-PR, the symbol's address lives in `.data`, whereas prior the symbol lives in `.rodata`.  I suspect that there's some kind of bug in faerie...but AFAICS, there's basically no functional change between faerie 0.12 and 0.14.

froydnj:
Ah, maybe this is due to bytecodeallianc/cranelift#1209, which changes `cranelift-faerie` module.

froydnj:
#448 resolves some of the crashiness seen with the `lucet-runtime` tests.

froydnj:
Hm, the build/test was green yesterday, but rebasing to fix conflicts results in:

```
    Compiling lucet-wiggle v0.6.2-dev (/github/workspace/lucet-wiggle)
   Compiling lucet-wasi-fuzz v0.6.2-dev (/github/workspace/lucet-wasi-fuzz)
error: proc macro panicked
  --> lucet-wiggle/tests/wasi.rs:51:1
   |
51 | / lucet_wiggle::from_witx!({
52 | |     witx: ["../wasi/phases/snapshot/witx/wasi_snapshot_preview1.witx"],
53 | |     ctx: LucetWasiCtx,
54 | |     constructor: { LucetWasiCtx::build(vmctx) },
55 | | });
   | |___^
   |
   = help: message: lucet loading witx: Parse(Error { inner: ErrorInner { text: Some(Text { line: 522, col: 4, snippet: "    (field $fd_readwrite $event_fd_readwrite)" }), file: Some("/github/workspace/wasi/phases/snapshot/witx/typenames.witx"), span: Span { offset: 15556 }, kind: Custom("expected an identifier") } })

error: aborting due to previous error
```

which seems...bad.  I must have mixed up some crate versions when rebasing or something.

pchickey:
That would happen if the wasi submodule was out-of-date - it didnt happen in CI, so I think you just need to run `git submodule update` locally.

froydnj:
Yeah, I had somehow bumped the wasi submodule some way, and that was causing bad things to happen.  Everything's back to green now.

lucetc/src/error.rs
```diff
@@ -32,7 +32,7 @@ pub enum Error {
     #[error("Artifact: {1}. {0:?}")]
     ArtifactError(ArtifactError, String),
     #[error("Failure: {1}. {0:?}")]
-    Failure(failure::Error, String),
+    Failure(anyhow::Error, String),
```
pchickey: not an issue with this PR, but we should take another pass at this enum with the latest changes that got rid of `failure`. This is the correct fix for now.

pchickey:
Just one more note - this depends on changes in the wasmtime submodule that are not present in the crates.io version, so we need to wait until theres a new cranelift release to delete the submodule

PR #448, opened 2020-03-11
lucetc: don't use define_zeroinit for guest_start data

froydnj:
After bytecodealliance/cranelift#1209, the semantics of
`DataContext::define_zeroinit` appear to only be for truly
zero-initialized data that won't have any relocations applied to it.
Since `guest_start` is clearly not that kind of data, move to using
`DataContext::define` instead.

### DIFF
--- a/lucetc/src/compiler.rs
+++ b/lucetc/src/compiler.rs
@@ -291,7 +291,7 @@ fn write_startfunc_data<B: ClifBackend>(
             .declare_data("guest_start", Linkage::Export, false, None)
             .map_err(Error::MetadataSerializer)?;
         let mut ctx = DataContext::new();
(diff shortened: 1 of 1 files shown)


pchickey:
@froydnj you should have a commit bit now, go ahead and merge your PRs once they go green

froydnj:
> @froydnj you should have a commit bit now, go ahead and merge your PRs once they go green

I do not have a commit bit for lucet; you might be thinking of my commit bit for wasmtime?

pchickey:
My bad, I jumped the gun, @tyler will add you to the lucet group when he gets a minute

tyler:
@froydnj you should have an invitation now :) Thanks!

PR #456, opened 2020-03-20
move to using cranelift-module to emit structures rather than faerie

froydnj:
This doesn't completely move to using `object`, but it seems like a good stopping point for discussion.

The only real ugliness is the `traps_to_module_traps`/`write_trap_table` division.  Ideally, these would be a single function, but the borrow checker says that the `ModuleCompiledFunction` that comes out of `Module::define_function` represents a borrow on the `Module` itself, so you can't have a single function that takes `&mut Module` and `&ModuleCompiledFunction`.  (`ModuleCompiledFunction` isn't actually exported, either, which seems like an oversight.) Ergo, we have to do a dance to satisfy the borrow checker.  Maybe it would have been a good idea to just make `Module::define_function` take a trap sink instead of trying to have `cranelift-module` backends manage the traps themselves, as suggested in @bytecodealliance/wasmtime#1184.

### DIFF
--- a/lucetc/src/compiler.rs
+++ b/lucetc/src/compiler.rs
@@ -6,26 +6,29 @@ use crate::error::Error;
 use crate::function::FuncInfo;
 use crate::heap::HeapSettings;
 use crate::module::ModuleInfo;
-use crate::output::{CraneliftFuncs, ObjectFile};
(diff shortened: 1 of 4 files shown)


froydnj:
> I agree that we should change cranelift-module to accept trap sinks rather than emit a TrapSite vector. Do you want to take that one or should I?

I will handle this.  Thanks for the review!

PR #458, opened 2020-03-24
Fixes #451: Introduce a custom heap size limit to new_instance_with()

### LINKED ISSUE #451: Add ability to set Limits per Instance
Currently a Region is created with a set of Limits. Those Limits are applied to the Region (in order to correctly lay out memory) as well as to the Instances in how they behave at runtime, how much memory they are allowed to allocate and so on.

We would like to be able to have Limits set per-Instance. 

That has some small complications. For instance, the Limits for the instance must generally be "smaller" than the Limits set on the Region in order for it to be safe and not turn into a crazy bin-packing problem. This will likely also require making *where* certain things are allocated for the instance within the Slot dynamic. Probably other little issues as well.

fst-crenshaw:
See https://github.com/bytecodealliance/lucet/issues/451 for description of the issue and proposed solution.

Still to do:
- [x] ~Replace `pub fn with_heap_size_limit(mut self, heap_memory_size: usize) -> Self` with trendier `pub fn with_heap_size_limit(&mut self, heap_memory_size: usize) -> &mut Self`.~. This isn't actually possible because it's not possible to Clone an InstanceBuilder.
- [x] ~I making a call to clone() that feels extraneous and would like to hear thoughts on a better way.~ Add the Copy trait to the Limits type.
- [x] Examine for the possibility of more tests.
- [x] Augment the Alloc, not the Slot Limits.


lucet-runtime/lucet-runtime-internals/src/region.rs
```diff
@@ -67,6 +68,7 @@ pub trait RegionInternal: Send + Sync {
     fn expand_heap(&self, slot: &Slot, start: u32, len: u32) -> Result<(), Error>;
 
     fn reset_heap(&self, alloc: &mut Alloc, module: &dyn Module) -> Result<(), Error>;
+    fn heap_memory_size_limit(&self) -> usize;
```
acfoltzer: I would recommend making this instead an accessor for the `Limits`, just for extra flexibility

lucet-runtime/lucet-runtime-internals/src/region.rs
```diff
@@ -98,9 +101,18 @@ impl<'a> InstanceBuilder<'a> {
             region,
             module,
             embed_ctx: CtxMap::default(),
+            heap_memory_size: region.heap_memory_size_limit(),
         }
     }
 
+    /// Add a custom limit for the heap memory size to the built instance.
+    /// This call is not necessary if the default heap memory size is adequate
```
acfoltzer:
```suggestion
    /// Add a custom limit for the heap memory size to the built instance.
    ///
    /// This call is not necessary if the default heap memory size is adequate
```

The blank line separates the short description from the rest in the rendered rustdoc

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -90,8 +92,23 @@ impl RegionInternal for MmapRegion {
             lucet_bail!("heap is not page-aligned; this is a bug");
         }
 
-        let limits = &slot.limits;
-        module.validate_runtime_spec(limits)?;
+        let mut limits = slot.limits.clone();
```
acfoltzer: It might be time to make `Limits` impl `Copy`

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -90,8 +92,23 @@ impl RegionInternal for MmapRegion {
...
+            return Err(Error::InvalidArgument(
+                "heap memory size requested for instance is larger than slot allows",
+            ));
+        }
+
+        // If a custom heap memory size is supplied, augment the limits
+        // with the value so that it may be validated.
+        if let Ordering::Less = heap_memory_size.cmp(&slot.limits.heap_memory_size) {
+            limits.heap_memory_size = heap_memory_size;
+        }
```
acfoltzer: I think this would be easier to follow as a single `match heap_memory_size.cmp(...) { ... }`, and would make us be more explicit about what happens in the case that they're equal

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -78,6 +79,7 @@ impl RegionInternal for MmapRegion {
         &self,
         module: Arc<dyn Module>,
         embed_ctx: CtxMap,
+        heap_memory_size: usize,
```
acfoltzer: In the context of a struct called `Limits`, it's clear that `heap_memory_size` is a limit. But just as a bare argument to this function, it's less clear. Maybe call this `heap_memory_size_limit` like you do with the builder method?

fst-crenshaw:
Thanks to @acfoltzer for the suggestions on how to expand testing for this PR!

I've begun work on creating a test similar to `expand_past_spec_max` but I realize I am feeling unsure about the interactions between:
- a Region, 
- a Mocked Module’s Heap Spec
- an Instance’s heap memory size limit
- making a call to `expand_heap`.

I have a test that does the following:
- Make a Region with default limits.
- Make a Module with a three-page heap spec.
- Make an Instance with a one-page heap memory size limit.
- Call `expand_heap()` on the instance.

I *think* the desired result is that the test ought to fail.  It currently does not because I've done no work on making enhancements to `expand_heap`.  But I also might be misunderstanding something.

Here's the test: https://github.com/bytecodealliance/lucet/pull/458/files#diff-b03e99218c26a9cc6576a1e98a226430R202

acfoltzer:
I think you've got it correctly. The heap spec is an extra complication, but you can think of yet another limit on heap size, this time specified at compile time rather than at runtime.

> * Make a Region with default limits.

I will point out that this is using a `LIMITS` constant in the test module, not `Limits::default()`. I don't think that should affect the correctness of your test, though.

fst-crenshaw:
I've carved out some more tests for this.  I believe I've addressed all the PR feedback, but would love to hear other approaches to improve.

lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
```diff
@@ -820,6 +860,82 @@ macro_rules! alloc_tests {
...
+
+            // Build a default-sized instance, to make sure the custom limits
+            // didn't break the defaults.
+            let default_inst = region
+                .new_instance(module.clone())
+                .expect("new_instance succeeds");
+
+            // Affirm that its heap is the expected size.
+            let heap_len = default_inst.alloc().heap_len();
+            assert_eq!(heap_len, THREEPAGE_INITIAL_SIZE as usize);
```
fst-crenshaw: Ugh.  I don't like what I did here.  I am going to revise this to use a different initial size in the HeapSpec.  Please hold.

lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
```diff
@@ -820,6 +860,88 @@ macro_rules! alloc_tests {
...
+            }
+        }
+
+        /// This test exercises custom limits on the heap_memory_size.
+        /// In this scenario, successfully create a custom-sized
+        /// instance followed by a default-sized instance to affirm
+        /// that a custom size doesn't somehow overwrite the default size.
+        #[test]
+        fn custom_size_does_not_break_default() {
+            let region = TestRegion::create(2, &LIMITS).expect("region created");
```
acfoltzer: (1/2)
```suggestion
            let region = TestRegion::create(1, &LIMITS).expect("region created");
```

lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
```diff
@@ -820,6 +860,88 @@ macro_rules! alloc_tests {
...
+                )
+                .with_heap_size_limit((THREE_PAGE_MAX_HEAP.initial_size * 2) as usize)
+                .build()
+                .expect("new instance succeeds");
+
+            // Affirm that its heap is the expected size, the size
+            // specified in the HeapSpec.
+            let heap_len = custom_inst.alloc().heap_len();
+            assert_eq!(heap_len, THREE_PAGE_MAX_HEAP.initial_size as usize);
+
```
acfoltzer: (2/2)
```suggestion
            drop(custom_inst);
```

fst-crenshaw: I get it.  If I alter the Slot, it's altered forever.  I ought to alter the Alloc.  Really nice catch.  Thank you so much.

fst-crenshaw:
It may not seem like it from the CI, but this is ready for another review when folks have the cycles.  Everything passed on 44bc0b0.  Since then, I corrected a spelling error and then retriggered the CI a few times.  It seems like the little mac mini farm might be having a bad time.

fst-crenshaw:
This is ready for another review.

fst-crenshaw:
One more time?  😄 😊 https://www.youtube.com/watch?v=FGBhQbmPwH8

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -114,6 +106,17 @@ impl RegionInternal for MmapRegion {
...
 
+        let slot = self
+            .freelist
+            .write()
+            .unwrap()
+            .pop()
+            .ok_or(Error::RegionFull(self.capacity))?;
+
+        if slot.heap as usize % host_page_size() != 0 {
+            lucet_bail!("heap is not page-aligned; this is a bug");
```
acfoltzer: Sorry I didn't catch this before, but this is also an early return. Since it would reflect a bug in Lucet, I think we should change it to a panic/assert rather than returning an `InternalError`.

fst-crenshaw: Thanks for catching this!

fst-crenshaw: Since the code you point at is in master, and since it feels a bit unrelated to my main task of customizing limits, I wonder if you might accept my fixing it in a PR over here:  https://github.com/bytecodealliance/lucet/pull/486.

acfoltzer: I left feedback there, but it is related to this PR because the early return at that location would cause the slot to leak.

acfoltzer: ... sorry, I got confused about which PR I was looking at. Not the one we already merged :blush:

fst-crenshaw: No worries.  I'm having trouble keeping the little changes straight as well.

PR #460, opened 2020-03-25
Fix lucetc reporting the number of tables incorrectly

iximeow:
A little while ago one of the changes to `lucetc` resulted in the field for "number of tables" instead holding "number of entries in the first table, or zero". This would be a bigger issue* if wasm ever used tables other than 0, but it doesn't for the time being, so the next likely place to run afoul of this is in `lucet-objdump` which crashes trying to read more tables than there really are, following stray pointers.

This change switches back to reporting the number of tables as the [length of the list of tables](https://github.com/bytecodealliance/lucet/blob/master/lucet-module/src/module.rs#L27), and includes a smoke-test through `lucet-objdump` to catch issues like this in the future.

* a longer description of why this isn't a severe issue for `lucet-runtime`: either there are zero tables, in which case this field will hold `0` anyway and no stray reads will happen, or there are >0 tables and `lucet-runtime` will care only about the first one, regardless of how many other tables are reported. If the number of entries is non-zero, there must be a table they exist in, and the first table in the list of tables will be valid. The concerning case I can imagine is where there is one table of zero entries, which would result in a `lucet-runtime` error [trying to get table entries](https://github.com/bytecodealliance/lucet/blob/master/lucet-runtime/lucet-runtime-internals/src/module/dl.rs#L241). So in the worst case `lucet-runtime` might fail with an error on a well-formed wasm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ test-except-fuzz: test-packages
 	# run the benchmarks in debug mode
 	cargo test --benches -p lucet-benchmarks -- --test
 	helpers/lucet-toolchain-tests/signature.sh
+	helpers/lucet-toolchain-tests/objdump.sh
 
 # run a single seed through the fuzzer to stave off bitrot
 .PHONY: test-fuzz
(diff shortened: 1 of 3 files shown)


pchickey:
I believe a rebase on master will fix the cargo audit error in CI.

iximeow:
Rebased, :crossed_fingers: that CI approves

PR #461, opened 2020-03-25
Update the minisign crate

jedisct1:
The version pinned in `lucet-module` didn't support modules larger than 1GB.


jedisct1:
CI surfaced a security advisory for the `bumpalo` crate version we pinned.

Not a big security risk since it's only used by `criterion`, but this is breaking CI, so I just added an update for that one, too.

fst-crenshaw:
@jedisct1 -- I don't see the criterion / bumpalo update on this PR.  Am I missing something?

jedisct1:
@fst-crenshaw https://github.com/bytecodealliance/lucet/pull/461/commits/e6da5624d43eb2f2cb3c7554b8f65efc37f0e13a -- You need to click `view diff`.

fst-crenshaw:
Thanks for the help.

PR #466, opened 2020-03-27
The lucet-module has an assumption about max_size that isn't checked when an instance is created.

fst-crenshaw:
As I'm introducing new kinds of `heap_memory_size` limits to an instance, I noticed a small inconsistency in assumption-checking that I'd like to discuss.  I realize that we may not necessarily merge this PR as I may be missing something regarding HeapSpec checks.  Always look forward to your comments to help me learn and be better.

A HeapSpec is defined here:  https://github.com/bytecodealliance/lucet/blob/master/lucet-module/src/linear_memory.rs#L37-L76

This assumption on a heap's `max_size` is recorded in the comments:

> The program may optionally define this value. If it does, it must be less than the `reserved_size`.

Furthermore, this assumption on a heap's `initial_size` is also recorded in the comments:

> [it] must be less than or equal to `reserved_size`.

I noticed that when an instance is created and its heap spec is validated, the latter assumptions is checked but the former is not.  This PR -- out of my healthy abundance of paranoia -- introduces a check on a heap's `max_size` to `validate_runtime_spec`.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
@@ -718,17 +718,37 @@ macro_rules! alloc_tests {
             assert_eq!(region.used_slots(), 0);
         }
 
-        fn do_nothing_module() -> Arc<dyn Module> {
-            extern "C" fn do_nothing(_vmctx: *mut lucet_vmctx) -> () {}
+        const REJECTMAX_INITIAL_SIZE: u64 = LIMITS_HEAP_MEM_SIZE as u64;
+        const REJECTMAX_MAX_SIZE: u64 = SPEC_HEAP_RESERVED_SIZE as u64 + (64 * 1024);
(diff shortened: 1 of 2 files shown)


pchickey:
I don't believe we want this behavior. The maximum linear memory size in the wasm module is not required to be satisfiable by the runtime - the memory grow opcode can fail before the maximum is reached. It is only required that the memory grow to no larger than the maximum (https://webassembly.github.io/spec/core/exec/runtime.html#syntax-meminst).

fst-crenshaw:
Perhaps, then, do you think it would be a better PR to remove the comment in the `lucet-module` code? 

> The program may optionally define this value. If it does, it must be less than the reserved_size.

This maximum value is being checked with the heap is expanded, but not at instance creation time.

pchickey:
Ah, yes, the comment you have quoted is incorrect. I do believe that the comments at the use site you linked to are correct.

fst-crenshaw:
Thanks so much for weighing in @pchickey .  I'll go ahead and close this PR.

PR #470, opened 2020-03-30
convert lucetc to use object as the I/O backend

froydnj:
This commit makes `lucetc` use `cranelift-object`/`object` as its interface for writing out native objects.  The intent is that this moves `lucetc` closer to getting PE/COFF support "for free".

We had discussed providing a runtime switch for compatibility/risk reasons, but given that so much is going through `cranelift-module` now and the amount of `object`-specific code is minimal, I believe that providing a runtime switch is just extra complexity for no reason.  Happy to hear pushback on that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,7 +23,7 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.28"
+version = "1.0.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -31,7 +31,7 @@ name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "hermit-abi 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hermit-abi 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.68 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -90,7 +90,7 @@ version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "byteorder 1.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.106 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.105 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -107,7 +107,7 @@ dependencies = [
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "peeking_take_while 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "proc-macro2 1.0.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 1.0.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "quote 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-hash 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -165,7 +165,7 @@ dependencies = [
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "memchr 2.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex-automata 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.106 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.105 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -299,7 +299,7 @@ dependencies = [
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "smallvec 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "target-lexicon 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "thiserror 1.0.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "thiserror 1.0.13 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -319,18 +319,6 @@ name = "cranelift-entity"
 version = "0.62.0"
 
 [[package]]
-name = "cranelift-faerie"
-version = "0.62.0"
-dependencies = [
- "anyhow 1.0.28 (registry+https://github.com/rust-lang/crates.io-index)",
- "cranelift-codegen 0.62.0",
- "cranelift-module 0.62.0",
- "faerie 0.15.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "goblin 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "target-lexicon 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
 name = "cranelift-frontend"
 version = "0.62.0"
 dependencies = [
@@ -344,11 +332,11 @@ dependencies = [
 name = "cranelift-module"
 version = "0.62.0"
 dependencies = [
- "anyhow 1.0.28 (registry+https://github.com/rust-lang/crates.io-index)",
+ "anyhow 1.0.27 (registry+https://github.com/rust-lang/crates.io-index)",
  "cranelift-codegen 0.62.0",
  "cranelift-entity 0.62.0",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "thiserror 1.0.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "thiserror 1.0.13 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -361,6 +349,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cranelift-object"
+version = "0.62.0"
+dependencies = [
+ "cranelift-codegen 0.62.0",
+ "cranelift-module 0.62.0",
+ "object 0.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "target-lexicon 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "cranelift-wasm"
 version = "0.62.0"
 dependencies = [
@@ -368,7 +366,7 @@ dependencies = [
  "cranelift-entity 0.62.0",
  "cranelift-frontend 0.62.0",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "thiserror 1.0.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "thiserror 1.0.13 (registry+https://github.com/rust-lang/crates.io-index)",
  "wasmparser 0.51.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -397,9 +395,9 @@ dependencies = [
  "plotters 0.2.12 (registry+https://github.com/rust-lang/crates.io-index)",
  "rayon 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.106 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.106 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_json 1.0.50 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.105 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.105 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.48 (registry+https://github.com/rust-lang/crates.io-index)",
  "tinytemplate 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "walkdir 2.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -474,7 +472,7 @@ dependencies = [
  "csv-core 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "itoa 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "ryu 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.106 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.105 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -530,7 +528,7 @@ dependencies = [
 
 [[package]]
 name = "errno"
-version = "0.2.5"
+version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "errno-dragonfly 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


pchickey:
I agree this is now a tiny enough patch that we can commit to it, or revert if it ends up being an issue. Thank you for all of the excellent work on this! You wrangled all the artificial complexity of the old code into something much better, and mitigated a ton of risk in the process.

iximeow:
It looks like this gets `lucet-objdump` upset for some reason, realizing now that specific test failure isn't too clear in CI.

froydnj:
```
froydnj@hawkeye:/opt/extra/froydnj/lucet$ target/debug/lucet-objdump foo.o 
Required Symbols:
  lucet_module                  : exists
Native module components:
  module_data_ptr               : exists; address: 0x200808
  module_data_len               : 401
  tables_ptr                    : exists; address: 0x6c1
  tables_len                    : 4
  function_manifest_ptr         : exists; address: 0x5a1
  function_manifest_len         : 7
thread 'main' panicked at 'attempt to multiply with overflow', lucet-objdump/src/main.rs:511:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

It looks like the `tables_len` is different between `object`- and `faerie`-using compilers.  Will have to investigate.

awortman-fastly:
Interesting.. `tables_len` almost certainly should be `1` - it was `4` before https://github.com/bytecodealliance/lucet/pull/460 because that length was actually reporting the number of items in the first table. Since the `lucet-objdump` test exists here, you should have the fix in that branch as well.

froydnj:
> Interesting.. `tables_len` almost certainly should be `1` - it was `4` before #460 because that length was actually reporting the number of items in the first table. Since the `lucet-objdump` test exists here, you should have the fix in that branch as well.

Ah, thank you for pointing that out.  I think my `lucet-objdump` was not rebuilt correctly since pulling changes, so I was running with an old version.  A newer version indeed reports `tables_len` for both, and errors in a different way:

```
froydnj@hawkeye:/opt/extra/froydnj/lucet$ RUST_BACKTRACE=full target/release/lucet-objdump object-foo.o 
Required Symbols:
  lucet_module                  : exists
Native module components:
  module_data_ptr               : exists; address: 0x200910
  module_data_len               : 401
  tables_ptr                    : MISSING!
  tables_len                    : 1
  function_manifest_ptr         : MISSING!
  function_manifest_len         : 7
```

But you can see that the symbols are there:

```
froydnj@hawkeye:/opt/extra/froydnj/lucet$ readelf -sW faerie-foo.o |grep lucet_module
     8: 0000000000000831    64 OBJECT  GLOBAL DEFAULT    6 lucet_module
    15: 0000000000200968   401 OBJECT  LOCAL  DEFAULT    9 lucet_module_data
    25: 0000000000000831    64 OBJECT  GLOBAL DEFAULT    6 lucet_module
froydnj@hawkeye:/opt/extra/froydnj/lucet$ readelf -sW object-foo.o |grep lucet_module
     8: 00000000002007f0    64 OBJECT  GLOBAL DEFAULT    8 lucet_module
    16: 0000000000200910   401 OBJECT  LOCAL  DEFAULT   10 lucet_module_data
    26: 00000000002007f0    64 OBJECT  GLOBAL DEFAULT    8 lucet_module
```

The only wrinkle is that in the `object` version, `lucet_module` lives in `.data.rel.ro`.  The number of relocations is the same, 19 in both cases:

```
froydnj@hawkeye:/opt/extra/froydnj/lucet$ readelf -rW faerie-foo.o

Relocation section '.rela.dyn' at offset 0x438 contains 19 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
00000000000006b1  0000000000000008 R_X86_64_RELATIVE                         600
00000000000006c1  0000000000000008 R_X86_64_RELATIVE                         610
00000000000006d1  0000000000000008 R_X86_64_RELATIVE                         620
00000000000006f9  0000000000000008 R_X86_64_RELATIVE                         699
0000000000000751  0000000000000008 R_X86_64_RELATIVE                         660
0000000000000761  0000000000000008 R_X86_64_RELATIVE                         711
0000000000000771  0000000000000008 R_X86_64_RELATIVE                         600
0000000000000781  0000000000000008 R_X86_64_RELATIVE                         731
0000000000000791  0000000000000008 R_X86_64_RELATIVE                         610
00000000000007a1  0000000000000008 R_X86_64_RELATIVE                         709
00000000000007b1  0000000000000008 R_X86_64_RELATIVE                         620
00000000000007c1  0000000000000008 R_X86_64_RELATIVE                         739
0000000000000811  0000000000000008 R_X86_64_RELATIVE                         630
0000000000000821  0000000000000008 R_X86_64_RELATIVE                         741
0000000000000841  0000000000000008 R_X86_64_RELATIVE                         200968
0000000000000851  0000000000000008 R_X86_64_RELATIVE                         6f9
0000000000000861  0000000000000008 R_X86_64_RELATIVE                         751
00000000000007d1  0000000500000001 R_X86_64_64            0000000000000000 lucet_vmctx_current_memory + 0
00000000000007f1  0000000200000001 R_X86_64_64            0000000000000000 lucet_vmctx_grow_memory + 0

froydnj@hawkeye:/opt/extra/froydnj/lucet$ readelf -rW object-foo.o

Relocation section '.rela.dyn' at offset 0x400 contains 19 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
00000000002006b8  0000000000000008 R_X86_64_RELATIVE                         601
00000000002006c8  0000000000000008 R_X86_64_RELATIVE                         60c
00000000002006d8  0000000000000008 R_X86_64_RELATIVE                         617
0000000000200710  0000000000000008 R_X86_64_RELATIVE                         5c8
0000000000200730  0000000000000008 R_X86_64_RELATIVE                         601
0000000000200750  0000000000000008 R_X86_64_RELATIVE                         60c
0000000000200770  0000000000000008 R_X86_64_RELATIVE                         617
00000000002007d0  0000000000000008 R_X86_64_RELATIVE                         626
0000000000200800  0000000000000008 R_X86_64_RELATIVE                         200910
0000000000200700  0000001000000001 R_X86_64_64            00000000002006a0 guest_table_0 + 0
0000000000200720  0000000c00000001 R_X86_64_64            0000000000000651 lucet_trap_table_guest_func_foo + 0
0000000000200740  0000000700000001 R_X86_64_64            0000000000000671 lucet_trap_table_guest_func_1 + 0
0000000000200760  0000000600000001 R_X86_64_64            0000000000000679 lucet_trap_table_guest_func_2 + 0
0000000000200780  0000000300000001 R_X86_64_64            0000000000000681 lucet_trap_table_guest_func_3 + 0
0000000000200790  0000000500000001 R_X86_64_64            0000000000000000 lucet_vmctx_current_memory + 0
00000000002007b0  0000000200000001 R_X86_64_64            0000000000000000 lucet_vmctx_grow_memory + 0
00000000002007e0  0000000900000001 R_X86_64_64            0000000000000689 lucet_trap_table_lucet_probestack + 0
0000000000200810  0000000a00000001 R_X86_64_64            0000000000200700 lucet_tables + 0
0000000000200820  0000000400000001 R_X86_64_64            0000000000200710 lucet_function_manifest + 0
```

I think what's happening is that `faerie` is using relative address relocations everywhere, and the offset is stored directly in the actual location being relocated (I think) -- and `lucet-objdump` depends on that.  But `object` is smarter, and uses relocations against actual symbols and stores a plain old 0 in the actual location being relocated, which `lucet-objdump` is not prepared to cope with.

I don't know that adding a bunch of relocation processing to `lucet-objdump` is a happy path to go down, especially if you ever wanted to run `lucet-objdump` on multiple OSes or architectures.  `object` at least looks like it tries to abstract over relocations, so maybe it wouldn't be tooooo messy...

awortman-fastly:
Aha, that'll do it! Your assessment of relocation changes makes sense to me. I don't think we have reason to get too fancy with relocations in the future, so just adjusting `lucet-objdump` for the relocation changes in `faerie`->`object` seems fine for now. `lucet-objdump` is pretty opinionated about using `goblin::elf` for module parsing in the first place, so a later `TODO` about playing nice on other architectures should be okay.

froydnj:
> `lucet-objdump` is pretty opinionated about using `goblin::elf` for module parsing in the first place

Is there a reason to use `goblin` rather than something like `object::read` (primarily thinking of https://docs.rs/object/0.18.0/object/read/struct.Relocation.html for some semblance of extensibility)?

awortman-fastly:
Consensus seems to be: no reason, just that `object` didn't exist when this stuff was first written :)

froydnj:
> Consensus seems to be: no reason, just that `object` didn't exist when this stuff was first written :)

Hah!  OK, great; I will see about updating `llvm-objdump` as a separate PR, then.

froydnj:
Two notes about the conversion:

1. `faerie` sets `DT_TEXTREL` in the dynamic section, which means that pages in the read-only segment will not be shareable given that relocations are being applied to them.  `object` does not set this, thanks to handling `.data.rel.ro` correctly, so maybe there will be a little bit of memory savings.
2. `faerie` automatically adds a `PT_GNU_STACK` segment indicating non-executability of the stack; `object` does not, so we'll have to figure out how to tack that on for `object`.  It's not obvious to me that `object`'s file writing API makes this easy, so this will probably involve some modifications to `object`.  I don't think this patch wants to land until `PT_GNU_STACK` is handled correctly.

awortman-fastly:
> `faerie` automatically adds a `PT_GNU_STACK` segment indicating non-executability of the stack

We shouldn't have to worry about this here, since at least for dynamically loaded Lucet modules we'll manage stacks explicitly in `lucet-runtime`, where they are very not executable. I don't know how this pans out if someone tries statically linking a Lucet module in an object without `PT_GNU_STACK` with objects that do have it - would that disable no-execute stacks for the whole resulting object?

froydnj:
> > `faerie` automatically adds a `PT_GNU_STACK` segment indicating non-executability of the stack
> 
> We shouldn't have to worry about this here, since at least for dynamically loaded Lucet modules we'll manage stacks explicitly in `lucet-runtime`, where they are very not executable. I don't know how this pans out if someone tries statically linking a Lucet module in an object without `PT_GNU_STACK` with objects that do have it - would that disable no-execute stacks for the whole resulting object?

`PT_GNU_STACK` is automatically added by the linker based on the presence of `.note.GNU-stack` sections in all the object files; a missing `.note.GNU-stack` section in a single object file in the entire link means the resulting linked object must (being conservative) have an executable stack.  (At least, that was the intent; looking at `lld` sources, it looks like it ignores `.note.GNU-stack` sections entirely and stack executability in the resulting object is controlled by a command-line switch.)

I guess I misspoke, then: `faerie` automatically adds the relevant `.note.GNU-stack` section (https://github.com/m4b/faerie/blob/e91f4f33858f56f55744ec69e81096e2f213337e/src/elf.rs#L1013-L1020) whereas `object` does not.  Adding that section would be reasonable enough to do, although maybe `cranelift-module` should just do that automatically?

philipc:
> I think what's happening is that faerie is using relative address relocations everywhere, and the offset is stored directly in the actual location being relocated (I think) -- and lucet-objdump depends on that. But object is smarter, and uses relocations against actual symbols and stores a plain old 0 in the actual location being relocated, which lucet-objdump is not prepared to cope with.

I haven't looked in detail, but I don't think faerie stores the offset in the location being relocated (.rela  relocations never do that). I think the difference is simply that faerie is using section relocations and object is using symbol relocations. This isn't a case of being smarter, it's just different, and may not be what you want. For local symbols this should make no difference, since the symbol relocations should be replaced by relative relocations during static linking. For global symbols the difference is that symbol relocations can be preempted (e.g with LD_PRELOAD), so they remain as symbol relocations after static linking.

The trap tables are currently declared as global symbols. They probably don't need to be global though? So you could change them to `Linkage::Local` or `Linkage::Hidden` to get simpler relocations and less dynamic symbols. I don't think there's currently a way to get cranelift-object to do exactly what cranelift-faerie does, but it should be possible without changes to `object` if it turns out to be required.

Another difference is that faerie is doing the equivalent of -ffunction-sections and -fdata-sections, but I don't think that will affect what you are seeing here. (Related to that, object in theory can support separate function/data sections too if required, and it wouldn't be hard to add configuration options to cranelift-object to support this.)

froydnj:
> The trap tables are currently declared as global symbols. They probably don't need to be global though? So you could change them to `Linkage::Local` or `Linkage::Hidden` to get simpler relocations and less dynamic symbols. I don't think there's currently a way to get cranelift-object to do exactly what cranelift-faerie does, but it should be possible without changes to `object` if it turns out to be required.

Ah, thanks for this tip.  Exporting fewer symbols from the shared library is probably a good thing anyway.

Yes, changing various things to non-`Export` linkage brings the relocations in line with what `faerie` does and makes `lucet-objdump` handle things without further changes.  For reasons that I haven't been able to figure out, it looks like changing the trap tables to `Local` causes `cranelift-object` to not correctly output the trap tables for `lucet_probestack`, so there is some weird thing happening.

froydnj:
After rebasing on top of #480 (thanks @philipc for the hint!), `helpers/lucet-toolchain-tests/objdump.sh` passes for me locally.  I diffed output from `lucetc -o $FILE lucetc/tests/wasm/icall_sparse.wat; lucet-objdump $FILE` for `faerie`-using `lucetc` and `object`-using `lucetc`:

```diff
--- /dev/fd/63  2020-04-02 13:33:37.290884668 -0400
+++ /dev/fd/62  2020-04-02 13:33:37.290884668 -0400
@@ -1,11 +1,11 @@
 Required Symbols:
   lucet_module                  : exists
 Native module components:
-  module_data_ptr               : exists; address: 0x2007c8
+  module_data_ptr               : exists; address: 0x200770
   module_data_len               : 401
-  tables_ptr                    : exists; address: 0x559
+  tables_ptr                    : exists; address: 0x200560
   tables_len                    : 1
-  function_manifest_ptr         : exists; address: 0x5b1
+  function_manifest_ptr         : exists; address: 0x200570
   function_manifest_len         : 7
 
 Module:
@@ -16,7 +16,7 @@
   MISSING!
 
 Tables:
-  Table 0: [TableElement { ty: 18446744073709551615, func: 0 }, TableElement { ty: 0, func: 1120 }, TableElement { ty: 0, func: 1136 }, TableElement { ty: 1, func: 1152 }, TableElement { ty: 18446744073709551615, func: 0 }, TableElement { ty: 18446744073709551615, func: 0 }]
+  Table 0: [TableElement { ty: 18446744073709551615, func: 0 }, TableElement { ty: 0, func: 1079 }, TableElement { ty: 0, func: 1090 }, TableElement { ty: 1, func: 1064 }, TableElement { ty: 18446744073709551615, func: 0 }, TableElement { ty: 18446744073709551615, func: 0 }]
 
 Signatures:
   Signature 0: () -> I32
@@ -26,7 +26,7 @@
 Functions:
   Function 0 (name: guest_func_foo):
     Signature (index 2): (I32) -> I32
-    Start: 0x000004c0
+    Start: 0x0000044d
     Code length: 57 bytes
     Trap information:
       4 traps ...
@@ -36,21 +36,21 @@
         $+0x0035: StackOverflow
   Function 1 (name: guest_func_1):
     Signature (index 0): () -> I32
-    Start: 0x00000460
+    Start: 0x00000437
     Code length: 11 bytes
     Trap information:
       1 trap ...
         $+0x0000: StackOverflow
   Function 2 (name: guest_func_2):
     Signature (index 0): () -> I32
-    Start: 0x00000470
+    Start: 0x00000442
     Code length: 11 bytes
     Trap information:
       1 trap ...
         $+0x0000: StackOverflow
   Function 3 (name: guest_func_3):
     Signature (index 1): (F32) -> F32
-    Start: 0x00000480
+    Start: 0x00000428
     Code length: 15 bytes
     Trap information:
       1 trap ...
@@ -67,7 +67,7 @@
     Code length: 0 bytes
   Function 6 (name: lucet_probestack):
     Signature (index 0): () -> I32
-    Start: 0x00000490
+    Start: 0x00000486
     Code length: 43 bytes
     Trap information:
       2 traps ...
```

The differences are due to `object` putting things in `.data.rel.ro` (vs. `faerie` putting those same things in `.rodata`) and thus changing addresses of various things.

iximeow:
Nice! That's a good diff :) I see functions are no longer 16-byte aligned, which is a bit surprising - I thought that was controlled by `cranelift-module`? That's not necessarily a severe issue, but 16-byte alignment is noted as a best practice by Intel and AMD for various cache friendliness reasons.

> a missing .note.GNU-stack section in a single object file in the entire link means the resulting linked object must (being conservative) have an executable stack

I see https://github.com/bytecodealliance/wasmtime/pull/1455 has landed - is there anything else to do for this here? Just updating wasmtime?

froydnj:
> Nice! That's a good diff :) I see functions are no longer 16-byte aligned, which is a bit surprising - I thought that was controlled by `cranelift-module`? That's not necessarily a severe issue, but 16-byte alignment is noted as a best practice by Intel and AMD for various cache friendliness reasons.

Ah, good catch.  It looks like `faerie` sets the alignment of executable sections to 16 by default:

https://github.com/m4b/faerie/blob/e91f4f33858f56f55744ec69e81096e2f213337e/src/elf.rs#L281-L289

`cranelift-object` supports declaring function alignment before the backend is created:

https://github.com/bytecodealliance/wasmtime/blob/3481e59673f9e7a90814fc714a6a374e9cc03207/cranelift/object/src/backend.rs#L52-L56

but defaulting to 1, otherwise.  I'll fix that.

> > a missing .note.GNU-stack section in a single object file in the entire link means the resulting linked object must (being conservative) have an executable stack
> 
> I see [bytecodealliance/wasmtime#1455](https://github.com/bytecodealliance/wasmtime/pull/1455) has landed - is there anything else to do for this here? Just updating wasmtime?

Yes, I should do that before this lands as well.  As @awortman-fastly noted upthread, it doesn't matter much for Lucet's case, but this would matter for something like Firefox's use case, where it just `dlopen`s the shared library and doesn't manage stacks itself.

froydnj:
Rebased on master to pick up the wasmtime update for non-executable stacks, and changed the code to use 16-byte alignment for functions.

PR #473, opened 2020-03-31
Update instructions for MacOS; homebrew installs LLVM 10 now

jedisct1:


pchickey:
I think you need to rebase on master to get the fix to that cargo-audit test failure

PR #474, opened 2020-03-31
use object instead of goblin in lucet-objdump

froydnj:
Over in #470, we discovered that `object` and `faerie` have different
approaches to writing relocations in shared objects.  In a nutshell,
`object`'s approach requires applying relocations to the shared object's
data whereas `faerie`'s does not (at least for the current structure of
`lucet-objdump`).

This change represents an intermediate step towards accommodating
`object`'s relocation approach by rewriting `lucet-objdump` to use
`object` instead of `goblin` for reading files.  `object` is a little
easier to work with, a little more abstract than `goblin`, and various
other bits of Lucet already use `object` anyway.

There are no differences in `lucet-objdump`'s output on a representative
file when `lucet-objdump` is compiled with and without this patch.

### DIFF
--- a/lucet-objdump/src/main.rs
+++ b/lucet-objdump/src/main.rs
@@ -7,7 +7,7 @@ use lucet_module::{
 
 use byteorder::{LittleEndian, ReadBytesExt};
 use colored::Colorize;
-use goblin::{elf, Object};
+use object::{Object, ObjectSection, SymbolKind, SymbolScope};
 use std::env;
 use std::fs::File;
 use std::io::Cursor;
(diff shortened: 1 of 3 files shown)


lucet-objdump/src/main.rs
```diff
@@ -118,19 +110,10 @@ impl<'a> ArtifactSummary<'a> {
     }
 
     fn get_func_name_for_addr(&self, addr: u64) -> Option<&str> {
```
awortman-fastly: Strictly speaking this is no longer checking the symbol being named is a function, but checking for that doesn't make sense for objects like PEs anyway - mind renaming this to just `get_symbol_name_for_addr`?

froydnj: Makes sense!

awortman-fastly:
grr. wrong tab

iximeow:
(tweaked the commit just a smidge to keep a link (`https://github.com/bytecodealliance/lucet/pull/470`) alongside `#470` in the message)

froydnj:
@awortman-fastly @iximeow Thank you!

@awortman-fastly Did you get a notification about review being re-requested?  I clicked the "request re-review" button on Github but it didn't seem to be changing state.

awortman-fastly:
I don't think I saw anything about review re-requests, I've seen it fail to do anything once or twice now, unfortunately. I just looked back at this tab and saw a force push I didn't recall being present before.

froydnj:
@awortman-fastly OK, thank you for the information!

PR #475, opened 2020-03-31
Conveying expectations:  Bad instances take up capacity.

fst-crenshaw:
One of the things that surprised me in the `alloc` tests was that a failed instance still takes up the slot capacity of a region.  I've written a test that demonstrates this scenario.  

- If this behaviour is expected, then I'm glad to have a test to demonstrate that the behaviour is expected.  No other test seems to exercise this logic and I prefer not to leave things unsaid.
- If this behaviour is unexpected, then we can take other steps.


fst-crenshaw:
I'm happy to take on the task of fixing this issue.  If others agree, I'm happy to merge this PR when the CI gives me the green circles of happy.

PR #476, opened 2020-03-31
update to wasmtime 0.15, cranelift 0.62

pchickey:


fst-crenshaw:
When I pull down your branch and run `cargo test` from `~/lucet`, I see the following:

```
running 70 tests
test address ... fatal runtime error: failed to initiate panic, error 5
error: test failed, to rerun pass '-p lucet-spectest --test wasm-spec'

Caused by:
  process didn't exit successfully: `/home/vagrant/Repos/lucet/target/debug/deps/wasm_spec-5a0ee7a11f204763` (signal: 6, SIGABRT: process abort signal)
```

I know we are still working towards getting the spectest to pass, but it appears to fail in a new way for me.  Is this expected?

pchickey:
The failure you discovered in lucet-spectest also reproduces in master. It appears to have been introduced by #429. It is not present at the merge of the PR prior to that, #471 (11f4b6bb)

katie-martin-fastly:
Taking a look at this now @pchickey :slightly_smiling_face:

PR #477, opened 2020-03-31
[WIP] Improving error handling around instance creation and slot management.

fst-crenshaw:
Following up on the unintended behaviour discovered in https://github.com/bytecodealliance/lucet/pull/475, this PR improves error handling around instance creation and slot management.  

**Shorthand Goal for this PR**: If I have 10 Slots in a Region, and I create 10 bad instances, I should still have 10 free Slots.


lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -127,6 +131,7 @@ impl RegionInternal for MmapRegion {
             region,
         };
 
+        // TLC: Here is another early return from this function.  Hm.
```
acfoltzer, line 134: Once the `Alloc` is created, you don't have to worry anymore about early returns, since the `Drop` impl on `Alloc` will put the slot back on the freelist

fst-crenshaw: Thanks for this guidance!

fst-crenshaw: I added a bit of documentation about this in the code.  This way, Future Tanya won't get confused again.  😊

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -103,8 +105,15 @@ impl RegionInternal for MmapRegion {
         ]
         .iter()
         {
-            // eprintln!("setting r/w {:p}[{:x}]", *ptr, len);
-            unsafe { mprotect(*ptr, *len, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE)? };
+            unsafe {
+                let res = mprotect(*ptr, *len, ProtFlags::PROT_READ | ProtFlags::PROT_WRITE);
+
+                // If the mprotect() call failed, manually return the slot.
+                if let Err(e) = res {
```
fst-crenshaw, line 112: Talking with @pchickey I am starting feel that perhaps a panic! would be good here.  Thinking through the ways that mprotect() could fail, I'm wondering how one might recover gracefully from that.

acfoltzer: I was wrong, an `expect` is the right call here, much as it already is when we drop the alloc and `mprotect` back to `PROT_NONE`

fst-crenshaw: Thanks for this additional comment.  Very helpful.

fst-crenshaw:
This is ready for another review.

PR #478, opened 2020-04-01
Faulted guests can run!

iximeow:
Over in #429 we adjusted `KillSwitch` to be useful in that a timeout before a guest begins is upheld, rather than ignored. As it turns out, our tests didn't cover one of the expectations we have around instance lifetimes: after a fault, instances _can_ be run without reset. Consequently, users of lucet-runtime that ran instances post-fault without reset (such as `lucet-spectest`) caught a bug: rerunning after a fault could cause us to panic in a way that aborts the process.

In normal guest exits we disable termination on the instance's `KillState`, then reset it to a new one so new `KillSwitch` can cancel execution for some eventual function call. To avoid the bad `KillState` state, we just needed to do the same when exiting through a signal handler, too.

The first commit here introduces a test that fails like
```
error: test failed, to rerun pass '-p lucet-runtime --test guest_fault'
                                                                     
Caused by:                                                                                
  process didn't exit successfully: `/home/iximeow/lucet/target/debug/deps/guest_fault-88464978bc082788` (signal: 6, SIGABRT: process abort signal)
```
(`SIGABRT` because we panic where there isn't full unwind information, and libunwind falls over)

where the second commit fixes this issue, and gets us back to passing all but five spectests: `data`, `elem`, `globals`, `imports`, `linking`!

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
@@ -65,12 +65,6 @@
 //!   - terminable: `true`
 //!   - termination result: `Ok(KillSuccess::Signalled)`
 //!   - execution_domain: `Guest`
-//! * `Guest function returns`
-//!   - terminable: `true`
-//!   - termination result: `Ok(KillSuccess::Pending)`
-//!   - execution_domain: `Guest`
-//!   - If the guest function has already returned, the instance will be placed into the
-//!     `Terminated` domain, and will not run any guest the next time it is run.
 //! * `Hostcall called`
 //!   - termination result: `Ok(KillSuccess::Pending)`
 //!   - execution_domain: `Hostcall`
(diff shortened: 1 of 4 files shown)


lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -215,7 +216,7 @@ macro_rules! timeout_tests {
             }
 
             // An instance that has faulted is not terminable.
-            assert_eq!(kill_switch.terminate(), Err(KillError::NotTerminable));
```
iximeow, line 218: The comment here says `not terminable`, and in a way the `Err` _does_ mean `not terminable`, but `Invalid` means the underlying `KillState` is gone and the `KillSwitch` is no longer valid, which is slightly different from the `NotTerminable` result. I think this is a case of the test being subtly incorrect, but I know we've talked a bit about this granularity of `KillError` maybe being a misdesigned interface anyway.

lucet-runtime/lucet-runtime-internals/src/instance/execution.rs
```diff
@@ -244,6 +244,11 @@ pub unsafe extern "C" fn exit_guest_region(instance: *mut Instance) {
                 // We finished executing the code in our guest region normally! We should reset
                 // the kill state, invalidating any existing killswitches' weak references.
                 mem::drop(current_domain);
+                // There should be only one strong reference to `kill_state`, so as a consistency
+                // check ensure that's still true. This is necessary so that when we drop this
+                // `Arc`, weak refs are no longer valid. If this assert fails, something cloned
+                // `KillState`, or a `KillSwitch` has upgraded it's ref - both of these are errors!
```
katie-martin-fastly, line 250:
```suggestion
                // `KillState`, or a `KillSwitch` has upgraded its ref - both of these are errors!
```

I appreciate the explanatory comments here. :heart:

lucet-runtime/lucet-runtime-tests/src/timeout.rs
```diff
@@ -184,8 +184,9 @@ macro_rules! timeout_tests {
                 res => panic!("unexpected result: {:?}", res),
             }
 
-            // If we try to terminate after the instance ran, the kill switch will fail, and the
-            // the instance will run normally the next time around.
+            // If we try to terminate after the instance ran, the kill switch will fail - the
+            // function we called is no longer running - and the the instance will run normally the
+            // next time around.
```
katie-martin-fastly, line 189: Good clarification here.

PR #479, opened 2020-04-01
In the guest_fault suite, check that signal handlers are restored

acfoltzer:
There was a bug in `illegal_instr_manuals_signal()` where the Lucet signal handler would remain installed for the remainder of the test suite execution. Somehow this wasn't causing any failures, but does if the test suite is run more than once.

To guard against this type of thing in the future, I added a wrapper to any tests that muck with the process's signal handlers to ensure that the handlers are the same before and after the test runs.

This is definitely a diff to view with whitespace diffs hidden, due to the indentation of the new closure wrapper.

### DIFF
--- a/lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
+++ b/lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
@@ -137,6 +137,36 @@ pub fn mock_traps_module() -> Arc<dyn Module> {
         .build()
 }
 
+pub fn with_unchanged_signal_handlers<F: FnOnce()>(f: F) {
+    fn get_handlers() -> Vec<libc::sigaction> {
+        use libc::*;
+        use std::mem::MaybeUninit;
+        const SIGNALS: &'static [c_int] = &[SIGBUS, SIGFPE, SIGILL, SIGSEGV, SIGALRM];
+
+        SIGNALS
+            .iter()
+            .map(|sig| unsafe {
+                let mut out = MaybeUninit::<sigaction>::uninit();
+                sigaction(*sig, std::ptr::null(), out.as_mut_ptr());
+                out.assume_init()
+            })
+            .collect()
+    }
+
+    let before = get_handlers();
+
+    f();
+
+    let after = get_handlers();
+
+    for (before, after) in before.into_iter().zip(after.into_iter()) {
+        assert_eq!(
+            before.sa_sigaction, after.sa_sigaction,
+            "signal handlers match before and after"
+        );
+    }
+}
+
 #[macro_export]
 macro_rules! guest_fault_tests {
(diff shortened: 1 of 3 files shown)


lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
```diff
@@ -286,46 +318,50 @@ macro_rules! guest_fault_tests {
...
+                    match inst.run("illegal_instr", &[]) {
+                        Err(Error::RuntimeFault(details)) => {
+                            assert_eq!(details.trapcode, Some(TrapCode::BadSignature));
+                        }
+                        res => panic!("unexpected result: {:?}", res),
                     }
-                    res => panic!("unexpected result: {:?}", res),
-                }
+
+                    lucet_runtime::remove_lucet_signal_handler();
```
acfoltzer, line 363: :bug:

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
```diff
@@ -563,53 +599,54 @@ macro_rules! guest_fault_tests {
...
-                    sigaction(
-                        INVALID_PERMISSION_SIGNAL,
-                        &SigAction::new(SigHandler::SigDfl, SaFlags::SA_RESTART, SigSet::empty()),
-                    )
-                    .expect("sigaction succeeds");
-                }
+                    // clean up
+                    unsafe {
+                        recoverable_ptr_teardown();
+                        sigaction(INVALID_PERMISSION_SIGNAL, &before_sa)
```
acfoltzer, line 644: :bug: 

Turns out `SigDfl` is not the same as the handler Rust installs for stack overflows.

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
```diff
@@ -137,6 +137,36 @@ pub fn mock_traps_module() -> Arc<dyn Module> {
...
+    let before = get_handlers();
+
+    f();
+
+    let after = get_handlers();
+
+    for (before, after) in before.into_iter().zip(after.into_iter()) {
+        assert_eq!(
+            before.sa_sigaction, after.sa_sigaction,
+            "signal handlers match before and after"
```
awortman-fastly, line 165: :+1:

lucet-runtime/lucet-runtime-tests/src/guest_fault.rs
awortman-fastly: this should be a no-op, since we've opted to manually control signal handling, right?

acfoltzer: It is for this test... but not for ones that might run after this one :imp:

PR #481, opened 2020-04-02
lucetc: declare more things as local

froydnj:
From a functionality standpoint, this is not strictly necessary.
However, it does two things: it exports less from the static library,
which seems like a good thing, and it encourages `cranelift-object` to
more closely match `cranelift-faerie` in terms of the relocations it
uses in the final output, which is good for #470.

### DIFF
--- a/lucetc/src/compiler.rs
+++ b/lucetc/src/compiler.rs
@@ -371,7 +371,7 @@ impl<'a> Compiler<'a> {
         function_manifest_ctx.define(function_manifest_bytes.into_inner().into());
         let manifest_data_id = self.clif_module.declare_data(
             FUNCTION_MANIFEST_SYM,
-            ClifLinkage::Export,
+            ClifLinkage::Local,
             false,
             false,
             None,
@@ -589,7 +589,7 @@ impl TrapSites {
         trap_sym_ctx.define(self.serialize());
 
         let trap_data_id =
-            module.declare_data(&trap_sym, ClifLinkage::Export, false, false, None)?;
+            module.declare_data(&trap_sym, ClifLinkage::Local, false, false, None)?;
 
         module.define_data(trap_data_id, &trap_sym_ctx)?;
 
--- a/lucetc/src/decls.rs
+++ b/lucetc/src/decls.rs
@@ -254,14 +254,14 @@ impl<'a> ModuleDecls<'a> {
         for ix in 0..info.tables.len() {
             let def_symbol = format!("guest_table_{}", ix);
             let def_data_id =
-                clif_module.declare_data(&def_symbol, Linkage::Export, false, false, None)?;
+                clif_module.declare_data(&def_symbol, Linkage::Local, false, false, None)?;
             let def_name = Name::new_data(def_symbol, def_data_id);
 
             table_names.push(def_name);
         }
 
         let tables_list_id =
-            clif_module.declare_data(TABLE_SYM, Linkage::Export, false, false, None)?;
+            clif_module.declare_data(TABLE_SYM, Linkage::Local, false, false, None)?;
         let tables_list = Name::new_data(TABLE_SYM.to_string(), tables_list_id);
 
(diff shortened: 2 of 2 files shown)


froydnj:
I do not understand the mac failures:

```
test timeout_in_hostcall ... FAILED
thread 'killswitch_2' panicked at 'assertion failed: `(left == right)`
  left: `Ok(Signalled)`,
 right: `Err(NotTerminable)`', lucet-runtime/tests/timeout.rs:3:1
thread 'killswitch_1' panicked at 'assertion failed: `(left == right)`
  left: `Err(NotTerminable)`,
 right: `Ok(Signalled)`', lucet-runtime/tests/timeout.rs:3:1
thread 'killswitch' panicked at 'assertion failed: `(left == right)`
  left: `Err(Invalid)`,
 right: `Ok(Pending)`', lucet-runtime/tests/timeout.rs:3:1
test double_timeout ... FAILED
test timeout_in_guest ... ok

failures:

---- timeout_in_hostcall stdout ----
thread 'timeout_in_hostcall' panicked at 'unexpected result: Ok(Returned(UntypedRetVal { fp: __m128(0.0, 0.0, 0.0, 0.0), gp: 1 }))', lucet-runtime/tests/timeout.rs:3:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- double_timeout stdout ----
thread 'double_timeout' panicked at 'killswitch_1 did not panic: Any', src/libcore/result.rs:1165:5
```

Is it possible that making certain symbols local to the library caused these tests to malfunction because the tests can't access certain symbols to call?  (The tests appear to be completely mocked, so it doesn't seem like that could be the case, but I am completely unfamiliar with this part of lucet.)

katie-martin-fastly:
Hello @froydnj!

```
thread 'killswitch_2' panicked at 'assertion failed: `(left == right)`
  left: `Ok(Signalled)`,
 right: `Err(NotTerminable)`', lucet-runtime/tests/timeout.rs:3:1
thread 'killswitch_1' panicked at 'assertion failed: `(left == right)`
  left: `Err(NotTerminable)`,
 right: `Ok(Signalled)`', lucet-runtime/tests/timeout.rs:3:1
```

So, this failure would be coming from [this test](https://github.com/bytecodealliance/lucet/blob/d1ad185a38ef13d9bdcca31de09ad08ccc1503e9/lucet-runtime/lucet-runtime-tests/src/timeout.rs#L326-L370) in `lucet-runtime/lucet-runtime-tests/src/timeout.rs`.

As for the [second test](https://github.com/bytecodealliance/lucet/blob/d1ad185a38ef13d9bdcca31de09ad08ccc1503e9/lucet-runtime/lucet-runtime-tests/src/timeout.rs#L226-L260), it seems like a very similar situation. This is somehow resolving _after_ our [slow hostcall](https://github.com/bytecodealliance/lucet/blob/d1ad185a38ef13d9bdcca31de09ad08ccc1503e9/lucet-runtime/lucet-runtime-tests/src/timeout.rs#L79-L86), even though we should be terminating the test instance before that hostcall finishes.

In each of these cases, threads seem to be running in the wrong order? I don't know _why_ that is specific to Mac OS, but hopefully this points us in the right direction. FWIW, we only recently reenabled Mac CI in [caf30a6](https://github.com/bytecodealliance/lucet/commit/caf30a6fe767d123633eed752eeb6574d6e34b0f), so it could be possible that these tests are flaky for unrelated reasons.

froydnj:
@katie-martin-fastly, thank you for the explanation!

I have definitely seen different thread behavior on Mac relative to Linux for cases like this.  I wonder if explicitly waiting for the thread closures to be reached prior to starting the instance(s) would make the tests any more stable, since it's not guaranteed that the killswitch threads are actually running when the instance(s) start running.  (I guess it's not guaranteed that the killswitch threads get rescheduled even after they started...)

I retriggered the tests, and they seem to have passed this time!

PR #482, opened 2020-04-02
bump wasmtime

froydnj:
This change picks up the change to `cranelift-object` to use non-executable stacks by default.


froydnj:
Meh, somehow lucet is using cranelift 0.62, but that version is newer than the wasmtime repo?  Need to investigate further.

pchickey:
I pointed at the v0.15.0 tag in wasmtime, which is three commits ahead of the actual PR that created 0.15.0 / 0.62.0, all of those were fixes to wasi-common. I believe that the crates.io published version is at the tag, rather than the PR.

pchickey:
Ah, I see now, it appears that the v0.15.0 tag is not on the same branch as master...

PR #485, opened 2020-04-04
Add instrumentation to force specific forms of races

iximeow:
This sprinkles in some points where, with a crate flag, we can force `lucet-runtime` to hang in race-sensitive code in ways we want to explicitly test. I think this is a good way for us to robustly test timing both for [platform-dependent ordering](https://github.com/bytecodealliance/lucet/pull/481#issuecomment-607923040) and for timing conditions that may be hard to otherwise artificially introduce (such as "two timeouts occur at the same time").

I've rewritten one test, that we correctly handle timeouts in guest code, here. By the time this is a real PR, I plan on porting most of the tests in `lucet-runtime-tests`' `timeouts` suite here, but for reason I don't yet understand these changes seem to have tripped up `guest_fault`. :unamused:

### DIFF
--- a/lucet-concurrency-tests/tests/killswitch.rs
+++ b/lucet-concurrency-tests/tests/killswitch.rs
@@ -1,0 +1,114 @@
+use lucet_runtime::{Error, KillSuccess, Limits, Region, TerminationDetails};
+use std::sync::Arc;
+use std::thread;
+
+use lucet_module::FunctionPointer;
+use lucet_runtime::MmapRegion;
+use lucet_runtime_internals::lock_testpoints::Syncpoint;
+use lucet_runtime_internals::module::Module;
+use lucet_runtime_internals::module::{MockExportBuilder, MockModuleBuilder};
+use lucet_runtime_internals::vmctx::lucet_vmctx;
+
+static mut ENTERING_GUEST: Option<Syncpoint> = None;
+
+pub fn mock_traps_module() -> Arc<dyn Module> {
+    extern "C" fn onetwothree(_vmctx: *mut lucet_vmctx) -> std::os::raw::c_int {
+        123
+    }
+
+    extern "C" fn infinite_loop(_vmctx: *mut lucet_vmctx) {
+        unsafe {
+            ENTERING_GUEST.as_ref().unwrap().check();
+        }
+        loop {}
+    }
+
+    extern "C" fn fatal(vmctx: *mut lucet_vmctx) {
+        extern "C" {
+            fn lucet_vmctx_get_heap(vmctx: *mut lucet_vmctx) -> *mut u8;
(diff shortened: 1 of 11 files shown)


lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,114 @@
...
+use lucet_module::FunctionPointer;
+use lucet_runtime::MmapRegion;
+use lucet_runtime_internals::lock_testpoints::Syncpoint;
+use lucet_runtime_internals::module::Module;
+use lucet_runtime_internals::module::{MockExportBuilder, MockModuleBuilder};
+use lucet_runtime_internals::vmctx::lucet_vmctx;
+
+static mut ENTERING_GUEST: Option<Syncpoint> = None;
+
+pub fn mock_traps_module() -> Arc<dyn Module> {
```
iximeow, line 14: This is lifted pretty directly from `timeouts.rs`, and includes functionality enough that I think it's sufficient for here too.

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::thread;
+
+use lucet_module::FunctionPointer;
+use lucet_runtime::MmapRegion;
+use lucet_runtime_internals::lock_testpoints::Syncpoint;
+use lucet_runtime_internals::module::Module;
+use lucet_runtime_internals::module::{MockExportBuilder, MockModuleBuilder};
+use lucet_runtime_internals::vmctx::lucet_vmctx;
+
+static mut ENTERING_GUEST: Option<Syncpoint> = None;
```
iximeow, line 12: I tried making this a `thread_local` only to eventually realize that actually doesn't cut it - `inst.run` happens in a different thread than the initial setup, so the test thread and instance thread have different `Syncpoint` :(

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,114 @@
...
+                Err(Error::RuntimeTerminated(TerminationDetails::Remote)) => {
+                    // this is what we want!
+                }
+                res => panic!("unexpected result: {:?}", res),
+            }
+        })
+        .expect("can spawn a thread");
+
+    let terminator = in_guest.wait_and_then(move || {
+        thread::spawn(move || {
```
iximeow, line 107: This should look weird! `terminate` blocks until the guest is in fact stopped. Since we know the thread will be sent a signal, I'm not sure if we really want to or need to wait until termination completes.

iximeow:
closing in favor of #497, which is this, but more complete. I pushed to the wrong branch and just made a new PR before remembering this still existed.

PR #486, opened 2020-04-06
When the heap is not page aligned, fail instead of `lucet_bail`

fst-crenshaw:
In a conversation over in https://github.com/bytecodealliance/lucet/pull/458, it was discovered that when creating an Instance, if its heap is not page-aligned, we'd like for the lucet runtime to panic.  It's difficult to imagine how one might recover from such misalignment.  

This PR makes this small fix.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
@@ -91,9 +91,8 @@ impl RegionInternal for MmapRegion {
             .pop()
             .ok_or(Error::RegionFull(self.capacity))?;
 
-        if slot.heap as usize % host_page_size() != 0 {
-            lucet_bail!("heap is not page-aligned; this is a bug");
-        }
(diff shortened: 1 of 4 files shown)


lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -91,9 +91,8 @@ impl RegionInternal for MmapRegion {
             .pop()
             .ok_or(Error::RegionFull(self.capacity))?;
 
-        if slot.heap as usize % host_page_size() != 0 {
-            lucet_bail!("heap is not page-aligned; this is a bug");
-        }
+        slot.is_heap_page_aligned()
+            .expect("Heap must be page aligned");
```
acfoltzer, line 95: I don't think the original `Result`-oriented code is worth preserving. A `Slot` should never have any fields that aren't page-aligned.

```suggestion
        assert_neq!(slot.heap as usize % host_page_size(), 0, "heap must be page-aligned");
```

PR #492, opened 2020-04-08
Add userfaultfd-based Region backend

tyler:
This pull request integrates Fastly's `userfaultfd` based memory management backend as an alternate Region implementation.

Userfaultfd is a Linux-specific mechanism for handling page faults within userspace. (See https://www.kernel.org/doc/html/latest/admin-guide/mm/userfaultfd.html for a good technical explanation of it.) For Lucet, the use case is that when an Instance is started, none of the linear memory has to be copied initially.

We register the entire region with userfaultfd. When an instance is started, we set up the stacks, metadata, etc as normal, but we leave the pages of the linear memory "missing". When the instance starts and tries to access one of the linear memory pages, this triggers a page fault, as there is no physical memory backing the virtual memory. Since the region is registered with userfaultfd, this triggers a message to be sent to the userfaultfd handler thread. The handler thread determines which instance and module has faulted, and copies the necessary memory into place, before reawakening the instance thread.

At its core what this does is provide a way of reducing startup time and increasing flexibility in how memory is handled at the cost of increased runtime overhead (by was of context switches especially). 

(All credit for this actually goes to @acfoltzer.)


acfoltzer:
@tyler there are a few things I'd like to add to the public API surface and documentation to go with this. It would probably be easiest if I did that on a branch and then PRed into this, is that alright with you?

lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
```diff
@@ -0,0 +1,512 @@
...
+                    }
+                    AddrLocation::SigStackGuard | AddrLocation::Unknown => {
+                        tracing::error!("UFFD pagefault at fatal location: {:?}", loc);
+                        uffd.wake(fault_page as *mut c_void, host_page_size())
+                            .map_err(|e| Error::InternalError(e.into()))?;
+                    }
+                    AddrLocation::Globals | AddrLocation::SigStack => {
+                        tracing::error!("UFFD pagefault at unexpected location: {:?}", loc);
+                        uffd.wake(fault_page as *mut c_void, host_page_size())
+                            .map_err(|e| Error::InternalError(e.into()))?;
```
fst-crenshaw, line 134: One of the things I've been seeing in the mmap-related PRs is a re-examination of errors like these.  If there had been a UFFD pagefault at an unexpected location, should the lucet runtime return its generic "InternalError" or should it just panic?  

See related PR: https://github.com/bytecodealliance/lucet/pull/486

lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
```diff
@@ -0,0 +1,512 @@
...
+        let limits = &slot.limits;
+        module.validate_runtime_spec(limits)?;
+
+        for (ptr, len) in [
+            // zero the globals
+            (slot.globals, limits.globals_size),
+            // zero the sigstack
+            (slot.sigstack, limits.signal_stack_size),
+        ]
+        .into_iter()
```
fst-crenshaw, line 239: I am new to userfaultfd, and I'm wondering about a small difference between its implementation of new_instance_with and the mmap implementation.  https://github.com/bytecodealliance/lucet/blob/83fd2248b36c780cc74c0abb49024cca4a0c053c/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs#L121-L122

lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
```diff
@@ -0,0 +1,512 @@
...
+
+        let uffd = Arc::new(
+            UffdBuilder::new()
+                .close_on_exec(true)
+                .non_blocking(true)
+                .create()
+                .map_err(|e| Error::InternalError(e.into()))?,
+        );
+
+        // map the chunk of virtual memory for all of the slots
```
fst-crenshaw, line 398: I'm guessing the answer to my question could be here.  The memory is already marked as READ/WRITE for all the slots.

acfoltzer: https://github.com/bytecodealliance/lucet/pull/492/files#diff-d382c16395ef5840f0796f78cf91450dR136-R139 is also a key part of why this works

fst-crenshaw: Hm.  If I'm reading that correctly, does that mean that the stack isn't zeroed out until it's actually needed?

acfoltzer: It may help to refer to the `ioctl_userfaultfd(2)` man page on this, since "zeropage" does a lot more than what it sounds like: http://man7.org/linux/man-pages/man2/ioctl_userfaultfd.2.html (search for `UFFDIO_ZEROPAGE` on that page).

Basically `zeropage` is one of several ways that the uffd worker thread can resolve a page fault in the memory that it has registered. The stack starts off as `mprotect` read/write but `madvise` don't need, so accesses will trigger a uffd fault. We then recognize that the fault address is in the stack, and then call `zeropage` to both zero out the memory and resolve the fault, so that the instance's thread wakes up and can make progress once again.

lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
tyler: This is a good call, it should just panic. I think @acfoltzer wanted to take care of that in a second PR. Though tbh I think we should just do it here.

lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
tyler: So yeah, that's intentional here. We're making the allocation of the stack lazy. If you look at the other lines changed in that commit, you'll see where I make page faults in the Stack area zero out the page on demand. 

That said, it does point to a lack of documentation here that I should fix.

tyler: Added some docs!

fst-crenshaw:
Now that CircleCI is running mmap and userfaultfd tests, will we be able to merge this?

acfoltzer:
> Now that CircleCI is running mmap and userfaultfd tests, will we be able to merge this?

CircleCI experiments are currently only happening on branches, but I'm hoping to get a proper PR together today.

fst-crenshaw:
So excited!

tyler:
Ready for review :)

fst-crenshaw:
The GitHub Actions tests are failing since they can't run the userfaultfd tests.  I thought I'd try a little PR to have the offensive tests get skipped by GitHub Actions, since they are being run by CircleCI.  See: https://github.com/bytecodealliance/lucet/pull/528

Issue #455, opened 2020-03-20
Instances are not wholly cleaned up when they error out due to unexpected signal stack sizes.


fst-crenshaw:
The moment when the lucet runtime switches into a guest to extract the instance's result is a function called `swap_and_return`.   If the call to `self.with_signals_on` returns with an Error resulting from an unexpected signal stack size, `swap_and_return` returns without cleaning up the borrowed instance.  As in, this doesn't happen.

```
        CURRENT_INSTANCE.with(|current_instance| {
            *current_instance.borrow_mut() = None;
        });

```

That moment is here:
https://github.com/bytecodealliance/lucet/blob/c765316fe108c10b6815236ede70b841b8ec3b57/lucet-runtime/lucet-runtime-internals/src/instance.rs#L972

To reproduce the issue:

1. Go to `~/lucet-runtime/lucet-runtime-internal`
2. Run `cargo test -- --test-threads=1`

The result:

```
---- alloc::tests::reject_sigstack_smaller_than_min stdout ----
thread 'main' panicked at 'no other instance is running on this thread', lucet-runtime/lucet-runtime-internals/src/instance.rs:956:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
failures:
  alloc::tests::reject_sigstack_smaller_than_min
test result: FAILED. 65 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
error: test failed, to rerun pass '--lib'
```

Tactically, what happens: 

This test fails (expectedly) but the cleanup doesn't happen.
alloc::tests::reject_debug_sigstack_smaller_than_12kib

So this test fails (unexpectedly):
alloc::tests::reject_sigstack_smaller_than_min

acfoltzer:
https://github.com/bytecodealliance/lucet/commit/058cc8562e58d4e1bd6ac041e0567f7c698c1c89

PR #494, opened 2020-04-10
[PR into tyler/uffd-integration] Fixes uncovered during review of #492

acfoltzer:
This first commit enables alloc tests for `UffdRegion`, adds a new test, and fixes some things:

- Instantiates the suite in `lucet_runtime_internals::alloc::tests` for `UffdRegion`

- Fixes early return issues in `UffdRegion` similar to #455

- Adds a test to show that the per-instance heap limit applies to runtime expansions, not just initial instantiation

- Refactors `validate_runtime_spec` to take the per-instance heap limit as an additional argument. This centralizes the logic for rejecting initially-oversized heap limits, and makes it clearer what's happening in each region's instantiation logic.

- Removes the `UffdRegion`'s assertion that signal stack size is a multiple of page size. Since the user can now control this as a parameter, we reject it gracefully when validating `Limits` rather than panicking.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
@@ -1017,7 +1017,7 @@ mod mmap {
     alloc_tests!(crate::region::mmap::MmapRegion);
 }
 
-#[cfg(all(test, feature = "uffd"))]
+#[cfg(all(test, target_os = "linux", feature = "uffd"))]
 mod uffd {
     alloc_tests!(crate::region::uffd::UffdRegion);
 }
--- a/lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
@@ -240,9 +240,9 @@ macro_rules! alloc_tests {
             max_size: Some(EXPANDPASTLIMIT_MAX_SIZE),
         };
 
-        /// This test shows that a heap refuses to grow past the alloc limits, even if the runtime
-        /// spec says it can grow bigger. This test uses a region with multiple slots in order to
-        /// exercise more edge cases with adjacent managed memory.
+        /// This test shows that a heap refuses to grow past the region's limits, even if the
+        /// runtime spec says it can grow bigger. This test uses a region with multiple slots in
+        /// order to exercise more edge cases with adjacent managed memory.
         #[test]
         fn expand_past_heap_limit() {
             let region = TestRegion::create(10, &LIMITS).expect("region created");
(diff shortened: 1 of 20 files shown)


lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
```diff
@@ -964,4 +1013,11 @@ macro_rules! alloc_tests {
 }
 
 #[cfg(test)]
-alloc_tests!(crate::region::mmap::MmapRegion);
+mod mmap {
```
tyler, line 1016: Wait, did I miss the alloc tests in my conversion??

lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
```diff
@@ -376,10 +352,6 @@ impl Drop for UffdRegion {
 
 impl UffdRegion {
     pub fn create(instance_capacity: usize, limits: &Limits) -> Result<Arc<Self>, Error> {
-        assert!(
```
tyler, line 379: I assume this is enforced elsewhere?

lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
acfoltzer: Yeah, though since they're part of `internals` rather than `lucet-runtime-tests`, I don't blame you for missing them.

lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
```diff
@@ -191,30 +190,8 @@ impl RegionInternal for UffdRegion {
...
-            Ordering::Equal => (),
-            Ordering::Greater => {
-                return Err(Error::InvalidArgument(
-                    "heap memory size requested for instance is larger than slot allows",
-                ))
-            }
-        }
-        module.validate_runtime_spec(&limits)?;
+        let limits = self.get_limits();
+        module.validate_runtime_spec(&limits, heap_memory_size_limit)?;
```
fst-crenshaw, line 194: Thank you for refactoring this.  I wanted to do this but hesitated to alter the function signature of validate_runtime_spec.

acfoltzer:
This PR now also enables the `uffd` feature by default, but filters on OS.

Notably, this should get us building and running uffd in Linux CI.

It turns out to be a tremendous pain to enable a feature flag for just one crate within a workspace. The situation is [being addressed][1], but in the meantime I believe the best route forward is to just have uffd on by default for Linux.

[1]: rust-lang/cargo#5364

acfoltzer:
Okay, the CI issue is still outstanding, but I'm happy with merging this into your branch, @tyler

PR #495, opened 2020-04-10
Add killswitch documentation

iximeow:
This adds a decently-sized chapter to our `Lucet` mdbook discussing `KillSwitch`, its implementation, and correctness in the face of various edge cases we must consider. It even has a diagram!


docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
+# `KillSwitch`
+
+`KillSwitch` is a mechanism by which users of `lucet-runtime` can
+asynchronously request, very sternly, that an instance be disabled from
```
fst-crenshaw, line 4: It feels like phrase "very sternly" -- as well as the specificity of the subsequent sentences -- suggests that there are scenarios in which one can request a KillSwitch but it doesn't have the intended effect.  If that's the case, would it be worthwhile to briefly introduce error scenarios.

iximeow: Yes, definitely! I hint later on at the implication made here, but this should at least directly reference that related reading.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
...
+they are originally defined.
+
+As much as is possible, `KillSwitch` tries to be self-contained; no members are
+public, and it tries to avoid leaking details of its state-keeping into public
+interfaces. Currently, `lucet-runtime` is heavily dependent on POSIX
+thread-directed signals to implement guest timeouts - for non-POSIX platforms
+alternate implementations may be plausible, but need careful consideration of
+the race conditions that can arise from other platform-specific functionality.
+
+`KillSwitch` fundamentally relies on two piece of state for safe operation,
```
fst-crenshaw, line 69: nit: two pieces.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+runtime bookkeeping including updating `execution_domain` to `Domain::Hostcall`
+(`begin_hostcall`) and checking that the instance can return to guest code,
+setting `execution_domain` back to `Domain::Guest` (`end_hostcall`). At some
+point the instance will hopefully exit, where in `lucet_context_backstop` we
+check that the instance may stop exiting (`KillState::terminable`). If it can
+exit, then do so. Finally, back in `lucet_runtime`, we can
+`KillState::deschedule` to tear down the last of the run-specific state - the
+`thread_id`.
+
+## Implementation Complexities (When You Have A Scheduler Full Of Demons)
```
katie-martin-fastly, line 157: :heart: I appreciate that the voice of this documentation is friendly, and this header specifically got a hearty chuckle out of me. Many devils do live in the details.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
...
+
+`KillSwitch` fundamentally relies on two piece of state for safe operation,
+which are encapsulated in a `KillState` held by the `lucet_runtime::Instance`
+it terminates:
+* `execution_domain`, a `Domain` that describes the kind of execution that is
+  currently happening in the `Instance`. This is kept in a `Mutex` since in
+  many cases it will need to be accessed either by a `KillSwitch` or
+  `KillState`, and for the duration either are considering the domain, it must
+  block other users.
+* `terminable`, an `AtomicBool` that indicates if the `Instance` may stop
```
fst-crenshaw, line 77: Would it worthwhile here to mention the characteristics which make an Instance terminable?

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Implementation Complexities (When You Have A Scheduler Full Of Demons)
+
+Many devils live in the details. The rest of this chapter will discuss the
+numerous edge cases and implementation concerns that Lucet's asynchronous
+signal implementation must consider, and arguments for its correctness in the
+face of these.
+
+First, a flow chart of the various states and their transitions:
+![state flow chart](states.png)
```
katie-martin-fastly, line 165: Second: The state transition diagram here is truly a treat. Great work @iximeow.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+
+This graph describes the various states that reflect values of
+`execution_domain`, and includes some timeout transitions reflecting a timeout
```
katie-martin-fastly, line 168:
```suggestion
`execution_domain` and `terminable`, and includes some timeout transitions reflecting a timeout
```

We include the `terminable` in our states, maybe we ought to mention that here too.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+occurring entirely in a specific domain, well away from racing with other
+machinery in wasm guest. The rest of this section discusses the correcness of
```
katie-martin-fastly, line 170:
```suggestion
machinery in a wasm guest. The rest of this section discusses the correctness of
```

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
...
+  currently happening in the `Instance`. This is kept in a `Mutex` since in
+  many cases it will need to be accessed either by a `KillSwitch` or
+  `KillState`, and for the duration either are considering the domain, it must
+  block other users.
+* `terminable`, an `AtomicBool` that indicates if the `Instance` may stop
+  executing.
+
+The astute observer may notice that `Lucet` contains both an instance `Domain`
+and an instance `State`. These discuss different aspects of the instance's
+lifecycle: `Domain` describes what the instance is doing right now, where
```
fst-crenshaw, line 82: Is it an oversimplification to write that an Instance's "Domain" is its State, but internal state that must not be exposed outside of `lucet_runtime`?  I'm not making any concrete suggestion here.  This question comes from a place of wanting to cement my own understanding.  😊

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
...
+`State` describes the state of the instance as visible outside
+`lucet_runtime`. This is fundamentally the reason why `State` does not and
+cannot describe a "making a hostcall" variant - this is not something
+`lucet_runtime` wants to expose, and at the moment we don't intend to make any
+commitments about being able to query instance-internal state like this.
+Additionally, this is why `Domain` only expresses that an instance has
```
fst-crenshaw, line 88: I don't understand the logical jump being made here.  Why does Domain only express that an instance has stopped?

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+
+Many devils live in the details. The rest of this chapter will discuss the
+numerous edge cases and implementation concerns that Lucet's asynchronous
+signal implementation must consider, and arguments for its correctness in the
+face of these.
+
+First, a flow chart of the various states and their transitions:
+![state flow chart](states.png)
+
+This graph describes the various states that reflect values of
```
katie-martin-fastly, line 167:
```suggestion
This graph describes the various states and their values of
```

Not _quite_ sure how to rework this sentense, but the first clause could be cleaned up a hair, I _think_.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+`execution_domain`, and includes some timeout transitions reflecting a timeout
+occurring entirely in a specific domain, well away from racing with other
```
katie-martin-fastly, line 169: I'm a bit confused by the `, well away from` part of this sentence. Maybe we could word that differently?

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
...
+commitments about being able to query instance-internal state like this.
+Additionally, this is why `Domain` only expresses that an instance has
+_stopped_ running, not if it successfully or unsuccessfully exited. TODO:
+there's even an argument that `Terminated` vs `Cancelled` is a stronger
+distinction than `Domain` needs to make!
+
+### Termination Mechanisms
+
+At a high level,`KillSwitch` picks one of several mechanisms to terminate an
+instance.  The choice of which mechanism depends on what kind of code is
```
fst-crenshaw, line 96: It seems that the choice of mechanism depends not just on what kind of code is running in the guest, but also whether or not it's pending.  It might be more precise to leave out the first part of this sentence and just offer that "The mechanism selected depends on `instance::execution::Domain`.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+`KillSwitch` fundamentally relies on two piece of state for safe operation,
+which are encapsulated in a `KillState` held by the `lucet_runtime::Instance`
+it terminates:
+* `execution_domain`, a `Domain` that describes the kind of execution that is
+  currently happening in the `Instance`. This is kept in a `Mutex` since in
+  many cases it will need to be accessed either by a `KillSwitch` or
+  `KillState`, and for the duration either are considering the domain, it must
+  block other users.
+* `terminable`, an `AtomicBool` that indicates if the `Instance` may stop
+  executing.
```
katie-martin-fastly, line 78: This description might benefit from mentioning how we are using `terminable` as a flag, for signal safety reasons.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
...
+
+#### Guest Signalling
+
+There are two other pieces of information attached to `KillState` that support
+the specific case where we need to send a `SIGALRM`: the thread ID we need to
+signal, and a [`Condvar`][condvar] we can wait on to know when the instance has
+been stopped. The thread ID is necessary because we don't record _where_ the
+instance is running anywhere else, and we keep it here because, so far,
+`KillState` is the only place we actually need to care. Meanwhile, the
+`Condvar` allows us to be smarter than a simple spin loop while waiting for
```
fst-crenshaw, line 121: I had to look up why Condvar "allows us to be smarter."  Would it be worthwhile to be explicit in how it makes us smart?  Perhaps, "Unlike a simple spin loop, a Condvar doesn't consume CPU time while it waits on the thread."

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+t.join().unwrap();
+```
+
+## Implementation
+
+As this section discusses implementation details of `lucet-runtime`, it
+will refer to structures and enums that are unexported.  Since there are
+no `lucet-runtime` paths for these items, this section will occasionally
+use fully-qualified paths to the `lucet_runtime_internals` path where.
+they are originally defined.
```
katie-martin-fastly, line 60:
```suggestion
no `lucet-runtime` paths for these items, this section will occasionally
use fully-qualified paths to the `lucet_runtime_internals` path where
they are originally defined.
```

Small `.` nit.

Aside from that, my technical writer brain does not like `occasionally`. Maybe we can clarify which things we're using `lucet-runtime-internals` paths. (`KillState`, and `Domain` IIRC)

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+in the program - one such example is a timeout in a hostcall that eventually
+faults; since timeouts cannot preempt hostcalls, the timeout may never be
+witnessed if the fault causes the host to never resume the Lucet instance.
+
+In this chapter we will describe both a typical usage of `KillSwitch` as a
+mechanism to enforce execution time limits, and the implementation complexities
+underlying `KillSwitch` correctness.
+
+`KillSwitch` are valid for termination only on the instance call after they are
+created, and until an instance is reset. When a call into a guest returns, the
```
katie-martin-fastly, line 18:
```suggestion
created, or until an instance is reset. When a call into a guest returns, the
```

Small nit

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,341 @@
...
+been stopped. The thread ID is necessary because we don't record _where_ the
+instance is running anywhere else, and we keep it here because, so far,
+`KillState` is the only place we actually need to care. Meanwhile, the
+`Condvar` allows us to be smarter than a simple spin loop while waiting for
+signal handling machinery to actually terminate an `Instance` in
+`Domain::Guest`.
+
+### Lifecycle
+
+Having described both `KillState` and the termination mechanisms it helps
```
fst-crenshaw, line 127: I like the description of this lifecycle.  There's something I feel unsure about that feels a little fundamental and I'm sorry I'm not getting it.  I *think* that every `Instance` has a lifecycle and that every `Instance` has a KillState that tracks that lifecycle in the way that you've described.  If so, I'd love a brief intro sentence to make this more concrete.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Termination Mechanisms
+
+At a high level,`KillSwitch` picks one of several mechanisms to terminate an
+instance.  The choice of which mechanism depends on what kind of code is
+running in the guest, indicated by the `instance::execution::Domain` enum.
+
+`Domain::Pending` is the easiest domain to stop execution in: we simply update
+the execution domain to `Cancelled` so when the instance is run we know that we
+in fact should not run it.
```
katie-martin-fastly, line 101:
```suggestion
the execution domain to `Cancelled` so when the instance is run we know that we
in fact should not run it.
```

The last part of this sentence is a little confusing, via "_run_" being used in two different tenses. Maybe we can mention briefly that once we enter a guest context, we check whether the guest was cancelled before handing control over. (via `enter_guest_region`)

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+
+`Domain::Guest` is likely to be the next most-immediate termination form, where
+`lucet_runtime` will send a thread-directed `SIGARLM` to the thread running the
```
katie-martin-fastly, line 104:
```suggestion
`Domain::Guest` is likely to be the next most-immediate termination form.
`lucet_runtime` will send a thread-directed `SIGARLM` to the thread running the
```
Rather than describing this in terms of immediacy, I'd list this case before the cancellation case. My reasoning here is that I think this is what I would think of as our "base" case. Other situations are edge cases due to the many devils in the details, but _this_ case is the one I would immediately point someone to if they were attempting to learn about our signalling.

katie-martin-fastly: The grammar suggestion above is a small nit, in the interest of short digestible sentences.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+* `A -> B`
+* `A -> D`
+* `B -> C`
+* `B -> E` (normal guest exit)
+* `B -> E` (from a guest fault/timeout)
+* `C -> B`
+* `C -> E` (hostcall terminates instance)
+* `C -> E` (hostcall observes timeout)
+  - not an internal state but we will also discuss timeouts during a hostcall fault
+* `D -> E`
```
katie-martin-fastly, line 183: I might add descriptive notes to the other cases as well, in addition to those already noted. For example, `A -> B` means the guest started. `A -> D` is a cancelled guest, and so on.

I'd add these to the headers below too, so that somebody doesn't need to reference the diagram to understand what these transitions mean.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+possible, and if the instance has not yet run, it will immediately exit with
+error when the `lucet` embedder attempts to run it. In some circumstances,
+however, a `KillSwitch` may successfully fire to no actual effect at any point
+in the program - one such example is a timeout in a hostcall that eventually
+faults; since timeouts cannot preempt hostcalls, the timeout may never be
+witnessed if the fault causes the host to never resume the Lucet instance.
+
+In this chapter we will describe both a typical usage of `KillSwitch` as a
+mechanism to enforce execution time limits, and the implementation complexities
+underlying `KillSwitch` correctness.
```
katie-martin-fastly, line 15:
```suggestion
that `KillSwitch` address to be correct.
```

"underlying" was hard for my brain to parse in the context of this sentence. Maybe that paragraph could even be divided into two sentences? "_In this chapter we will ... time limits. Then, we will discuss the complexities ..._"

docs/src/lucet-runtime/states.dot
```diff
@@ -0,0 +1,48 @@
+digraph G {
```
fst-crenshaw, line 1: Thanks so much for adding a visual to the docs.  I'm a very very visual person, so pictures always help me immensely.  There's something that seems not quite right about the picture.  States A - D all have a single domain and a single possible value for `terminable`.  State E stands apart in that it can have three possible Domain values and both possible values for `terminable`.  This may very well represent why KillSwitch was so challenging and even partially motivate the rationale for writing these beautiful docs.  That said, I wonder whether more states could be added to make the diagram less muddled?  I don't know the answer to this question.  It may very well make things worse.  😊

docs/src/lucet-runtime/killswitch.md
fst-crenshaw: Plus 1.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+
+This is a timeout while another `KillSwitch` has already fired, timing out a
+guest before exeuction. Because another `KillSwitch` must have fired for there
+to be a race, one of the two will acquire `terminable` and actually update
+`execution_domain`, while the other simply exits.
+
+### `B -> C` timeout
+
+If a timeout occurs during a transition from guest (B) to hostcall (C) code,
+there are two circumstances also contingent on if the instance has or has not
```
fst-crenshaw, line 221: Maybe, "...on whether the instance has switched to `Domain::Hostcall`.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+yet switched to `Domain::Hostcall`.
+
+#### Before switching to `Domain::Hostcall`
+
+The `KillSwitch` that fires locks `execution_domain` and sees the execution
+domain is `Domain::Guest`. It then uses typical guest termination macinery and
+signals the guest. An important correctness subtlety here is that `KillSwitch`
```
fst-crenshaw, line 228: Love.  Love.  Love

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+fault or timeout has occurred, the guest is actually in
+`lucet_runtime_internals::instance::signals::handle_signal`, and `terminable`
+may be `true` (guest fault), or `false` (handling a timeout). If the guest is
+currently handling a timeout signal, the `KillSwitch` trying to fire will fail
+to acquire `terminable` and exit with an err. However..
+
+#### Timeout while handling a guest fault
+
+In the case that a timeout occurs during a guest fault, the `KillSwitch` may
+acquire `terminable`. POSIX signal handlers are highgly constrained, see `man 7
```
fst-crenshaw, line 258: nit: highgly -> highly.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+#### After switching to `Domain::Hostcall`
+
+The `KillSwitch` that fires acquires `terminable` and then attempts to acquire
+a lock on `execution_domain`. Because the instance is switching to or has
+switched to `Domain::Hostcall`, the `KillSwitch` will select the termination
+style for hostcalls. It will update the execution domain to
+`Domain::Terminated` and the instance will return when the hostcall exits and
+the `Terminated` domain is observed.
+
+### `B -> E` timeout, during guest fault or timeout
```
fst-crenshaw, line 242: This section is really hard for me to read.  I think it's related to the visual in which state E is really a lot of subtly different kind of states.  I'm also noticing this section has the word "if" a lot.  With that in mind, it might be beneficial to break this down even further.  Do you think it would be worthwhile to have separate sections for B -> E  (guest fault) and B -> E (timeout).  ?

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+As promised, a note about what happens when a timeout occurs directly when a
+hostcall faults: since by definition the instance's `execution_domain` is
+`Domain::Hostcall`, the `KillSwitch` may or may not fire before the signal
+handler disables termination. Even if it does fire, it will lock the shared
+`execution_domain` and see `Domain::Hostcall`. It then will update this to
+`Domain::Terminated`, but since the hostcall will not resume, `end_hostcall`
+will never see that the instance should stop, and no further effect will be
+had.
+
+[1]: For exampe, the code we would _like_ to interrupt may hold locks, which we
```
fst-crenshaw, line 341: nit: exampe -> example

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+In a more unfortunate circumstance, the `KillSwitch` is what observed the
+`true` out of `terminable`. In this case, the guest observed `false` and must
+not proceed, so that whenever an imminent `SIGALRM` from the corresponding
+`KillSwitch` arrives, it will be in a guaranteed-to-be-safe spin loop, or on
+its way there with only signal-safe state.
+
+The `KillSwitch` itself will signal the guest as any other `Domain::Guest`
+interruption.
+
+### `C -> B` timeout
```
fst-crenshaw, line 300: Rather than handle the state transitions alphabetically, it might be worthwhile to discuss all the straight-forward transitions first and then offer a breakdown of the gnarly transitions to `E`.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,346 @@
...
+
+### `D -> E` timeout
+
+A timeout while observing a cancelled guest will have no effect - a timeout
+must have occurred already, so the `KillSwitch` that fired will not acquire
+`terminable`, and will return without ceremony.
+
+### Timeout in hostcall fault
+
+As promised, a note about what happens when a timeout occurs directly when a
```
fst-crenshaw, line 332: Much sentence.  Please shorten.  This feels like the really important part about "sternly".  It seems that this is the scenario in which a KillSwitch cannot guarantee that an instance will be killed.  I'm not feeling confident that I understand this scenario.   I *think* it's this:

- A hostcall faults.  Domain::Hostcall.
- At the very same moment, a KillSwitch is invoked.
- In some cases, execution_domain gets updated to Domain:Terminated.

But then what happens?  I'm not sure what it means when you write, "no further effect will be had."  Does that mean the KillSwitch has no further effect, but the instance will live out the remainder of its lifecycle from `Domain::Terminated`?

docs/src/lucet-runtime/killswitch.md
iximeow: Instances can exit only once and `terminable` is true only until one specific exit mechanism has been picked. "being terminable" really means "the instance's fate isn't sealed yet". Does that sound helpful?

docs/src/lucet-runtime/killswitch.md
iximeow: That's an entirely fair way to phrase it! I think `Domain` is more granular than we'd want `State` to be if it weren't for timeouts requiring it, but for timeout correctness `Domain` must be an accurate reflection of the Instance's state.

docs/src/lucet-runtime/killswitch.md
iximeow: Oops, this was poor phrasing. `Domain` only has the ability to express that an instance has stopped, not how. Specifically, `Domain` has `Domain::Terminated` but not `Domain::Faulted`.

I'll rephrase this to not suggest that `Domain` specifically expresses an instance has stopped, and instead emphasize that `Domain::Terminated` implies the instance has stopped, and that details for that termination are over on the instance's `State`.

docs/src/lucet-runtime/killswitch.md
iximeow: Agreed. Related only by suggestion text, I should strengthen the `some transitions reflecting a timeout` part of this sentence, since _all_ timeouts that happen without races should be described in the diagram (and are currently, afaik)

docs/src/lucet-runtime/killswitch.md
iximeow: Ah, you're right that this should be explicitly written here. There's a unit smaller than `Instance` that a `KillState` is associated with, and I don't have a great word for it, which is probably why I skirted around it. The lifecycle and `KillState` described are for one _execution_ of an `Instance` function (one call to `Instance::run`). `KillState` lives on `Instance` because only one execution can happen at a time. A single `Instance` could be terminated dozens of times, because the thing being terminated is the function being run _in_ the `Instance` being terminated.

docs/src/lucet-runtime/killswitch.md
iximeow: I was afraid that copying the descriptions on each header would be too repetitive, but looking again after a few days away.. it's definitely good for the reader :grin:

docs/src/lucet-runtime/killswitch.md
fst-crenshaw: Yes!  That helps my own understanding!

docs/src/lucet-runtime/killswitch.md
iximeow: I agree that "run" in two tenses was not the easiest read. I've fiddled with the wording here - I wanted to avoid referencing functions by name here just because that seems a very easy way for this kind of doc to bitrot, but `enter_guest_region` is already uttered later anyway!

docs/src/lucet-runtime/killswitch.md
iximeow: I like your point about which mechanism people will likely _want_ to see. I'll reorder these items.

(also thank you for calling out my run-ons. there are many.)

docs/src/lucet-runtime/killswitch.md
iximeow: `whether` is the word I was looking for! Thank you.

docs/src/lucet-runtime/states.dot
iximeow: My approach was to write states with the diagram showing permissible values in `KillState`, for the state being described. State E is really the catch-all of "what lucet_runtime sees right after `swap_and_run` returns". It can be reached from a hostcall exit (where Domain is `Hostcall`), from a guest returning (`Domain::Guest`), from running a cancelled guest (`Domain::Cancelled`), but really the `KillState` when reaching E no longer matters. The instance that any KillSwitch might terminate has just finished running, so they're no longer effectual, and we should be about to replace `kill_state` with a fresh new `KillState` for the next run anyway. This is the same as `Instance::run returns` over in `execution`'s [module-level docs](https://docs.rs/lucet-runtime-internals/0.6.1/lucet_runtime_internals/instance/execution/index.html), which is similarly weird in having three `execution_domain` states.

I was initially sketching this out with `execution_domain` and `terminable` as both `N/A` for state E, walked that back for being imprecise, but I'll switch back to `N/A` - that's more appropriate for readers here. Maintainer-facing documentation of what permissible states _should_ be is already in the module-level commentary, and we don't need to duplicate that.

docs/src/lucet-runtime/killswitch.md
iximeow: That's the right sequence of events for this section, yep! "no further effect will be had" refers to the fact that a hostcall fault happened, so one way or the other the instance will not resume, meaning the timeout will never be observed (which would require resuming and reaching `end_hostcall`)

I've expanded on that point and rephrased this for improved human-friendliness.

docs/src/lucet-runtime/killswitch.md
iximeow: In thinking this over, I realized my mental model for how someone would read this is kind of like a reference. Rather than reading top to bottom, I expect the reader to be looking for an answer to a specific "how is Lucet correct when ...". I think it's important to be legible top to bottom as well, but I think we'll have a different document that's ordered more amenably to "start reading here" cases. I'm not sure if that's actually the kind of readership this will have, either. Do you have an ordering in mind that satisfies both reading styles?

I'm curious about your thoughts maintaining a different ordering - it feels like it would be confusing to move one of these subsections around if it becomes less straightforward, or move one down if it's simplified in the future. Is that just me overthinking it?

fst-crenshaw: My own reading experience of it was educational, "I have no idea how any of this works."  That's quite different from your view of it as a reference.  I support whatever choice you make in the matter.

docs/src/lucet-runtime/killswitch.md
iximeow: Rephrased in favor of "timeouts in non-race scenarios"

docs/src/lucet-runtime/killswitch.md
iximeow: I mostly reworked this paragraph and it now just explicitly says `Domain` and `KillState` are not public, and where to find them in `lucet_runtime_internals`.

docs/src/lucet-runtime/killswitch.md
iximeow: I've fiddled with the wording here and teased apart some of the sentences I overloaded. I hope this is clearer, though I'm concerned the actual `Timeout while handling a guest fault` section is still a bit dense. If you have a moment later, I'd appreciate your thoughts with a re-read.

docs/src/lucet-runtime/killswitch.md
iximeow: I reordered the sections. If some of them get trickier or simpler, we can just move them - no biggie. :grin:

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,385 @@
...
+* `B -> C` (guest makes a hostcall)
+* `B -> E` (normal guest exit)
+* `B -> E` (from a guest fault/timeout)
+* `C -> B` (hostcall returns to guest)
+* `C -> E` (hostcall terminates instance)
+* `C -> E` (hostcall observes timeout)
+  - not an internal state but we will also discuss timeouts during a hostcall fault
+* `D -> E` (cancelled guest is run)
+
+These will be coved in rough order of complexity, starting with the simplest
```
fst-crenshaw, line 216: spelling nit: coved -> covered.

docs/src/lucet-runtime/states.dot
fst-crenshaw: Agreed.  Thanks.

docs/src/lucet-runtime/killswitch.md
```diff
@@ -0,0 +1,385 @@
...
+`State` describes the state of the instance as visible outside `lucet_runtime`.
+This is fundamentally the reason why `State` does not and cannot describe a
+"making a hostcall" variant - this is not something `lucet_runtime` wants to
+expose, and at the moment we don't intend to make any commitments about being
+able to query instance-internal state like this.  Additionally, this is why
+`Domain::Terminated` only expresses that an instance has _stopped_ running, not
+if it successfully or unsuccessfully exited. From `Domain`'s perspective, the
+instance has stopped running, and that's all there is to it. For more
+information, the `Instance`'s `State` will describe how the instance stopped,
+and if that's by a return or a fault. TODO: there's even an argument that
```
fst-crenshaw, line 110: I notice this TODO.

awortman-fastly: This is gone now. I actually think `Cancelled` vs `Terminated` is a rough spot in our model of termination, and wanted to say as much in this document, but wasn't sure how to phrase better than a TODO. @katie-martin-fastly understandably removed it in an edit and I'll just do a PR later this week to actually try the simplification.

PR #496, opened 2020-04-14
Fixes #452: Introduce a random slot allocation strategy to the Region.

### LINKED ISSUE #452: Add ability to randomize slot choices from a Region
We'd like to be able to have the Slots chosen from a Region be randomized. Currently we tend to end up reusing the same Slots frequently. As a security-in-depth measure, randomizing where Instances are allocated is useful.

fst-crenshaw:
This PR does two things:

1. Stays up to date with improvements over in https://github.com/bytecodealliance/lucet/pull/494 in which @acfoltzer refactored validate_runtime_spec() to be more accommodating.

2. Introduces an AllocStrategy for the Region.  When creating a new instance in a Region, one can choose to allocate the new instance using the original strategy (give me the next one you got) or randomly from the set of available slots.

**See original issue over at:** 
https://github.com/bytecodealliance/lucet/issues/452


lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -1,4 +1,6 @@
-use crate::alloc::{host_page_size, instance_heap_offset, Alloc, Limits, Slot};
+extern crate rand;
```
jedisct1, line 1: You could simply use `use rand` :)

jedisct1:
Looks great!

And this is a nice improvement, thanks!

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -80,38 +82,32 @@ impl RegionInternal for MmapRegion {
...
+                    .unwrap()
+                    .pop()
+                    .ok_or(Error::RegionFull(self.capacity))?
             }
-            Ordering::Equal => (),
-            Ordering::Greater => {
-                return Err(Error::InvalidArgument(
-                    "heap memory size requested for instance is larger than slot allows",
-                ))
+            _ => {
```
tyler, line 100: I think it would be preferable to use `AllocStrategy::Random` instead of `_`. That way when we add a new strategy, this gives you an informative error instead of defaulting to random.

fst-crenshaw: Agreed.  That is a bit of leftover cruft.  Thanks for catching this!

fst-crenshaw:
I believe I have a code-complete first draft of this work.  Reviews are appreciated as your input will go towards improving it.

lucet-runtime/lucet-runtime-internals/Cargo.toml
```diff
@@ -24,9 +24,12 @@ memoffset = "0.5.2"
 nix = "0.17"
 num-derive = "0.3.0"
 num-traits = "0.2"
+rand = "0.7"
+rand_core = "0.5"
 raw-cpuid = "6.0.0"
 thiserror = "1.0.4"
 tracing = "0.1.12"
+xorshift = "0.1"
```
jedisct1, line 32: `xorshift doesn't seem to be used anywhere?

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -80,38 +80,42 @@ impl RegionInternal for MmapRegion {
...
+                let mut free_slot_vector = self.freelist.write().unwrap();
+                if free_slot_vector.len() == 0 {
+                    return Err(Error::RegionFull(self.capacity));
+                }
+                // Instantiate a random number generator and get a random slot.
+                let mut rng = thread_rng();
+                let rnd_idx = rng.gen_range(0, free_slot_vector.len());
+                slot = free_slot_vector.swap_remove(rnd_idx);
+            }
+            AllocStrategy::CustomRandom(mut rng) => {
```
jedisct1, line 109: It feels a little bit weird to expose something only intended to be used internally in the public API.

Do the tests really need `CustomRandom`? Having non-deterministic input may actually be better to trigger edge cases that a deterministic sequence would never trigger, but real-world usage would.

lucet-runtime/lucet-runtime-internals/Cargo.toml
fst-crenshaw: Thanks for catching this.  I forgot to push one last time.

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
fst-crenshaw: I see your point.  So much so that I began writing the tests non-deterministically, as you see here:  https://github.com/bytecodealliance/lucet/pull/496/commits/421578f2730a7b9c4082bc8f02e912eee21c5330.  😊  I agree that non-deterministic input may trigger issues that could happen in real-world use.  

In talking with @acfoltzer about crafting these tests, he expressed a preference for them to be deterministic.  This approach has its own advantage.  He expressed a worry about having tests in our repo that could break 1 time out of 100, which is a difficult error situation to debug.

Acknowledging the many times that I have endured flakey tests in other repos I have worked on, my own preference is to have non-deterministic tests for this scenario.  But it's only a slight preference and I acknowledge that I do not know this repo as well as others on the team.  It would be great to hear more conversation on this matter.

pchickey: In wiggle (wasmtime repo) we use `proptest`s in CI. When those tests fail, they print out the seed and instructions to use that seed to reproduce locally. Can we use that approach, either re-creating it or better yet just using proptest's machinery?

acfoltzer: It should be possible to do the same here; use `thread_rng()` to create a random seed, print it, and then seed the `StdGen` that the test uses. I really do not want to have to rely on getting lucky in order to reproduce breakages.

fst-crenshaw: Thanks for the discussion.  Here's what I'm hearing:
- It is undesirable to rely on lucky seeds.
- It is undesirable to rely on luck to reproduce breaks.
- Let's keep the `CustomRandom` enum in the `AllocStrategy`.
- Use `thread_rng()` to create a random seed and print it so that any breaking seeds are immediately known at the time of failure.

Do I have that right?

lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
```diff
@@ -2,9 +2,11 @@
 macro_rules! alloc_tests {
     ( $TestRegion:path ) => {
         use libc::c_void;
+        use rand::rngs::StdRng;
+        use rand_core::SeedableRng;
```
acfoltzer, line 6:
```suggestion
        use rand::SeedableRng;
```
This is reëxported by `rand`, so you don't need the explicit `rand_core` dependency

lucet-runtime/lucet-runtime-internals/src/alloc/tests.rs
```diff
@@ -774,6 +776,162 @@ macro_rules! alloc_tests {
...
+                let region = TestRegion::create(total_slots, &LIMITS).expect("region created");
+                assert_eq!(region.capacity(), total_slots);
+                assert_eq!(region.free_slots(), 10);
+                assert_eq!(region.used_slots(), 0);
+
+                // Randomly allocate all of the slots in the region.
+                for i in 1..=total_slots {
+                    let inst = region
+                        .new_instance_builder(module.clone())
+                        .with_alloc_strategy(AllocStrategy::CustomRandom(Box::new(rng.clone())))
```
acfoltzer, line 830: I think you'll need to thread the same `rng` through to all of these invocations, otherwise each RNG invocation is just one step after the initial seeding on L813. This is why I was thinking we might need `Arc<Mutex<dyn RngCore>>` rather than `Box<dyn RngCore>`

lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
```diff
@@ -80,38 +80,42 @@ impl RegionInternal for MmapRegion {
...
-                    heap_memory_size: heap_memory_size_limit,
-                    ..*limits
-                };
-                limits = &custom_limits;
+        let limits = self.get_limits();
+        module.validate_runtime_spec(&limits, heap_memory_size_limit)?;
+
+        let slot;
+
+        match alloc_strategy {
```
acfoltzer, line 90: Could you abstract this into a method on `AllocStrategy`, so that the logic could be reused with `UffdRegion`? It seems like you could get away with something like:

```rust
impl AllocStrategy {
    fn next(&mut self, free_slots: usize) -> usize { ... }
}
```

Then the region could handle its own `swap_remove` logic, etc.

fst-crenshaw: Neat.  I'll give this a try.

fst-crenshaw: One small worry.  What if `free_slots` is 0?  I'm tempted to return a `Result` instead of a usize.  What are your thoughts?

acfoltzer: Good idea!

fst-crenshaw:
Thanks @jedisct1.  You make a great point about secure defaults.  I'd love to make Random the default strategy, but I'd like to do some integration testing in a few places first.  I'd also like to submit the "Random is a Default" work as a separate stand-alone PR; it would be one of those big fiddly PRs where a whole bunch of tests would need to be slightly altered and not much else happens.

jedisct1:
Makes sense! Thanks Tanya!

PR #497, opened 2020-04-16
KillSwitch race testing

iximeow:
This moves testing of `KillSwitch` out of `lucet-runtime-tests` and into a new crate, `lucet-concurrency-tests`. It also adds instrumentation into `lucet-runtime`, conditionally present on a new `concurrent_testpoints` crate flag, to force specific orderings in execution. This instrumentation lets us replace some OS-scheduler-dependent timing fickleness with strict happens-before guarantees, which should fix the flakiness of `double_timeout` in Mac CI.


lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,731 @@
...
+        }
+
+        unsafe {
+            let heap_base = lucet_vmctx_get_heap(vmctx);
+
+            // Using the default limits, each instance as of this writing takes up 0x200026000 bytes
+            // worth of virtual address space. We want to access a point beyond all the instances,
+            // so that memory is unmapped. We assume no more than 16 instances are mapped
+            // concurrently. This may change as the library, test configuration, linker, phase of
+            // moon, etc change, but for now it works.
```
katie-martin-fastly, line 67: :new_moon: :crescent_moon: :full_moon:

wonderful comments, as always.

iximeow: this is an @acfoltzer special! Much of this is a straight copy of `lucet-runtime-tests`' `guest_fault`, with the timeout mock module mixed in. I'm actually on the fence about keeping several of these hostcalls because there's a fair bit not used in these tests.

acfoltzer: This is actually adapted from a comment in the original C runtime's tests, so probably a @pchickey special originally! :cowboy_hat_face:

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,731 @@
...
+pub fn test_c_with_instrumented_guest_entry<F, R>(dir: &str, cfile: &str, f: F) -> R
+where
+    F: FnOnce(InstanceHandle) -> R,
+{
+    test_ex(|| {
+        unsafe {
+            ENTERING_GUEST = Some(Syncpoint::new());
+        }
+        let module = test_module_c(dir, cfile).expect("build and load module");
+        let region = MmapRegion::create(1, &Limits::default()).expect("region can be created");
```
fst-crenshaw, line 163: I have worried feelings about your using the MmapRegion directly.  I don't comprehend how big an ask this is, but can this work for generic `Regions`?

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,731 @@
...
+pub fn yielding_hostcall(vmctx: &mut Vmctx) {
+    vmctx.yield_();
+}
+
+pub fn test_c_with_instrumented_guest_entry<F, R>(dir: &str, cfile: &str, f: F) -> R
+where
+    F: FnOnce(InstanceHandle) -> R,
+{
+    test_ex(|| {
+        unsafe {
```
fst-crenshaw, line 159: Asking from a place of curiosity and wanting to learn more, why is this `unsafe`?

pchickey: Modifying a `static mut` is `unsafe` because there is no language-provided protections against race conditions. https://doc.rust-lang.org/reference/items/static-items.html#mutable-statics

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,731 @@
...
+/// and asserting that it returned the expected result.
+fn run_onetwothree(inst: &mut Instance) {
+    let retval = inst
+        .run("onetwothree", &[])
+        .expect("instance runs")
+        .unwrap_returned();
+    assert_eq!(libc::c_int::from(retval), 123);
+}
+
+pub fn mock_killswitch_module() -> Arc<dyn Module> {
```
fst-crenshaw, line 31: It feels like the purpose of this module is to provide a function that represents every step of the KillSwitch lifecycle.  Do I have that right?

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,731 @@
...
+
+        let inst = region
+            .new_instance(module)
+            .expect("instance can be created");
+
+        f(inst)
+    })
+}
+
+pub fn test_exclusive_instance_with_instrumented_guest_entry<F, R>(f: F) -> R
```
fst-crenshaw, line 173: I am curious about the difference between an **instance** and an **exclusive instance**.

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,731 @@
...
+            .expect("instance can be created");
+
+        f(inst)
+    })
+}
+
+// Test that a timeout that occurs in a signal handler is handled cleanly without signalling the
+// Lucet embedder.
+#[test]
+fn terminate_in_guest() {
```
fst-crenshaw, line 211: Does this case map to the docs description of "Terminated while handling guest fault?"

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,731 @@
...
+            racepoint.wait_and_then(|| {
+                kill_switch.terminate().expect("can terminate in guest");
+            });
+
+            guest.join().expect("guest exits without panic");
+        })
+    }
+}
+
+// Test a termination that completes right before `exit_guest_region` takes ownership of termination.
```
fst-crenshaw, line 291: I'm realizing that what would help me with the clarity of these tests is if the brief header comments described the test cases using the same terms or language as the docs.  I believe this the case where the KillSwitch is in a race with teardown and KillSwitch acquires terminable.  I don't feel confident in that assessment.

fst-crenshaw: Nope.  I'm wrong.  I think that's the test case on line 352.  Maybe?

lucet-runtime/lucet-runtime-internals/src/lock_testpoints.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        let res = f();
+
+        self.proceed.store(true, Ordering::SeqCst);
+
+        res
+    }
+}
+
+pub struct Syncpoint {
```
katie-martin-fastly, line 32: Do you think you could add some `///` comments briefly explaining what a sync point is, and how we use this in tests?

lucet-runtime/lucet-runtime-internals/src/lock_testpoints.rs
```diff
@@ -0,0 +1,117 @@
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+
+pub struct SyncWaiter {
```
katie-martin-fastly, line 5: Ditto here, would be really nice to have a brief explanation of this type.

lucet-runtime/lucet-runtime-internals/src/lock_testpoints.rs
katie-martin-fastly: We could link to the docs you wrote in #495 btw, I don't think this has to go into the details of `KillSwitch` so much as elaborating on how this helps force certain race conditions

lucet-concurrency-tests/tests/killswitch.rs
iximeow: this is more a 'test' and 'test exclusive' delineation to control process-wide state - this is taken from `lucet-runtime-tests` like much else, where it's to control what signal handler is installed there. Here, it's also necessary to ensure use of `ENTERING_GUEST` doesn't happen in an unsafe way (with multiple threads clobbering the shared global).

This deserves a note here and on `test_instance_with_...` too :D

lucet-concurrency-tests/tests/killswitch.rs
iximeow: Yep! Additionally, functions to distinguish cases we want to test even though they are the same with respect to the KillSwitch lifecycle: this exposes a hostcall that does nothing, and a hostcall that yields so we can test both, even though they're both just hostcalls to `KillSwitch`.

lucet-concurrency-tests/tests/killswitch.rs
iximeow: This should be read as `test_...` and `test_exclusive..`, rather than `exclusive` being a modifier on "instance". I've added some documentation on these helpers to clarify, but just for the viewers at home: some tests require access to process-wide resources (signal handler, the `ENTERING_GUEST` syncpoint), so to guarantee the state and safety of those a test will need a guarantee of exclusive access.

`test_ex` and `test_nonex` in `lucet-runtime-tests` give this same building block, but `lucet-concurrency-tests` adds a new item to manage with `ENTERING_GUEST` and these are convenient points to move out repeated instance setup.

lucet-concurrency-tests/tests/killswitch.rs
iximeow: The comment here was a lie :grimacing: There isn't a specific "while handling guest fault" test here, yet! That's the next PR as soon as this lands :)

lucet-concurrency-tests/tests/killswitch.rs
iximeow: "link these tests with cases discussed in the documentation" is a good idea, and I've done just that.

lucet-concurrency-tests/tests/killswitch.rs
iximeow: (this is done now! [fec7714](https://github.com/bytecodealliance/lucet/pull/497/commits/fec77141bf19f840d56b9e7518cad9f980ab84d7))

lucet-concurrency-tests/src/killswitch.rs
```diff
@@ -0,0 +1,806 @@
+//! `KillSwitch` has documentation about the correctness of its various edge cases as part of
+//! Lucet's wider documentation, see this section of the book for more:
+//! https://github.com/bytecodealliance/lucet/blob/master/docs/src/lucet-runtime/killswitch.md#implementation-complexities-when-you-have-a-scheduler-full-of-demons
+//!
+//! This module contains actual test cases exercising these conditions where possible. Phrasing
+//! from the above-mentioned section are re-used below in describing test cases. New tests should
+//! have corresponding cases in the Lucet `KillSwitch` chapter!
```
fst-crenshaw, line 7: Oooh.  A lovely mandate.

lucet-concurrency-tests/src/killswitch.rs
```diff
@@ -0,0 +1,806 @@
...
+use lucet_runtime::vmctx::Vmctx;
+use lucet_runtime::{
+    // lucet_hostcall, Error,
```
fst-crenshaw, line 10: Perhaps a bit of leftover cruft?

lucet-concurrency-tests/src/killswitch.rs
```diff
@@ -0,0 +1,806 @@
...
+    extern "C" fn hit_sigstack_guard_page(vmctx: *mut lucet_vmctx) {
+        extern "C" {
+            fn lucet_vmctx_get_globals(vmctx: *mut lucet_vmctx) -> *mut u8;
+        }
+
+        unsafe {
+            let globals_base = lucet_vmctx_get_globals(vmctx);
+
+            // Using the default limits, the globals are a page; try to write just off the end
+            *globals_base.offset(0x1000) = 0;
```
fst-crenshaw, line 95: I wish I had mentioned this before, but I missed it.  I do apologize.

I'm curious about where this default 0x1000 lives.  I'm not seeing it over at: https://github.com/bytecodealliance/lucet/blob/83fd2248b36c780cc74c0abb49024cca4a0c053c/lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs#L461

fst-crenshaw: Also.  Hex is not decimal.  Yep.  Thanks peer shame.

lucet-concurrency-tests/src/killswitch.rs
iximeow: heck. yep, I was trying to pare down which imports are necessary and forgot to clean up :(

lucet-concurrency-tests/src/killswitch.rs
fst-crenshaw: https://www.youtube.com/watch?v=OjYoNL4g5Vg

lucet-concurrency-tests/src/killswitch.rs
```diff
@@ -0,0 +1,806 @@
...
+                    });
+
+                    guest.join().expect("guest exits without panic");
+                })
+            }
+        }
+
+        // Test a termination that completes right before `exit_guest_region` takes ownership of termination.
+        //
+        // This corresponds to a race during the documentation's State B -> State E "due to normal exit"
```
fst-crenshaw, line 341: 😂 Thanks for mirroring your docs.  This helps me immensely with navigating all the test cases.

lucet-concurrency-tests/tests/killswitch.rs
```diff
@@ -0,0 +1,3 @@
+use lucet_concurrency_tests::killswitch_tests;
+
+killswitch_tests!(lucet_runtime::MmapRegion);
```
fst-crenshaw, line 3: How best to represent the work of doing this same line for UserfaultFd?  Perhaps an issue?  Perhaps a comment in https://github.com/bytecodealliance/lucet/pull/492?

fst-crenshaw:
I see there's a merge conflict.  I'm happy to do a quick reapproval when you are ready.

lucet-concurrency-tests/tests/killswitch.rs
iximeow: - [ ] If this lands before #492, add a comment there to copy `killswitch_tests!()`

:grin:

PR #498, opened 2020-04-17
Fix hostcalls when building in release mode

pchickey:
As part of the upgrade to rust 1.42, executables which export hostcalls through the usual methods (e.g. `lucet_wasi::export_wasi_funcs`) apparently broke, but only when compiling in release mode.

Adam determined that changing the Rc leak of the hostcall function pointers to `assert_ne!(fptr, null)` fixed the issue. Previously, I noticed that enabling LTO fixed the issue as well, but the assert_ne fix seems a lot more likely to keep working across changes to rustc and llvm.

```sh
cargo build --release -p lucetc -p lucet-wasi
./target/release/lucetc lucet-wasi/tests/guests/duplicate_import.wat --bindings lucet-wasi/bindings.json
./target/release/lucet-wasi a.out
```

Prior to this patch, `target/release/lucet-wasi` will fail to dlopen the module because the wasi hostcall symbols are missing, and `nm` of the executable confirms this. After this patch, the symbols are present in the release executable again, and `nm` confirms it.

Our CI doesn't check that release mode executables work correctly. That should be corrected in a follow-up PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,3 @@
-# Lucet version 0.3.1
-
 [workspace]
 members = [
   "benchmarks/lucet-benchmarks",
@@ -25,6 +23,3 @@ exclude = ["wasmtime"]
 
 [profile.test]
 rpath = true
-
-[profile.release]
-lto = true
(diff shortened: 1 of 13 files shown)


pchickey:
LTO turned out to not be the correct fix. @acfoltzer discovered that replacing the Rc leak of the ptr with `assert_ne!(fptr, null())` fixes it with and without LTO.

PR #500, opened 2020-04-17
factor makefile & ci jobs to run more parallel jobs

pchickey:
we have run out of disk space on github actions. We're building an awful
lot of suff serially, so lets split them into many jobs so that each job
has some headroom.

* test-ci should just do a `profile = test` build. I'm not 100%
confident I got that right but lets run it and see.

* smoke testing benchmarks does a `profile = bench` build, so
  we can move that to its own makefile target and job.

* smoke testing fuzz does a `profile = debug` build, so
  we can move that to its own makefile target and job.

* cargo audit doesnt need to be in docker container. it can be its own
job in parallel. Just need stable rust and cargo-audit installed.

* docs build doesnt need to be in docker container. it can be its own
job in parallel. Just need stable rust and mdbook

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,41 @@ jobs:
     # This action builds the container and executes the test suite inside it.
     - uses: ./.github/actions/test
       with:
-        target: test-full
+        target: test-ci
+
+    - name: Ensure testing did not change sources
+      run: git diff --exit-code
+
+  smoke_test_benchmarks:
+    name: Smoke-test benchmarks
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: 'recursive'
+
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -14,7 +14,41 @@ jobs:
...
+    - uses: ./.github/actions/test
+      with:
+        target: test-benchmarks
+
+    - name: Ensure testing did not change sources
+      run: git diff --exit-code
+
+  smoke_test_fuzz:
+    name: Smoke-test fuzz
+    runs-on: ubuntu-latest
```
acfoltzer, line 41: This eventually gets run in our `bionic` container, right?

pchickey: Yes. The `uses: ./.github/actions/test` step builds (without a cache, grr) and runs the container. The `runs-on` is just the host of the docker environment, afaict.

PR #501, opened 2020-04-18
Add release mode testing to CI

pchickey:
Adds makefile targets and CI jobs to test in `--release` mode, and fixes all tests to work in that mode.

Optimizations run in release builds which often lead to dead-code-elimination or link-time garbage collection opportunities, and change how our runtime dynamic linking ends up working. This testing should prevent further regressions like we had when upgrading to rust 1.42.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 export GUEST_MODULE_PREFIX:=$(abspath .)
 
+CRATES_NOT_TESTED = lucet-spectest lucet-benchmarks lucet-runtime-example
+
 .PHONY: build-dev
 build-dev:
 	@echo Creating a DEBUG build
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -25,20 +27,23 @@ test: indent-check test-packages
...
-            -p lucet-runtime-internals \
-            -p lucet-runtime \
-            -p lucet-module \
-            -p lucetc \
-            -p lucet-wasi-sdk \
-            -p lucet-wasi \
-            -p lucet-wasi-fuzz \
-            -p lucet-validate \
-            -p lucet-wiggle
+	cargo test --no-fail-fast --all $(CRATES_NOT_TESTED:%=--exclude %)
```
fst-crenshaw, line 30: I very much like this exclusion approach.

PR #504, opened 2020-04-22
update deps

pchickey:
Delete some old dep pins and run `cargo update` so that lucetc is closer to compiling for wasm32-wasi.

This regeneration of Cargo.lock also moves us to the new format that is supposed to tolerate git merges better.


lucet-runtime/lucet-runtime-internals/Cargo.toml
```diff
@@ -17,24 +17,17 @@ anyhow = "1.0"
...
-memoffset = "0.5.2"
+memoffset = "0.5.3"
 nix = "0.17"
 num-derive = "0.3.0"
 num-traits = "0.2"
 raw-cpuid = "6.0.0"
 thiserror = "1.0.4"
 tracing = "0.1.12"
 
-# This is only a dependency to ensure that other crates don't pick a newer version as a transitive
```
iximeow, line 31: :tada:

PR #506, opened 2020-04-24
🎬 Add explicit API to run start functions; don't run it implicitly

acfoltzer:
Previously, the start function would be implicitly run upon instance creation or resetting. Since the environment can be different at instantiation time to when export functions are typically called (missing embedder contexts, etc), this has tripped people up. It can also just be surprising that any code runs at all at those points, which combined with the inherently confusing nature of Wasm start functions just led to too many messes.

`Instance::run_start()` is now part of the public API, and will run the start function if it is present in that instance's Wasm module. It does nothing if there is no start function. `Instance::run()` will now return `Err(Error::InstanceNeedsStart)` if the start function is present but hasn't been run since the instance was created or reset.

`Instance::run_start()` will now return an error if run on an instance more than once without an intervening `Instance::reset()`.

If a start function attempts to call a `#[lucet_hostcall]`-decorated import function, it will be terminated.

This commit bumps the development version to `0.7.0-dev`, because while the API changes are semver-compatible, the semantics are not.


lucet-module/src/functions.rs
```diff
@@ -125,9 +125,15 @@ impl OwnedFunctionMetadata {
...
 
+#[derive(Clone, Copy, Debug)]
 pub struct FunctionHandle {
     pub ptr: FunctionPointer,
     pub id: FunctionIndex,
+    /// This is `true` if and only if this handle was referenced through the start section.
+    ///
+    /// The same function may be referenced via other contexts, and will appear with `false` in this
+    /// field in those cases.
+    pub is_start_func: bool,
```
acfoltzer, line 136: This is a little weird; I could be convinced this belongs in `lucet-runtime-internals` only, but since `FunctionHandle` is only used downstream of `lucet-module` it seemed okay.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -530,9 +525,43 @@ impl Instance {
...
+    /// as long as the start function does not attempt to use any imported functions.
+    ///
+    /// # Safety
+    ///
+    /// The foreign code safety caveat of [`Instance::run()`][run]
+    /// applies.
+    ///
+    /// [run]: struct.Instance.html#method.run
+    /// [start]: https://webassembly.github.io/spec/core/syntax/modules.html#syntax-start
+    pub fn run_start(&mut self) -> Result<(), Error> {
```
iximeow, line 551: should `run_start` return an error if it has already been run and the instance not reset? I don't think it's terribly likely to have been calling the start function manually before, so accidentally rerunning it was unlikely, but double-initialization seems more likely with this as public API.

acfoltzer: Yeah, I actually was just thinking this. Good call

sunfishcode: Not a blocking concern, just a question that I'm curious about: would it be feasible to return an error if any exports are called on an instance before its start function has run?

acfoltzer: Done in 21a8ff3 :sparkles:

acfoltzer: > Not a blocking concern, just a question that I'm curious about: would it be feasible to return an error if any exports are called on an instance before its start function has run?

Aaa, race condition on the comments, sorry. This is a good idea; I will add it while my brain is in this corner of the project.

acfoltzer: @sunfishcode your suggestion is implemented in 50f808f. Thank you!

acfoltzer:
There is now a fair amount of whitespace diff in the `start` test suite. When reviewing, I recommend filtering that out:

image

lucet-runtime/lucet-runtime-internals/src/instance.rs
sunfishcode: If I read the patch right, it's issuing an error if the start function calls any imports. My question was about issuing errors if any exported function in a module is called before the module's start function is called, which isn't normally possible.

acfoltzer: Well, I definitely read that wrong, whoops. Yes, an error is returned if you try to call any other entrypoints before running the start function, if it exists.

Should I keep the terminate-on-import-function behavior? Now that I'm looking at the spec again I can't actually figure out whether import functions can be called in the start section.

sunfishcode: Imports can be called from the start function.

acfoltzer: Where's my dunce cap emoji? Reverted in 1e84d31

PR #512, opened 2020-05-06
Allow use of pinned heap registers

shravanrn:
@pchickey @awortman-fastly @iximeow Slightly cleaned up and rebased version of https://github.com/bytecodealliance/lucet/pull/273 with some comment fixes

I've left the code mostly as is although some of the changes here seem orthogonal. Wanted to ask about this first before removing the code
- Context save for sandbox is also saving the value for register r8
- In stack.rs, there is a stack overflow test where the size of the stack seems to be bumped.

Let me know if I should remove either of these, and I can update the PR

### DIFF
--- a/lucet-module/src/module_data.rs
+++ b/lucet-module/src/module_data.rs
@@ -59,6 +59,8 @@ pub struct ModuleFeatures {
     pub lzcnt: bool,
     pub popcnt: bool,
     pub instruction_count: bool,
+    pub pinned_heap: bool,
+    pub pinned_heap_register: u16,
     _hidden: (),
 }
 
@@ -75,6 +77,8 @@ impl ModuleFeatures {
             lzcnt: false,
             popcnt: false,
             instruction_count: false,
+            pinned_heap: false,
+            pinned_heap_register: 0,
             _hidden: (),
         }
     }
--- a/benchmarks/lucet-benchmarks/src/context.rs
+++ b/benchmarks/lucet-benchmarks/src/context.rs
@@ -9,7 +9,8 @@ fn context_init(c: &mut Criterion) {
 
     c.bench_function("context_init", move |b| {
         b.iter(|| {
-            ContextHandle::create_and_init(&mut *stack, f as usize, &[]).unwrap();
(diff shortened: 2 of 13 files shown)


shravanrn:
Sorry, just wrapping some other work. Will update the or shortly

shravanrn:
Made the changes. Hopefully the r15 bit is not too hacky --- i was trying to avoid pulling in new dependencies.
Also, restored the stack size in the test. Haven't tested this though. Waiting for CI to complete, and will fix things

shravanrn:
@iximeow looks like the tests pass with the old stack size. I think this is ready for review?

shravanrn:
ping :slightly_smiling_face:

lucet-module/src/module_data.rs
```diff
@@ -59,6 +59,8 @@ pub struct ModuleFeatures {
     pub lzcnt: bool,
     pub popcnt: bool,
     pub instruction_count: bool,
+    pub pinned_heap: bool,
```
iximeow, line 62: How's it sound to make this an `Option<HeapPinStyle>`? Then `None` means "no pinning", and `HeapPinStyle` can be written like
```
#[repr(u8)]
enum HeapPinStyle {
    PinR15 = 1
}
```
that we can extend if need be in the future?

I mostly want to avoid a cranelift `RegUnit`, even in `u16` form in `lucet-module` - the stability of those numbers isn't guaranteed, so we could misinterpret an old module, or an old runtime could misinterpret a new module.

pchickey:

iximeow:
Just a heads up, I pushed a small commit [here](https://github.com/bytecodealliance/lucet/commit/0fa4484dd89428a23986a6b22967fe6c6dccdabf) that specifically lists what heap pinning approach lucetc builds a module expecting. With that (or a change like it) I'd be happy to :heavy_check_mark: but I can't push to this branch so either you can cherry pick it or I can rebase these two and put up a new PR. Either works!

We've enabled CircleCI for some of our CI workflow since this branch got put up, so either way it'll need to pick up that configuration to get a pass on CI now.

PR #514, opened 2020-05-08
move `host_page_size` to sysdeps

froydnj:
I have this quixotic idea that most OS/arch-specific stuff should be able to live in `lucet_runtime_internals::sysdeps`; this patch is a first step along that path.  Ideally, adding Windows support can be boiled down to figuring out how to expose the right interfaces from `sysdeps` that everybody can implement.  Signals make this very tricky, among other things, but perhaps since Firefox doesn't care about signals + Lucet, the signal support for Windows can somehow be punted down the road.

Happy to hear thoughts about what might be a better path to structuring the codebase for Windows support.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
@@ -1,37 +1,11 @@
 use crate::error::Error;
 use crate::module::Module;
 use crate::region::RegionInternal;
+use crate::sysdeps::host_page_size;
 use libc::c_void;
 use lucet_module::GlobalValue;
-use nix::unistd::{sysconf, SysconfVar};
 use rand::{thread_rng, Rng, RngCore};
-use std::sync::{Arc, Mutex, Once, Weak};
-
-pub const HOST_PAGE_SIZE_EXPECTED: usize = 4096;
-static mut HOST_PAGE_SIZE: usize = 0;
-static HOST_PAGE_SIZE_INIT: Once = Once::new();
-
-/// Our host is Linux x86_64, which should always use a 4K page.
-///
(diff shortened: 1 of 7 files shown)


iximeow:
just nothing the CI failure on master after merging this commit:
how dare! mac CI failed

this looks like a transient network issue, from looking at the logs:
```
2020-05-11T19:12:40.6580090Z info: syncing channel updates for 'stable-x86_64-apple-darwin'
2020-05-11T19:13:10.6593780Z error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/Users/runner/.rustup/tmp/vgmrlcpzfdoxx_mx_file'
2020-05-11T19:13:10.6597480Z info: checking for self-updates
2020-05-11T19:13:40.6927990Z error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/rustup-update1Ao2i3/release-stable.toml'
2020-05-11T19:13:40.6928260Z error: caused by: failed to make network request
2020-05-11T19:13:40.6928950Z error: caused by: https://static.rust-lang.org/rustup/release-stable.toml: timed out
```

I'll rerun the job, and it'll probably pass, so not a big deal.

froydnj:
Thanks for the reviews, everyone!

> I hadn't thought about this before in reading this code, but do we blow up on systems with hugepages?

Yes, the check that we get back `HOST_PAGE_SIZE_EXPECTED` will blow things up on hugepage machines, or in some future world where we run on a machine with multiple supported page sizes (e.g. AArch64).  Adjusting software to runtime-modifiable page sizes is hard. :(

PR #515, opened 2020-05-08
Allow flexible memory management using Uffd

tyler:
This adds a UffdStrategy trait to allow users of UffdRegion to configure the way faults are handled.

This will be important especially to Lucet users who have highly concurrent applications and potentially large linear memories. Uffd in its most naive form is likely to be slower than the Mmap region because of the additional work done to shunt faults back into userspace, and the reaction to those faults over to the kernel.

This can be dealt with by being smarter about how those faults are handled. For instance, work can be batched up. Instead of copying one host page per fault, one could copy a batch of them. This doesn't implement that, but does provide the facilities to make it possible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -853,6 +853,8 @@ dependencies = [
 [[package]]
 name = "linux-version"
 version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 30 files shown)


tyler:
Cherry picked the relevant bits over to #492

PR #523, opened 2020-05-14
lucet-wiggle: only borrow heap inside GuestMemory::base

pchickey:
Changes Wiggle's borrow of vmctx's heap to be just at the use-site of a GuestPtr.

The GuestMemory::base function, which is only called by GuestMemory::validate_size_align, is in turn used by `GuestPtr::as_raw`, `GuestType::read`, and `GuestType::write`. The latter two only keep the resulting pointer for the lifetime of their stack frame. GuestPtr::as_raw does keep the resulting pointer alive as part of the `*mut [T]` returned, so that borrow *could* possibly be alive across the yield, though the current safety properties in the docs explicitly say it cannot be.

To resolve that safety hole, we need to move to a whole-hostcall GuestBorrows as we discussed on the call with Alex on monday. Will work on that as a subsequent PR upstream. For now, this means we can yield in wiggle without an unsafe `as_mut` on vmctx, which is brilliant.

### DIFF
--- a/lucet-wiggle/src/lib.rs
+++ b/lucet-wiggle/src/lib.rs
@@ -11,25 +11,23 @@ pub mod generate {
 
 pub mod runtime {
     use lucet_runtime::vmctx::Vmctx;
-    use std::cell::RefMut;
     use wiggle::GuestMemory;
 
     pub struct LucetMemory<'a> {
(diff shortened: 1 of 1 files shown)


acfoltzer:
Thank you!

PR #525, opened 2020-05-18
🎬 Encode the start function in ModuleData rather than the linker

acfoltzer:
This long-overdue refactoring moves the metadata required to get a module's Wasm start function to the `lucet-module` types, rather than encoding it directly as a pointer-to-a-pointer in the shared object's symbols.

Of particular benefit is the fact that once a `DlModule` is created, we no longer use the underlying `libloading::Library`. Before the version bump in this patch, that library had a mutex around calls to `dlerror` to try preventing races between all threads in the process, which could lead to contention when threads attempted to concurrently use Lucet instances.


lucetc/src/compiler.rs
```diff
@@ -516,32 +515,6 @@ fn write_module_data<B: ClifBackend>(
...
 
-fn write_startfunc_data<B: ClifBackend>(
-    clif_module: &mut ClifModule<B>,
-    decls: &ModuleDecls<'_>,
-) -> Result<(), Error> {
-    use cranelift_module::{DataContext, Linkage};
-
-    if let Some(func_ix) = decls.get_start_func() {
-        let name = clif_module
-            .declare_data("guest_start", Linkage::Export, false, false, None)
```
iximeow, line 527: getting rid of a symbol a day keeps the dynamic linker away :sparkles:

acfoltzer:
We could definitely just use `libc` to call libdl functions directly, but I would probably want to write wrappers around them that would turn out very very similar to what `libloading` provides. But eventually just getting rid of it entirely would be :100:

PR #531, opened 2020-05-20
Point to wasmtime 0.16.0 / cranelift 0.63.0 release

pchickey:
No code changes!

If we want to do a crates.io release soon, we should do so after merging this PR and before merging any other updates to the `wasmtime` submodule.


pchickey:
Wasmtime 0.16 has a wasi-common regression, we'll probably be able to ask for a 0.17 soon once thats fixed, so lets skip this one and not try to do a crates.io release based on 16

PR #535, opened 2020-05-21
Update the Dockerfile

### LINKED ISSUE #533: Hello World example fails
I'm not sure if this is a documentation bug or an actual bug--I'm new to this project and can't quite figure that out.

I'm trying to run the [hello world example from the docs](https://bytecodealliance.github.io/lucet/Your-first-Lucet-application.html), with the dockerfile from the docs, but on the object building phase I get the error:
```
root@da558b6501ef:~/hello# lucetc-wasi -o hello.so hello.wasm
Error: Lucet Module: Unknown module for symbol `wasi_unstable::fd_prestat_get
```

I expected this line to successfully complete and produce `hello.so`. It did not.

Steps to repro:
1. Follow the [compiling Lucet from scratch guide (with docker)](https://bytecodealliance.github.io/lucet/Compiling.html). For the rest of this, I'll be inside the docker image.
  _Side note_: to anyone else who encounters this, the dockerfile needs a `CMD /bin/bash` or something at the end, or else it won't run.
3. `cd /root/`
2. `git clone https://github.com/bytecodealliance/lucet.git` (Latest on master at time-of-filing is:  12ae2a0cec4a7726d5d023b9ac227af336c59e7d)
1. `git submodule update --init --recursive`: Get all the submodules, as requested by the [compiling for linux guide](https://bytecodealliance.github.io/lucet/Compiling-on-Linux.html).
1. `make install`: Build lucet
2. `source /opt/lucet/bin/setenv.sh`: This is needed to put the tools below in `$PATH`
1.  `mkdir /root/hello/ && cd /root/hello`
1.  `printf '#include <stdio.h>\nint main(void){ puts("Hello World"); return 0;}' > hello.c`
1.  `wasm32-wasi-clang -Ofast -o hello.wasm hello.c`
1. `lucetc-wasi -o hello.so hello.wasm`
1. *Error*:
  ```
~/hello# lucetc-wasi -o hello.so hello.wasm
Error: Lucet Module: Unknown module for symbol `wasi_unstable::fd_prestat_get
```

Let me know what other information might be helpful!

I would love to be just a documentation bug, so if there's some extra modules I need to put into my path I'd love to know what I need to do.

Thanks for taking the time to look at this!

### PR DESCRIPTION
**Pull request by jedisct1**
- Use the current Ubuntu LTS version
- Install and use the default LLVM/Clang versions. clang 6 is very old.
- Add wabt
- Update binaryen
- Update the install script to look for LLVM in /usr/lib/llvm-* by default
- Update the documentation. People should not install `wasi-sdk` version 8 any more, but version 10.

This also fixes #533

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:20.04
 
 # This env variable makes sure installing the tzdata package doesn't hang in prompt
 ENV DEBIAN_FRONTEND=noninteractive
@@ -12,7 +12,7 @@ RUN apt-get update \
 	git \
 	libbsd-dev \
 	doxygen \
-	python-sphinx \
+	sphinx-doc \
 	cmake \
 	ninja-build \
 	ca-certificates \
@@ -23,12 +23,13 @@ RUN apt-get update \
 	libcsmith-dev \
 	creduce \
 	gcc-multilib \
-	clang-6.0 \
-	llvm-6.0 \
+	clang \
+	llvm \
+	lld \
+	wabt \
 	&& rm -rf /var/lib/apt/lists/*
 
-RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-6.0 100
-RUN update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-6.0 100
+RUN update-alternatives --install /usr/bin/wasm-ld wasm-ld /usr/bin/wasm-ld-10 100
 
 # Setting a consistent LD_LIBRARY_PATH across the entire environment prevents unnecessary Cargo
 # rebuilds.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:20.04
```
acfoltzer, line 1: Can we revert this, so that we're testing against a broader range of environments? Ideally we would still be testing on xenial, but even with a HWE kernel, xenial doesn't have new enough headers in its `linux-libc-dev` to build the `userfaltfd-sys` crate.

Dockerfile
```diff
@@ -23,12 +23,13 @@ RUN apt-get update \
 	libcsmith-dev \
 	creduce \
 	gcc-multilib \
-	clang-6.0 \
-	llvm-6.0 \
+	clang \
+	llvm \
+	lld \
+	wabt \
```
acfoltzer, line 29: Does this allow the Rust `wabt` crate to build faster? If not, I would prefer not to add things that aren't used in CI.

Dockerfile
```diff
@@ -43,16 +44,21 @@ RUN rustup target add wasm32-wasi
...
+RUN curl -sS -L https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-10/libclang_rt.builtins-wasm32-wasi-10.0.tar.gz | \
+	tar x -zf - -C /usr/lib/llvm-10/lib/clang/10.0.0
 
-ENV WASI_SDK=/opt/wasi-sdk
+RUN curl -sS -L https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-10/wasi-sysroot-10.0.tar.gz | \
+	tar x -zf - -C /opt
+
+ENV WASI_SYSROOT=/opt/wasi-sysroot
+ENV CLANG=/usr/bin/clang
+ENV CLANG_ROOT=/usr/lib/llvm-10/lib/clang/10.0.0
```
acfoltzer, line 55: We should continue testing against the `wasi-sdk` tools.

docs/src/Compiling-on-Linux.md
```diff
@@ -86,9 +86,8 @@ Install the WASI sysroot:
 
 ```sh
 mkdir -p /opt
-curl -L https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-8/wasi-sdk-8.0-linux.tar.gz | \
-sudo tar x -zv -C /opt -f - wasi-sdk-8.0/share && \
-  sudo ln -s /opt/wasi-sdk-*/share/wasi-sysroot /opt/wasi-sysroot
```
acfoltzer, line 91: We still need this symlink for the default path in the Makefile to work

docs/src/Compiling-on-Linux.md
```diff
@@ -86,9 +86,8 @@ Install the WASI sysroot:
...
+RUN curl -sS -L https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-10/wasi-sysroot-10.0.tar.gz | \
+  tar x -zf - -C /opt
```
acfoltzer, line 90: Some stray Dockerisms: `RUN` and a lack of `sudo`

PR #545, opened 2020-06-03
integrate wiggle tracing

pchickey:
Update wasmtime dep to include improvements to wiggle to support error transformations and the switch to use `tracing`:
https://github.com/bytecodealliance/wasmtime/pull/1796

Had to pull in some cranelift changes in the process - cranelift-wasm now keeps track of both the wasm function type and the cranelift signature, ends up being a net win that gets rid of some fallible code where we had to reverse-engineer the wasm function type back out of the signature. thank you @fitzgen!

### DIFF
--- a/lucet-wiggle/generate/src/config.rs
+++ b/lucet-wiggle/generate/src/config.rs
@@ -3,9 +3,17 @@ use syn::{
     braced,
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
-    Error, Ident, Result, Token,
+    Error, Result, Token,
 };
 use wiggle_generate::config as w;
+
+mod kw {
+    syn::custom_keyword!(witx);
+    syn::custom_keyword!(witx_literal);
+    syn::custom_keyword!(ctx);
+    syn::custom_keyword!(errors);
+    syn::custom_keyword!(constructor);
+}
 
 #[derive(Debug, Clone)]
 pub struct Config {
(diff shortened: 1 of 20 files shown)


pchickey:
Requires the CI fixes from #546 to get circle ci to pass.

lucet-wiggle/generate/src/config.rs
```diff
@@ -3,37 +3,79 @@ use syn::{
     braced,
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
-    Error, Ident, Result, Token,
+    Error, Result, Token,
 };
 use wiggle_generate::config as w;
 
+mod kw {
```
fst-crenshaw, line 10: You named a number of motivations for this PR: `tracing` and `error transformations` and an update to keep up with cranelift's enhancement's to wasm function type and cranelift signature.  This being the first PR I've reviewed for wiggle, could you please comment on what category of change these `custom_keywords` fall in?

pchickey: Great question! I refactored the parser for the config upstream, and mirrored that refactoring here. Using the custom keywords allows us to use `syn`'s `lookahead` parser-combinator, which means we get better error messages when we come across an unexpected field in the config. The downside is that we have to repeat some of the parsing logic for the various elements from the core config that we are looking for. This isn't the greatest trade-off, but it seems to be the style that `syn` and other parser-combinator users (the `wast` crate exposes very similar primitives) typically do, so hopefully when someone comes across it they say "ah i know whats going on here".

pchickey:
One other change in this PR: lucet-wiggle can insert `pre_hook` and `post_hook` in each hostcall it generates, which is useful for timing and statistics gathering for now. We really ought to make lucet responsible for inserting hooks for timing and statistics gathering w/r/t hostcalls (or even just ordinary wasm calls) because wiggle isn't an ideal mechanism for this - we can't easily insert hooks for lucet-wasi if the hook is defined in a downstream crate. But it is a stopgap that will keep existing c@e stats working until we make it a runtime feature.

lucet-wiggle/generate/src/config.rs
```diff
@@ -3,37 +3,79 @@ use syn::{
...
-                id.to_string().as_ref(),
-                input,
-                id.span(),
-            )?)),
+        let lookahead = input.lookahead1();
+        if lookahead.peek(kw::constructor) {
+            input.parse::<kw::constructor>()?;
+            input.parse::<Token![:]>()?;
+            let contents;
+            let _lbrace = braced!(contents in input);
```
fst-crenshaw, line 43: I'm curious about the assignment to `_lbrace` here.  Is the assignment necessary, or is the variable name used simply to convey meaning?

pchickey: `let _ = braced!(...)` would be sufficient, but the token that macro returns is specifically the left brace token, which could be used for span information. This is basically the same as the example in the `braced!` docs except I'm not preserving any of the tokens I don't actually need: https://oschwald.github.io/maxminddb-rust/syn/macro.braced.html

lucetc/src/types.rs
```diff
@@ -52,58 +32,25 @@ impl Display for SignatureError {
     }
 }
 
-pub fn to_lucet_signature(value: &ir::Signature) -> Result<Signature, SignatureError> {
-    let mut params: Vec<ValueType> = Vec::new();
-
-    let mut param_iter = value.params.iter();
-
-    // Enforce that the first parameter is VMContext, as Signature assumes.
-    // Even functions declared no-arg take VMContext in reality.
```
fst-crenshaw, line 61: I love deletion.

PR #547, opened 2020-06-09
Minor cleanups

aturon:
This PR contains a few minor cleanups made along the way while getting to know the lucetc code:

* Using `impl AsRef<Path>` and similar rather than indirecting through generics.
* Using `std::fs::read`.

These are both Rust features that likely post-date the original code.

### DIFF
--- a/lucetc/src/lib.rs
+++ b/lucetc/src/lib.rs
@@ -261,7 +261,7 @@ impl<T: AsLucetc> LucetcOpts for T {
 }
 
 impl Lucetc {
-    pub fn new<P: AsRef<Path>>(input: P) -> Self {
+    pub fn new(input: impl AsRef<Path>) -> Self {
         let input = input.as_ref();
         Self {
             input: LucetcInput::Path(input.to_owned()),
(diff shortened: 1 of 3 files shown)


lucetc/src/load.rs
```diff
@@ -28,12 +22,11 @@ pub fn read_module<P: AsRef<Path>>(
...
-    let converted = if wasm_preamble(&bytes) {
-        bytes
+    if wasm_preamble(&bytes) {
+        Ok(bytes)
     } else {
         wat2wasm(bytes).map_err(|err| {
-            let mut result = String::from("wat2wasm error: ");
-            result.push_str(&format!("{}", err));
+            let mut result = format!("wat2wasm error: {}", err);
             match unsafe { std::mem::transmute::<wabt::Error, wabt::ErrorKind>(err) } {
```
iximeow, line 30: oh uh. this shouldn't be necessary anymore. I'll do a cleanup PR for this once this one lands.

lucetc/src/lib.rs
```diff
@@ -261,7 +261,7 @@ impl<T: AsLucetc> LucetcOpts for T {
 }
 
 impl Lucetc {
-    pub fn new<P: AsRef<Path>>(input: P) -> Self {
+    pub fn new(input: impl AsRef<Path>) -> Self {
```
fst-crenshaw, line 264: I like to treat PRs as an opportunity for me to learn something new about Rust and try out my vocab words.  If I understand correctly, this `impl Trait` syntax is a way to introduce a concrete type parameter bounded by the supplied trait.  So, in this case, `input` must implement `AsRef<Path>`.  Do I have that right?

PR #549, opened 2020-06-12
Remove a no-longer-necessary transmute

iximeow:
Once upon a time `wabt::Error` had no way to access the inner `wabt::ErrorKind`. In those days, I'd noticed that `Error` was a struct that contained only an `ErrorKind`, so this transmute was a hack to get the `ErrorKind` and the corresponding message so we could print that out while `wabt` was being tweaked to make this doable safely.

`wabt` long since has been tweaked to make this doable safely.


iximeow:
spurious macos network failures

I will fight the macintosh OS

PR #551, opened 2020-06-19
lucet-runtime: Add async run to Instance and Vmctx

pchickey:
# What

This PR introduces the ability for hostcalls to run `async` Rust code.
The new method `Vmctx::block_on` can run an `async` block of code when used
in concert with the new `async` method `InstanceHandle::run_async`. Machinery
built on top of the yielding mechanism sends a `Future` from the guest stack
to the host stack of `InstanceHandle::run_async`, where the future is blocked on.
The future's result is sent back to the guest stack so that `Vmctx::block_on`
may resume and return the result.

This PR introduces an necessary but critical internal change to the runtime
borrow checking mechanism for the `Vmctx::yield*` family of functions: it
disables runtime borrow checking of all `Vmctx` resources across the yield.
This is not safe in the general case, but we do believe it is safe in this
specific case due to the construction of `InstanceHandle::run_async`, see
the `Safety` section below for more details.

# Why

Presently, if hostcall code (that is, native Rust code which is called on the
guest stack, and only has access to `Vmctx`, not `InstanceHandle`) needs to
run `async` code, it needs to essentially implement these APIs, or something
strictly less expressive, as a library. Adding them directly to Lucet allows
better error reporting (see `TerminatedReason::BlockOnNeedsAsync` below) and
also allows us to disable runtime borrow checking, which allows for more
programs to be written without using `unsafe`. (Presently we use `unsafe`
coercions to escape the runtime borrow checking in our edge daemon.)

# How

In `lucet-runtime-internals`:

The new `mod future` defines the new methods on `Vmctx` and `InstanceHandle`
introduced in this PR. Adam wrote some very good internal docs that I won't
repeat here.

`instance::state::State` has been modified such that the `Running` variant
now has a `{ async_context: bool }` field. This field is inspected by
`Vmctx::block_on` to panic on any calls that happen from a non-async context.
For example, if `Vmctx::block_on` is used from a guest stack spawned by
`InstanceHandle::run`, it is impossible (by construction) for
`InstanceHandle::run`'s caller to downcast and execute the private
`BoxedFuture` type in the yield value. Rather than let the user get into
this bad state, `Vmctx::block_on` will terminate the instance with the new
`TerminatedReason` variant `BlockOnNeedsAsync`.

Runtime borrow checking can be safely skipped for the `Vmctx::yield_impl` call
in `Vmctx::block_on`. We have added an additional paramter `borrow_check:
bool` to `Vmctx::yield_impl` so that `Vmctx::block_on` can disable borrow
checking, but all other users of `Vmctx::yield_impl` still retain it. I judged
this to be less fragile than factoring the call of `Vmctx::ensure_no_borrows`
to every other call site of `Vmctx::yield_impl`.

In `lucet-runtime-tests` a new test suite has been implemented that exercises
the `Vmctx::block_on` and `InstanceHandle::run_async` methods.

# Safety

The `Vmctx::yield*` family of functions perform a runtime borrow check on
the heap, globals, and embedder context objects before yielding. This is
necessary because the same objects are accessible through both `Vmctx`
and `InstanceHandle`, and would be unsafe for the suspended guest stack to
hold a reference to an object (through `Vmctx`) while arbitrary safe code on
the host stack can alias those same objects.

The argument for disabling this check for `Vmctx::block_on`:

* Any code in the `BoxedFuture` is morally equivelant to running on the guest
  stack. Any safety rules we grant to code on the guest stack, we should grant
  to the code in the `BoxedFuture`.
    * The lifetime of the future cannot live longer than `Vmctx`. This
      lifetime is erased (unsafely coerced to `'static`) as part of the
      boxing.
    * When unboxed in `InstanceHandle::run_async`, the lifetime is narrowed
      from `'static` to `'a`, the lifetime bound to `&'a mut self`. This
      ensures the lifetime does not exceed the scope of
      `InstanceHandle::run_async`.
    * The only arbitrary user code aside from the `BoxedFuture` in in
      `InstanceHandle::run_async` is the `wrap_blocking` closure. This
      closure does is protected from being able to construct or inspect
      anything about the `BoxedFuture` by the privacy rules of the opaque
      `Bounce` wrapper type.
    * `InstanceHandle::run_async` does not attempt to access any resources
      that are accessible from `Vmctx`.


lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -0,0 +1,135 @@
...
+            let run = if self.is_yielded() {
+                // A previous iteration of the loop stored the ResumeVal in
+                // `resume_val`, send it back to the guest ctx and continue
+                // running:
+                self._resume_with_val(
+                    resume_val
+                        .take()
+                        .expect("is_yielded implies resume_value is some"),
+                    true,
+                )
```
acfoltzer, line 86: We need to provide a means to run this in [`tokio::task::spawn_blocking`](https://docs.rs/tokio/0.2.21/tokio/task/fn.spawn_blocking.html), but parameterizing this function by that type requires a rank 2 type! Argh.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -0,0 +1,135 @@
...
+            } else {
+                // This is the first iteration, call the entrypoint:
+                let func = self.module.get_export_func(entrypoint)?;
+                self.run_func(func, args, true)
```
acfoltzer, line 90: This likewise needs to be in a blocking context.

lucet-runtime/lucet-runtime-internals/src/future.rs
acfoltzer: Addressed by fac6457

lucet-runtime/lucet-runtime-internals/src/future.rs
acfoltzer: Addressed by fac6457

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -0,0 +1,189 @@
...
+            });
+            match run {
+                Ok(run_result) => {
+                    // Check if the yield came from Vmctx::block_on:
+                    if run_result.has_yielded::<YieldedFuture>() {
+                        let YieldedFuture(future) = *run_result
+                            .yielded()
+                            .unwrap()
+                            .downcast::<YieldedFuture>()
+                            .unwrap();
```
acfoltzer: Instead of these unwraps, how about adding a method:

```rust
impl RunResult {
    fn downcast_yielded<T: Any + 'static>(self) -> Result<T, Self> { ... }
}
```

lucet-runtime/lucet-runtime-internals/src/instance/state.rs
```diff
@@ -21,7 +21,7 @@ pub enum State {
     ///
     /// Transitions to `Ready` when the guest function returns normally, or to `Faulted`,
     /// `Terminating`, or `Yielding` if the instance faults, terminates, or yields.
-    Running,
+    Running { async_context: bool },
```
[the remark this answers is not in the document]
pchickey: https://github.com/bytecodealliance/lucet/pull/551/commits/bdb535091ccfc4367051bbdca26c505485649d36

lucet-runtime/lucet-runtime-tests/src/async_hostcall.rs
```diff
@@ -0,0 +1,104 @@
+#[macro_export]
+macro_rules! async_hostcall_tests {
+    ( $( $region_id:ident => $TestRegion:path ),* ) => {
+
+        use lucet_runtime::{vmctx::Vmctx, lucet_hostcall};
+
+        #[lucet_hostcall]
+        #[no_mangle]
+        pub fn hostcall_containing_yield(vmctx: &Vmctx, value: u32) {
```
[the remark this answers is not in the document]
pchickey: https://github.com/bytecodealliance/lucet/pull/551/commits/24ad65e2c9ee00e6394d206d56446e9de6f5c8bf

lucet-runtime/lucet-runtime-tests/src/async_hostcall.rs
```diff
@@ -0,0 +1,104 @@
...
+                    ///
+                    /// [tokio]: https://docs.rs/tokio/0.2.21/tokio/task/fn.block_in_place.html
+                    fn block_in_place<F, R>(f: F) -> R
+                    where
+                        F: FnOnce() -> R,
+                    {
+                        f()
+                    }
+
+                    test_nonex(|| {
```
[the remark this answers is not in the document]
pchickey: https://github.com/bytecodealliance/lucet/pull/551/commits/24ad65e2c9ee00e6394d206d56446e9de6f5c8bf

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -0,0 +1,189 @@
...
+    ///   threads than otherwise would be necessary.
+    ///
+    /// This `block_on` operator instead yields yields a future back to a loop that runs in
+    /// `Instance::run_async`, which `.await`s on it and then resumes the instance with the
+    /// result. The future runs on the same runtime that invoked `run_async`, avoiding problems of
+    /// nesting, and allowing the current OS thread to continue performing other async work.
+    ///
+    /// Note that this method may only be used if `Instance::run_async` was used to run the VM,
+    /// otherwise it will terminate the instance with `TerminationDetails::AwaitNeedsAsync`.
+    pub fn block_on<'a, R>(&'a self, f: impl Future<Output = R> + 'a) -> R
```
aturon, line 45: I *really* like how clear this API signature is!

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -0,0 +1,189 @@
...
+                    panic!(TerminationDetails::AwaitNeedsAsync)
+                }
+            }
+            _ => unreachable!("Access to vmctx implies instance is Running"),
+        }
+        // Wrap the Output of `f` as a boxed ResumeVal. Then, box the entire
+        // async computation.
+        let f = Box::pin(async move { ResumeVal(Box::new(f.await)) });
+        // Change the lifetime of the async computation from `'a` to `'static.
+        // We need to lie about this lifetime so that `YieldedFuture` may impl
```
aturon, line 62: Mind adding a note about why this is safe? AIUI the explanation is basically: the relevant stack frame (that `'a` is tied to) is essentially frozen in place during this process.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// block on an executor using `futures::executor::block_on` or
+    /// `tokio::runtime::Runtime::block_on`, that has two drawbacks:
+    ///
+    /// - If the Lucet instance was originally invoked from an async context, trying to block on the
+    ///   same runtime will fail if the executor cannot be nested (all executors we know of have this
+    ///   restriction).
+    ///
+    /// - The current OS thread would be blocked on the result of the computation, rather than being
+    ///   able to run other async tasks while awaiting. This means an application will need more
+    ///   threads than otherwise would be necessary.
```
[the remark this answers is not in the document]
pchickey: Adam wrote this :)

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -0,0 +1,189 @@
...
+    ///
+    /// This method is similar to `Instance::run()`, but allows the Wasm program to invoke hostcalls
+    /// that use `Vmctx::block_on` and provides the trampoline that `.await`s those futures on
+    /// behalf of the guest.
+    ///
+    /// # `Vmctx` Restrictions
+    ///
+    /// This method permits the use of `Vmctx::block_on`, but disallows all other uses of `Vmctx::
+    /// yield_val_expecting_val` and family (`Vmctx::yield_`, `Vmctx::yield_expecting_val`,
+    /// `Vmctx::yield_val`).
```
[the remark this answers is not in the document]
pchickey: Right now, we have some non-public code that doesn't quite work correctly with `block_on` yet. I think that leaving these in, for now, is basically a hedge that we can manually create and pass datatype continuations if `block_on` is insufficient.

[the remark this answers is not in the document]
aturon: Roger that!

PR #554, opened 2020-06-25
dockerfile: install lld-9 and use it as system ld

pchickey:
this reportedly improves `cargo build` times a bunch


iximeow:
While `lld-9` works and does appear to link faster, it [causes some problems](https://github.com/bytecodealliance/lucet/issues/563) that would block unwinding/backtraces from landing, so I'm gonna close this for the time being.

PR #559, opened 2020-06-26
Add `wrap_blocking` argument to `run_async`; rename `run_await`

acfoltzer:
This godawful type signature enables callers to make sure Wasm execution happens on a blocking thread, for example by passing `|f| tokio::task::spawn_blocking(f)` as the `wrap_blocking` argument.

Renamed `run_await` to `block_on` for consistency with other async libraries.


pchickey:
this looks really great. CI seems to be complaining about the doc code samples, but once those are fixed is this ready to go?

acfoltzer:
Marking this as ready for review once again; I thought there might be a cleaner way to write this but that's looking unlikely. @aturon and I are going to take a look later but I don't want to block this

pchickey:
Cool, we can always pull more stuff into #551 if we need it

PR #567, opened 2020-07-09
Ensure a requested amount of stack is avaiable before hostcalls

iximeow:
This changes `lucetc`'s generation of hostcalls to, instead of directly calling hostcalls, call a synthetic trampoline to do some safety checks. That trampoline then will call the original hostcall, if it passes the required checks.

In our case, the only check is that the guest's stack use is below some threshold that the instance has set, ensuring that an amount of space is available for hostcalls. My arbitrary selection for a hostcall stack space guarantee is 16kb. If the guest fails the stack availability check, it terminates with a stack overflow - this mitigates the issues if a stack overflow in host code by making it an explicit stack overflow in the guest when reasonable. Hostcall stack overflows may still occur, and are still fatal, but we can now guarantee a certain amount of space for hostcall authors. This means hostcall implementors have the same kind of constraints as writing any other function - a stack overflow in any other host code would be fatal in a way beyond our control, just the same.

This adds a notable testing complication: since lucetc inserts non-wasm-derived details now, the mock modules we use in `lucet-runtime` tests are a little further from being real modules. Mock modules don't have trampolines between mocked "guest" code and hostcalls they would make, and I don't think we can insert trampolines if we wanted (without a worrying maintenance burden and risk of worse mock/real skew). My conclusion is that testing details related to hostcall stack overflows must be done using real wasm modules through lucetc, and tests not involving hostcall stack overflows (so, the ones that exist already), can continue being against mock modules.


lucetc/src/compiler.rs
```diff
@@ -308,6 +321,113 @@ impl<'a> Compiler<'a> {
...
+        //         terminate_with_stack_overflow();
+        //     }
+        //
+        //     $hostcall(vmctx, $hostcall_args..)
+        // }
+        // ```
+        //
+        // but are specified here as Cranelift IR for lack of source to generate them from.
+        for (hostcall_name, (trampoline_id, hostcall_func_index)) in &self.trampolines {
+            let mut trampoline_context = ClifContext::new();
```
iximeow, line 341: the enclosing function is getting a bit big, I'm planning on taking this loop body and moving it to a `fn synthesize_trampoline(trampoline_id, hostcall_func_index)`

lucetc/src/function.rs
```diff
@@ -431,9 +470,24 @@ impl<'a> FuncEnvironment for FuncInfo<'a> {
             .expect("function indices are valid");
         let func_decl = self.module_decls.get_func(unique_index).unwrap();
         let signature = func.import_signature(func_decl.signature.clone());
+
+        // if we're setting up a function ref for a call to an imported function, we'll need a
+        // trampoline to check the stack first. in that case, return the trampoline function
+        // instead.
         let colocated = !func_decl.imported();
+        let name = if colocated {
```
iximeow, line 478: I also think this might be not _quite_ right, Cranelift docs suggest that guest functions can be `!colocated` if the module is sufficiently large (f.ex larger than 128mb of code). We would then insert a trampoline between calls internal to the guest, and then have guests fail due to not having enough stack for a hostcall when not making a hostcall at all.

I haven't _actually seen_ guest functions that are `!colocated`, and I'm not sure if there's a better way to select imported functions here. Even guest functions were erroneously picked here, it would work (with an unintentional penalty of extra function call) so long as stack is available - the error mode in case this _is_ wrong is not _too_ bad.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -851,6 +851,25 @@ impl Instance {
...
+    #[inline]
+    pub fn set_hostcall_stack_reservation(&mut self, reservation: usize) {
+        let slot = self
+            .alloc
+            .slot
+            .as_ref()
+            .expect("reachable instance has a slot");
+
+        if reservation > slot.limits.stack_size {
+            panic!("TODO: error this");
```
iximeow, line 864: I'm actually not sure if making this `Result`-y _is_ what we want. It means either we make `Instance::new` a `Result`-y function, or make it panicky; a custom `Limit` the default hostcall reservation may become excessively sized.

On the other hand, someone might want to gracefully error on instances with too-small stacks?

Either way, I don't think we can push this check anywhere else (like `Limits` creation), since `set_hostcall_stack_reservation` can be called at any arbitrary point later in the `Instance`'s lifecycle.

lucet-wasi/src/main.rs
```diff
@@ -173,13 +173,12 @@ fn main() {
...
-        heap_memory_size,
-        heap_address_space_size,
-        stack_size,
-        globals_size: 0, // calculated from module
-        ..Limits::default()
-    };
+    let limits = Limits::default()
+        .with_heap_memory_size(heap_memory_size)
+        .with_heap_address_space_size(heap_address_space_size)
+        .with_stack_size(stack_size)
```
iximeow, line 179: If you `lucet-wasi` with a stack smaller than `32KiB` we'll panic here. I don't see a reason to make the hostcall reservation configurable here, so that seems fine?

lucet-runtime/lucet-runtime-internals/src/c_api.rs
```diff
@@ -182,13 +185,15 @@ impl From<lucet_alloc_limits> for Limits {
...
-            heap_address_space_size: limits.heap_address_space_size as usize,
-            stack_size: limits.stack_size as usize,
-            globals_size: limits.globals_size as usize,
-            signal_stack_size: limits.signal_stack_size as usize,
-        }
+        Limits::default()
+            .with_heap_memory_size(limits.heap_memory_size as usize)
+            .with_heap_address_space_size(limits.heap_address_space_size as usize)
+            .with_stack_size(limits.stack_size as usize)
+            .expect("lucet_alloc_limts has a valid stack size")
```
iximeow, line 192: The `Limits` changes also make the C API a little more panicky, but we can't enforce the same consistency in building a `lucet_alloc_limits`, so this is the first place to find an invalid setting.

lucet-runtime/lucet-runtime-internals/src/instance.rs
iximeow: I ended up moving `hostcall_reservation` onto `Limits`, and added a hidden field so a `Limits` instance can only be constructed through methods that check the consistency of `stack_size`/`hostcall_reservation`. Those are now `Result`-y builder methods, and seems to be a pretty usable way to spell this out.

lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
```diff
@@ -484,10 +488,56 @@ impl Limits {
...
+    }
+    pub fn with_heap_address_space_size(mut self, heap_address_space_size: usize) -> Self {
+        self.heap_address_space_size = heap_address_space_size;
+        self
+    }
+    pub fn with_stack_size(mut self, stack_size: usize) -> Result<Self, Error> {
+        if stack_size <= self.hostcall_reservation {
+            return Err(Error::InvalidArgument(
+                "stack size must be greater than hostcall reserved space",
+            ));
```
acfoltzer, line 515: As discussed in Slack, we should be able to do these checks in `Limits::validate()` rather than when constructing limits. However, I really like having these interfaces!

Would you mind moving the check, and then removing `_hidden` and making the fields public once more?

As a bonus, with the checks gone these builder methods should all be `const fn`-able.

lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
```diff
@@ -484,10 +488,56 @@ impl Limits {
...
+
+        self.stack_size = stack_size;
+        Ok(self)
+    }
+    pub fn with_hostcall_reservation(mut self, hostcall_reservation: usize) -> Result<Self, Error> {
+        // We allow `hostcall_reservation == self.stack_size`, a circumstance that guarantees
+        // any hostcalls will fail with a StackOverflow.
+        if hostcall_reservation > self.stack_size {
+            return Err(Error::InvalidArgument(
+                "hostcall reserved space must be less than stack size",
```
acfoltzer, line 526:
```suggestion
                "hostcall reserved space must be less than or equal to stack size",
```

lucet-runtime/lucet-runtime-internals/src/c_api.rs
acfoltzer: This should be resolved by moving the check to `validate`

acfoltzer:
Also, if you're up for it, adding a page describing this to `/docs/src` would be greatly appreciated. There is hopefully material that exists already about the high-level design that you can adapt, but also a description of the low-level implementation via Cranelift IR would be :100:

lucet-wasi/src/main.rs
iximeow: Done!

lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
iximeow: Builder methods are now `const fn`'d, `validate` is absolutely the right place for these checks. Done!

lucetc/src/compiler.rs
```diff
@@ -496,6 +526,117 @@ impl<'a> Compiler<'a> {
...
+        .create_global_value(ir::GlobalValueData::VMContext);
+    let stack_limit_gv = trampoline_context
+        .func
+        .create_global_value(ir::GlobalValueData::Load {
+            base: vmctx,
+            offset: (-(std::mem::size_of::<InstanceRuntimeData>() as i32)
+                + (offset_of!(InstanceRuntimeData, stack_limit) as i32))
+                .into(),
+            global_type: ir::types::I64,
+            readonly: false,
```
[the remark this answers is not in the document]
iximeow: Nope, I'd filled in thinking about a different API. This is now `readonly: true`.

lucetc/src/function.rs
```diff
@@ -265,6 +277,32 @@ impl<'a> FuncInfo<'a> {
...
+) -> Result<ir::ExternalName, ClifModuleError> {
+    // decls.declare_new_function(clif_module, name, linkage, 0
+    use std::collections::hash_map::Entry;
+    let funcid = match trampolines.entry(func_decl.name.symbol().to_string()) {
+        Entry::Occupied(o) => o.get().0,
+        Entry::Vacant(v) => {
+            let trampoline_name = format!("trampoline_{}", func_decl.name.symbol());
+            println!("declaring trampoline {}", trampoline_name);
+
+            // decls.declare_function(clif_module, trampoline_name, Linkage::Local)
```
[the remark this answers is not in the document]
iximeow: yep :(

iximeow:
Docs and tests (and cleanup) have been added! The two additional tests I've added are pulled from `stack` and position the stack _near_ but not quite overflowing, testing behavior with a default, and then custom (very shallow) hostcall reservation.

Docs were good because I had to go through a minor exercise justifying why the trampoline stack check's condition was correct; Cranelift's operand order for `ifcmp_sp` is backwards from what you might think, if you haven't read the documentation recently!

iximeow:
`Wrote only 9728 of 10240 bytes tar: home/circleci/project/target/release/deps/lucet_wasi-8e6fe6c44d08f0fe.d: Cannot write: No space left on device` okay, CI

docs/src/lucetc/hostcalls.md
```diff
@@ -0,0 +1,130 @@
...
+
+Why do this trampoline and stack usage test, instead of observing failures in
+guest stack guards? The issue here is twofold: first, we can't robustly
+distinguish a stack overflow from an unlucky errant memory access for other
+reasons. If hosts are built using stack probes, stack overflows will probably
+be observed in places we expect, with patterns we can expect. But this is by no
+means a guarantee, and stack accesses might not be through the stack pointer
+directly (perhaps the address is loaded into an alternate register, and a fault
+occurs without referencing the stack pointer at all). Second, if a hostcall
+fault were to be recoverd by Lucet, the runtime may have to unwind a guest that
```
acfoltzer, line 121:
```suggestion
fault were to be recovered by Lucet, the runtime may have to unwind a guest that
```

PR #568, opened 2020-07-13
📉 Reduce CircleCI cache sizes by keying on `rust-toolchain`

acfoltzer:
This should prevent the cache of the `/target` directory from accruing complete sets of artifacts from multiple `rustc` versions over time.

This patch also fixes the use of `{{ .Environment.CACHE_VERSION }}` so that it applies to Linux, and so that it forms proper prefixes in the macOS cache keys.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,12 +95,15 @@ commands:
                   # 2. A Linux cache built on a matching branch name with any `Cargo.lock` checksum
                   # 3. Any Linux cache
                   #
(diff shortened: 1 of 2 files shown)


acfoltzer:
We still will have some issues with cruft accruing as Cargo.lock dependencies evolve _while_ we are pinned to a particular rustc version. Hopefully we'll be able to continue updating the rustc version regularly enough that it won't get too bad before we wipe the slate clean for the new version.

PR #571, opened 2020-07-29
lucetc: add command line option to disable wat translation

pchickey:
wat translation is a potential attack vector, since it happens in the C++ `wabt` library, so we want a way to turn it off if required.

```sh
[lucet]% cargo run -p lucetc -- lucetc/tests/wasm/globals_export.wat --no-translate-wat
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/lucetc lucetc/tests/wasm/globals_export.wat --no-translate-wat`
Error: Input does not have Wasm preamble
```

### DIFF
--- a/lucetc/src/load.rs
+++ b/lucetc/src/load.rs
@@ -7,6 +7,7 @@ pub fn read_module(
     path: impl AsRef<Path>,
     pk: &Option<PublicKey>,
     verify: bool,
+    translate_wat: bool,
 ) -> Result<Vec<u8>, Error> {
     let contents = std::fs::read(&path)?;
     if verify {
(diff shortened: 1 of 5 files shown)


lucetc/src/load.rs
```diff
@@ -18,7 +19,11 @@ pub fn read_module(
                 .ok_or(Error::Signature("public key is missing".to_string()))?,
         )?;
     }
-    read_bytes(contents)
+    if translate_wat {
+        read_bytes(contents)
+    } else {
+        Ok(contents)
```
iximeow, line 25: we ought to still check `wasm_preamble` in the no-`wat` case, since the only thing Lucet should accept is a `wasm`.

pchickey: Fixed. Thanks

PR #572, opened 2020-07-29
💢 Package unknown hostcall panics into termination details

acfoltzer:
This prevents panics that aren't intentionally initiated by the Lucet runtime from attempting to unwind across Wasm stack frames. When #254 lands, this will no longer be necessary, but until then this allows us to safely transport the unknown panic payload across the FFI boundary so that panicking can resume on the other side.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/c_api.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/c_api.rs
@@ -309,14 +309,20 @@ pub mod lucet_result {
                             },
                             TerminationDetails::Provided(p) => lucet_terminated {
                                 reason: lucet_terminated_reason::Provided,
-                                provided: p
+                                provided: *p
                                     .downcast_ref()
-                                    .map(|CTerminationDetails { details }| *details)
-                                    .unwrap_or(ptr::null_mut()),
+                                    .map(|CTerminationDetails { details }| details)
+                                    .unwrap_or(&ptr::null_mut()),
                             },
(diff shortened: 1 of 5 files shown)


lucet-runtime/lucet-runtime-internals/src/c_api.rs
```diff
@@ -309,15 +309,21 @@ pub mod lucet_result {
...
+                                    .unwrap_or(&ptr::null_mut()),
                             },
                             TerminationDetails::Remote => lucet_terminated {
                                 reason: lucet_terminated_reason::Remote,
                                 provided: std::ptr::null_mut(),
                             },
+                            TerminationDetails::OtherPanic(p) => lucet_terminated {
+                                reason: lucet_terminated_reason::OtherPanic,
+                                // double box the panic payload so that the pointer passed to FFI
+                                // land is thin
```
iximeow, line 324: took a moment to remember that non-thin-pointer types don't have FFI-stable layouts. That's what's going on here, right?

acfoltzer: Exactly; `Box::into_raw(x)` where `x` is `Box<dyn T>` returns a `*mut dyn T`, which is two machine pointers wide for the vtable, so casting it to `*mut c_void` to store in the struct would be Bad. The extra box makes the returned raw pointer have the type `*mut Box<dyn T>`, which is fine because the vtable pointer is then stored in the box on the heap.

PR #575, opened 2020-08-07
move lucet-validate into lucetc crate

pchickey:
`lucet-validate` was creating more trouble as a separate crate than it was worth. And, to fix some design problems, I ended up rewriting it piece-by-piece.

* `lucet-validate` crate is gone. What is left of it and its tests are merged into `lucetc`. This eliminates a dev-dependencies loop that caused us problems with `cargo publish`.
* `lucetc::validate` now uses a builder interface for construction, which is more flexible than before. It supports both wasi commands and reactors, though we don't do anything with reactors at the moment (lucet-runtime doesnt know what to do with them yet).
* a `Validator`, once constructed, gets its `register_import` and `register_export` methods called by the `ModuleInfo` struct's `ModuleEnvironment` impl. This means that as `cranelift_wasm::translate_module` parses the wasm module, we are validating each function import and export. Previously, `lucet-validate` had a `ModuleType` struct that was constructed by a totally separate parsing pass on the wasm module; this was never a great design, it was just the quick and dirty version.
* Rather than fail on the first error, a `Validator` collects all of its errors internally. After `translate_module` is complete, we get the whole set of validation errors out of `Validator` and, if there are any, report them all and stop compilation. 
* Tests were added for the new behavior.

The PR is arranged as a series of commits that transform the `lucet-validate` code into the final form step by step, with tests passing at each commit, if you want to read it that way.

This PR closes #438: now all validation errors are reported, with positive and negative tests.


katie-martin-fastly (dismissed):
Great work! Thank you :slightly_smiling_face:

PR #583, opened 2020-09-09
🈵 Add an option to terminate the runtime when `memory.grow` fails

acfoltzer:
This new option determines whether to terminate the guest with a new `TerminationDetails::HeapOutOfMemory` variant when `memory.grow` fails, rather than returning `-1`. It is disabled by default, but can be set via `InstanceBuilder` or directly on the `Instance`.

This behavior deviates from the WebAssembly spec, but is useful in practice for determining when guest programs fail due to an exhausted heap.

Most languages will compile to code that includes an `unreachable` instruction if allocation fails, but this same instruction might also appear when other types of assertions fail, `panic!()` is called, etc. Terminating allows the error to be more directly identifiable.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/region.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/region.rs
@@ -51,13 +51,7 @@ pub trait Region: RegionInternal {
 
 /// A `RegionInternal` is a collection of `Slot`s which are managed as a whole.
 pub trait RegionInternal: Send + Sync {
-    fn new_instance_with(
-        &self,
-        module: Arc<dyn Module>,
-        embed_ctx: CtxMap,
-        heap_memory_size_limit: usize,
-        alloc_strategy: AllocStrategy,
-    ) -> Result<InstanceHandle, Error>;
+    fn new_instance_with(&self, args: NewInstanceArgs) -> Result<InstanceHandle, Error>;
 
     /// Unmaps the heap, stack, and globals of an `Alloc`, while retaining the virtual address
     /// ranges in its `Slot`.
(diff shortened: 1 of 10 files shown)


lucet-runtime/lucet-runtime-internals/src/region.rs
```diff
@@ -51,13 +51,7 @@ pub trait Region: RegionInternal {
...
 /// A `RegionInternal` is a collection of `Slot`s which are managed as a whole.
 pub trait RegionInternal: Send + Sync {
-    fn new_instance_with(
-        &self,
-        module: Arc<dyn Module>,
-        embed_ctx: CtxMap,
-        heap_memory_size_limit: usize,
-        alloc_strategy: AllocStrategy,
-    ) -> Result<InstanceHandle, Error>;
+    fn new_instance_with(&self, args: NewInstanceArgs) -> Result<InstanceHandle, Error>;
```
fst-crenshaw, line 54: Thanks for doing this added refactor.  It was time.

fst-crenshaw (approved):
Thanks so much for dropping everything today to do this urgent work.  I am grateful for your Good Decisions Under Pressure.  ❤️   

Concretely, I am grateful for your trio of tests, with especial appreciation for making sure the instance still works as expected when it stays within its heap bounds.  I am also grateful for the useful refactor you did to create `NewInstanceArgs` instead of adding yet another parameter to `new_instance_with`.

acfoltzer:
Blocking this pending #584

fst-crenshaw (approved):
Thanks for addressing the UB that emerged from the recent addition of the [lucet_hostcall]!  Thanks for updating this PR!

PR #584, opened 2020-09-10
☢️ Remove UB from `extern "C"` ABI on annotated hostcalls

acfoltzer:
This was a programming error introduced in the transition between the `lucet_hostcalls!` macro and the `#[lucet_hostcall]` attribute.

The expansion of the code in the attribute called the annotated function from within a `catch_unwind` without modifying its ABI. Since it is UB to unwind from a non-Rust ABI function, release mode optimizations would elide the code which would otherwise handle an unwind from the function. This caused the process to abort if the function ever _did_ unwind, since the system unwinding runtime would reach the end of the guest stack without finding any frames willing to handle the exception.

I believe this behavior was very difficult, if not impossible, to observe in practice with the current state of the Lucet codebase. The only unwinding that these functions could directly induce are the assertions in `vmctx::instance_from_vmctx` that should never be violated except in the case of a Lucet programming error, or a dynamic linking problem.

However, in #583 we are adding an option to instances that would cause `lucet_vmctx_grow_memory` to potentially terminate the instance via the `lucet_hostcall_terminate!()` mechanism that is implemented via unwinding. We discovered this problem during testing on that branch, and so we need to land this fix before it can proceed.

This patch contains a fix to the immediate problem by removing the erroneous `extern "C"` specifiers. It also prevents the problem recurring in the future by generating a compile-time error from `#[lucet_hostcall]` if there is an ABI specifier present on the function it annotates.

### DIFF
--- a/lucet-runtime/lucet-runtime-macros/src/lib.rs
+++ b/lucet-runtime/lucet-runtime-macros/src/lib.rs
@@ -1,6 +1,6 @@
 extern crate proc_macro;
 use proc_macro::TokenStream;
-use quote::quote;
+use quote::{quote, quote_spanned};
 use syn::spanned::Spanned;
 
 /// This attribute generates a Lucet hostcall from a standalone Rust function that takes a `&mut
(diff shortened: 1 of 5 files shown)


lucet-runtime/lucet-runtime-macros/src/lib.rs
```diff
@@ -39,6 +39,12 @@ pub fn lucet_hostcall(_attr: TokenStream, item: TokenStream) -> TokenStream {
     };
 
     let mut hostcall = syn::parse_macro_input!(item as syn::ItemFn);
+    if let Some(abi) = &hostcall.sig.abi {
+        let err = quote_spanned! {abi.span()=>
+            compile_error!("Functions annotated with `#[lucet_hostcall]` may not have an `extern` ABI");
```
iximeow, line 44: dear future readers: if you're `extern "Rust"`ing and this incorrectly rejects your program, please say hello! also, why are you using `extern "Rust"`!

iximeow (approved):
What a frustrating bug, good catch!

PR #588, opened 2020-09-16
lucet-wasi executable: add tracing_subscriber

pchickey:
required for e.g. RUST_LOG=wasi-common=trace


katie-martin-fastly (approved):
Pardon the slow review, balancing aliases is hard! This looks wonderful to me, thank you for this thoughtful addition.

PR #590, opened 2020-09-23
lucet-wasi-generate: accept pre_hook and post_hook as proc macro arguments

pchickey:
in the process, convert the option parsing to use kw & lookahead like its friends do in wiggle-generate


iximeow (approved):
looks good, thank you!

PR #592, opened 2020-10-05
use $GITHUB_PATH rather than ##[add-path]

pchickey:
In accordance with this advisory it's recommended we move to a
different scheme of setting env vars and updating PATH:

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Thank you @alexcrichton for bringing this to my attention


katie-martin-fastly (approved):
:eagle: :eyes:  Good eye! Thank you for putting this together, and for resolving the funky CI failure.

PR #593, opened 2020-10-08
Update wasmtime: new clif module, wasmparser late validation

pchickey:
Contains fixes for cranelift-module API: https://github.com/bytecodealliance/wasmtime/pull/2249

Uses the latest `wasmparser` integration with `cranelift-wasm`, which performs validation of function bodies as they are consumed. This means we can remove the up-front `wasmparser` validation of the entire module, and rely entirely on `cranelift-wasm` for validation!

These changes to `cranelift-wasm` meant we had to change the way function bodies (& their new associated validator) were owned by `lucetc::Compiler`. This was mostly code motion.

We had a test for `lucetc` that it rejected an invalid wasm module, but the sources were lost to time[1] . This test has been rewritten based on `call.wat` to introduce a type error at a specific opcode, and test that we get a validation error at that opcode.

[1]: The comments had this to say about `invalid.wasm`, which I derived from the spec tests a very long time ago: 
```
        // I used the `wast2json` tool to produce the file invalid.wasm from an assert_invalid part
        // of a spectest (call.wast)
        let wasmfile = PathBuf::from("tests/wasm/invalid.wasm");
```


lucetc/tests/wasm.rs
```diff
@@ -370,32 +370,6 @@ mod module_data {
...
-        let mut m = Vec::new();
-        let mut file = File::open(&wasmfile).expect("open module file");
-        file.read_to_end(&mut m).expect("read contents of module");
-
-        let b = Bindings::empty();
-        let builder = Compiler::builder();
-        let c = builder.create(&m, &b);
-        assert!(
-            c.is_err(),
-            "compilation error because wasm module is invalid"
```
alexcrichton, line 390: FWIW a crate like `wat` makes it easy to make wasm modules that are invalid at least at the type level (binary level is a bit more difficult), but the new test here seems fine either way.

alexcrichton (approved):

lucetc/tests/wasm.rs
pchickey: I didn't actually try using `wat` because my intuition was that it would reject creating a binary of an improperly typed module. This test is a bit more janky than I usually like, but I wasn't able to come up with a better idea.

PR #595, opened 2020-10-22
lucetc: parallelize codegen with rayon

pchickey:
This PR enables parallel code generation in `lucetc` by way of the `rayon` crate.

The internals of lucetc were re-organized in terms of ownership and send/sync safety, but should be functionally identical.

`cranelift-module` was modified to enable parallel code generation upstream: https://github.com/bytecodealliance/wasmtime/pull/2371

One side-effect of this parallelism is that section ordering in the resulting object file is no longer deterministic. This would be a desirable property in the long run but isn't necessary right now.

I authored the first bit of this and @alexcrichton finished it up, so we'll each sign-off on the others work here.


lucet-module/Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2018"
 
 [dependencies]
 anyhow = "1.0"
-cranelift-entity = { path = "../wasmtime/cranelift/entity", version = "0.67.0" }
+cranelift-entity = { path = "../wasmtime/cranelift/entity" }
```
pchickey, line 14: afaik, we need to leave these `version` indicators in the tree for the sake of publishing to crates.io (which we rarely do, but dont want to box ourselves out of). I know its a big pain to keep them synced up, but I don't know how to publish without them there - maybe there is some other fix for that?

lucetc/src/compiler.rs
```diff
@@ -261,152 +255,153 @@ impl<'a> Compiler<'a> {
...
-                &mut self.decls,
-                &mut self.clif_module,
-                &mut function_map,
-                hostcall_name,
-                *trampoline_id,
-                *hostcall_func_index,
-            )?;
+        let trampoline_metas = codegen_context
+            .trampolines()
+            .iter()
```
alexcrichton, line 303: FWIW this is probably far less impactful than the main body of code but `.par_iter()` is likely imminently plausible here.

lucetc/src/compiler.rs
```diff
@@ -516,28 +505,34 @@ struct TrapMetadata {
...
 
-#[derive(Clone)]
 pub struct CodegenContext {
+    isa: Box<dyn TargetIsa>,
     // the `FuncId` references the declared trampoline function Cranelift knows, but the
     // `UniqueFuncIndex` references the hostcall being trampoline'd to.
-    trampolines: Arc<Mutex<HashMap<String, (FuncId, UniqueFuncIndex)>>>,
-    clif_module: Arc<Mutex<ObjectModule>>,
+    trampolines: Mutex<HashMap<String, (FuncId, UniqueFuncIndex)>>,
+    clif_module: Mutex<ObjectModule>,
```
pchickey, line 513: thank you for getting rid of this Arc!

alexcrichton (approved):
I've got parts of this PR too but wanted to log that I looked at all the previous parts as well! I think there's one more possibility for parallelism but either way lgtm

lucetc/src/compiler.rs
```diff
@@ -516,28 +505,34 @@ struct TrapMetadata {
...
 impl CodegenContext {
-    pub fn new(isa: Box<dyn TargetIsa>) -> Result<CodegenContext, Error> {
+    // Note that two `TargetIsa` objects are required here since one will live
+    // inside of the lock holding `ObjectModule` and one will live outside. Not
+    // exactly a great state to be in but should hopefully get the job done for
+    // now.
+    pub fn new(
+        isa: Box<dyn TargetIsa>,
+        isa_copy: Box<dyn TargetIsa>,
+    ) -> Result<CodegenContext, Error> {
```
pchickey, line 524: nice

lucetc/src/compiler.rs
pchickey: great idea, done in aedf40349df48c2ac950d6ac526893b62e69dd40

lucet-module/Cargo.toml
alexcrichton: Ah I figured that publication to crates.io wasn't on the table, but if that's desired I'll add these back in and update them all

pchickey: yeah :/ the last time we got around to publishing was back in February, which is pretty embarrassing. We use git, not the crate version, to manage this dependency internally, so publishing is never on our mind.

alexcrichton (dismissed):

alexcrichton (approved):

alexcrichton (approved):

PR #600, opened 2020-11-06
Compiling on linux: MD; fix typo for linking

Horki:
Fix 'typo' linking path for "wasi sysroot"

`sudo ln -s /opt/wasi-sdk-*/share/wasi-sysroot /opt/wasi-sysroot`
=>
`sudo ln -s /opt/wasi-sdk-11.0/share/wasi-sysroot /opt/wasi-sysroot`

### DIFF
--- a/docs/src/Compiling-on-Linux.md
+++ b/docs/src/Compiling-on-Linux.md
@@ -88,7 +88,7 @@ Install the WASI sysroot:
 mkdir -p /opt
 curl -L https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-11/wasi-sdk-11.0-linux.tar.gz | \
 sudo tar x -zv -C /opt -f - wasi-sdk-11.0/share && \
-  sudo ln -s /opt/wasi-sdk-*/share/wasi-sysroot /opt/wasi-sysroot
+  sudo ln -s /opt/wasi-sdk-11.0/share/wasi-sysroot /opt/wasi-sysroot
 ```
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi!

This is not a typo :)

`*` is replaced by the shell with any file matching the pattern.

Using this in the documentation instead of hardcoding a specific version helps the instructions stay relevant even as new versions of the SDK are released.

jedisct1:
Just noticed that the previous lines hardcode version numbers. Since the whole paragraph would need to be updated on a new release anyway, we can indeed hardcode that one as well.

A better alternative would be to retrieve the latest version using the `/latest` API as done by most update scripts. But that can be done later.

Thanks!

jedisct1 (approved):

Horki:
@jedisct1 when I use '*' with GNU ln (v8.32), I receive an error
` no matches found: /opt/wasi-sdk-*/share/wasi-sysroot`

pchickey:
Thanks for this fix. The CI failures are due to `cargo audit` and have been fixed upstream - can you pleaserebase on or merge with `main`?

Horki:
@pchickey rebased and pushed

pchickey (approved):

pchickey (changes requested):
It appears your rebased added a bunch of commits to your branch that are already merged into main - not sure what went wrong here.

Horki:
@pchickey fixed

pchickey:
Thanks!

PR #603, opened 2020-11-09
Don't enable the Cranelift verifier in release mode

alexcrichton:
The verifier can end up taking a significant chunk of time. On a local
test case this change decreases the CPU time used by lucetc from 30s to
6.5s.


iximeow (approved):
the very best optimization: do fewer things. thank you!

pchickey:
I have enabled `Build forked pull requests` in circleci - can we see if that gets a status reported on this PR? try force-pushing to trigger it perhaps.

Obsolete by #604

PR #605, opened 2020-11-09
dlmodule version checks optional, uffd configuration options expanded

pchickey:
Two changes, one PR:

1. lucet-runtime::DlModule load now optionally enforces version match
    This check exists to make sure that lucetc and lucet-runtime are built
against the same revision of lucet-module - the use of serde depends
on both sides defining types the same.
    This check can be overridden for cases where the user's runtime needs to
load modules built with a different lucetc. The user is responsible for
ensuring that the modules were created with the same lucet-module type
definitions.

1. lucet-runtime: change uffd configurations to have many more options.
    The user can now configure the strategy of stack initialization and heap initialization as well as the page size. also, configurations are given by a struct now instead of a trait.


acfoltzer (approved):
Gah, sorry this slipped through the cracks. Thank you!

PR #608, opened 2020-11-17
Compling: MD; add a path for ARCH linux

Horki:
Added correct path for ARCH linux.
Please review


pchickey (approved):
I don't know enough about Arch to know whether this is correct, but it doesn't seem unreasonable so we'll take your word for it.

Horki:
@pchickey Hi, thanks for approving PR, this path is [correct](https://www.archlinux.org/packages/extra/x86_64/clang/files/)

PR #612, opened 2020-11-21
Add a "bounded runtime" mode to guest execution.

cfallin:
This modifies the instruction-counting mechanism to track an
instruction-count bound as well; and when a bound is set, the Wasm guest
will make a hostcall to yield back to the host context.

This is all placed under the `run_async()` function, so bounded-execution
yields manifest as futures that are immediately ready to continue execution.
This should give an executor main loop a chance to do other work at regular
intervals.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/future.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/future.rs
@@ -1,10 +1,11 @@
 use crate::error::Error;
-use crate::instance::{InstanceHandle, RunResult, State, TerminationDetails};
+use crate::instance::{InstanceHandle, InternalRunResult, RunResult, State, TerminationDetails};
 use crate::val::{UntypedRetVal, Val};
 use crate::vmctx::{Vmctx, VmctxInternal};
 use std::any::Any;
 use std::future::Future;
 use std::pin::Pin;
+use std::task::{Context, Poll};
 
 /// This is the same type defined by the `futures` library, but we don't need the rest of the
 /// library for this purpose.
@@ -75,7 +76,7 @@ impl Vmctx {
         // Wrap the computation in `YieldedFuture` so that
         // `Instance::run_async` can catch and run it.  We will get the
         // `ResumeVal` we applied to `f` above.
-        self.yield_impl::<YieldedFuture, ResumeVal>(YieldedFuture(f), false);
+        self.yield_impl::<YieldedFuture, ResumeVal>(YieldedFuture(f), false, false);
         let ResumeVal(v) = self.take_resumed_val();
         // We may now downcast and unbox the returned Box<dyn Any> into an `R`
         // again.
@@ -83,15 +84,30 @@ impl Vmctx {
     }
 }
 
-/// This struct needs to be exposed publicly in order for the signature of a
-/// "block_in_place" function to be writable, a concession we must make because
-/// Rust does not have rank 2 types. To prevent the user from inspecting or
-/// constructing the inside of this type, it is completely opaque.
-pub struct Bounce<'a>(BounceInner<'a>);
+/// A simple future that yields once. We use this to yield when a runtime bound is reached.
+///
+/// Inspired by Tokio's `yield_now()`.
+struct YieldNow {
+    yielded: bool,
+}
 
-enum BounceInner<'a> {
-    Done(UntypedRetVal),
-    More(BoxFuture<'a, ResumeVal>),
+impl YieldNow {
+    fn new() -> Self {
+        Self { yielded: false }
+    }
+}
+
+impl Future for YieldNow {
+    type Output = ();
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
+        if self.yielded {
+            Poll::Ready(())
+        } else {
+            self.yielded = true;
+            cx.waker().wake_by_ref();
+            Poll::Pending
+        }
+    }
 }
 
 impl InstanceHandle {
@@ -101,51 +117,20 @@ impl InstanceHandle {
     /// that use `Vmctx::block_on` and provides the trampoline that `.await`s those futures on
     /// behalf of the guest.
     ///
+    /// If `runtime_bound` is provided, it will also pause the Wasm execution and yield a future
+    /// that resumes it after (approximately) that many Wasm opcodes have executed.
+    ///
     /// # `Vmctx` Restrictions
     ///
     /// This method permits the use of `Vmctx::block_on`, but disallows all other uses of `Vmctx::
     /// yield_val_expecting_val` and family (`Vmctx::yield_`, `Vmctx::yield_expecting_val`,
     /// `Vmctx::yield_val`).
-    ///
-    /// # Blocking thread
-    ///
-    /// The `wrap_blocking` argument is a function that is called with a closure that runs the Wasm
-    /// program. Since Wasm may execute for an arbitrarily long time without `await`ing, we need to
-    /// make sure that it runs on a thread that is allowed to block.
-    ///
-    /// This argument is designed with [`tokio::task::block_in_place`][tokio] in mind. The odd type
-    /// is a concession to the fact that we don't have rank 2 types in Rust, and so must fall back
-    /// to trait objects in order to be able to take an argument that is itself a function that
-    /// takes a closure.
-    ///
-    /// In order to provide an appropriate function, you may have to wrap the library function in
-    /// another closure so that the types are compatible. For example:
-    ///
-    /// ```no_run
-    /// # async fn f() {
-    /// # let instance: lucet_runtime_internals::instance::InstanceHandle = unimplemented!();
-    /// fn block_in_place<F, R>(f: F) -> R
-    /// where
-    ///     F: FnOnce() -> R,
-    /// {
-    ///     // ...
-    ///     # f()
-    /// }
-    ///
-    /// instance.run_async("entrypoint", &[], |f| block_in_place(f)).await.unwrap();
-    /// # }
-    /// ```
-    ///
-    /// [tokio]: https://docs.rs/tokio/0.2.21/tokio/task/fn.block_in_place.html
-    pub async fn run_async<'a, F>(
+    pub async fn run_async<'a>(
         &'a mut self,
         entrypoint: &'a str,
         args: &'a [Val],
-        wrap_blocking: F,
-    ) -> Result<UntypedRetVal, Error>
-    where
-        F: for<'b> Fn(&mut (dyn FnMut() -> Result<Bounce<'b>, Error>)) -> Result<Bounce<'b>, Error>,
-    {
+        runtime_bound: Option<u64>,
+    ) -> Result<UntypedRetVal, Error> {
         if self.is_yielded() {
             return Err(Error::Unsupported(
                 "cannot run_async a yielded instance".to_owned(),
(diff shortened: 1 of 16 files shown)


lucet-runtime/src/c_api.rs
```diff
@@ -507,6 +508,15 @@ pub unsafe fn lucet_vmctx_yield(vmctx: &Vmctx, val: *mut c_void) -> *mut c_void
         .unwrap_or(std::ptr::null_mut())
 }
 
+/// TODO: do we need the `lucet_hostcall` macro invocation here, which performs interesting
+/// contortions related to panic-unwind catching, if this is a frequent invocation and we are
+/// performing only a stack swap / control transfer to the host and back?
```
iximeow, line 513: yes, but more because `lucet_hostcall!` also takes care of signalling the appropriate timeout mechanism. so without a `lucet_hostcall!` a guest timeout while yielded would SIGALRM the embedder!

lucet-runtime/tests/instruction_counting.rs
```diff
@@ -110,6 +110,93 @@ pub fn check_instruction_count() {
...
+        );
+
+        // There is some variance to this in the real world, because yields happen only at the ends
+        // of basic blocks, and there may be an off-by-one; so allow a small range.
+        let expected_yields = instruction_count / 1000;
+        let diff = if expected_yields > yields {
+            expected_yields - yields
+        } else {
+            yields - expected_yields
+        };
```
iximeow, line 191:
```suggestion
        let diff = (yields - expected_yields).abs();
```
? i think this ought to work, anyway.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -195,6 +195,9 @@ impl InstanceHandle {
                             ))
                         }
                     }
+                    RunResult::BoundExpired => {
+                        panic!("BoundExpired was not expected when no bound was set");
```
alexcrichton, line 199: I think here what you'll probably want to do is to add a new variant to `BounceInner` like `BounceInner::Pause`. Then below in the `match bounce` when it gets `Pause` you'll want to do something like [`tokio::task::yield_now`](https://docs.rs/tokio/0.2.9/tokio/task/fn.yield_now.html).

alexcrichton: Ah I see though this will probably also need a new argument to `run_async` to specify the bound itself!

cfallin: This worked, thanks!

lucet-runtime/tests/instruction_counting.rs
cfallin: Unsigned types, so alas I don't think this'll work... it seems there's even an open Rust issue to add something to the stdlib for this: rust-lang/rust#62111.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -195,17 +189,22 @@ impl InstanceHandle {
                             ))
                         }
                     }
+                    InternalRunResult::BoundExpired => Ok(Bounce::BoundExpired),
                 }
-            })?;
+            }?;
             match bounce {
```
alexcrichton, line 195: FWIW now that there's no `wrap_blocking` I think it should be fine to do all this processing of `Bounce` directly inline above instead of having it separated down here.

cfallin: Indeed, that's much cleaner; thanks.

lucet-runtime/src/c_api.rs
acfoltzer: This should be temporary, though, since the bounded execution mechanism should be able to replace the uses of `SIGALRM`. That shouldn't happen until a followup PR, so it's right to use `#[lucet_hostcall]` here

cfallin:
Just made a few tweaks to (hopefully) get a green CI again; recent changes in `wasmtime` mean we'll need to fix a few things in Lucet before merging this PR.

1. The `GuestMemory` API changed; @pchickey I think the impl of this trait in Lucet needs to be updated as well?
2. There were some build errors related to hostcall types; @fst-crenshaw this is I think related to your work in bytecodealliance/wasmtime#2419 -- will there be a Lucet update as part of that work too?

I'm happy to do either or both of these if needed, just let me know and I'll dig in further to understand the changes better :-) Thanks!

iximeow:
@cfallin i've cooked up a patch for point 1, and am figuring out point 2 at the moment - i expect we can piece those changes out with a wasmtime update independent of this PR landing, so we don't have to worry about tying up too much in one PR

lucet-wiggle/src/lib.rs
```diff
@@ -40,13 +40,22 @@ pub mod runtime {
         fn has_outstanding_borrows(&self) -> bool {
             self.bc.has_outstanding_borrows()
         }
-        fn is_borrowed(&self, r: Region) -> bool {
+        fn is_shared_borrowed(&self, r: Region) -> bool {
```
pchickey, line 43: this isnt quite correct - in addition to the GuestMemory api changing, the underlying borrow checker has evolved as well, and because it was copy-pasted in a variety of places it now comes from the `wiggle-borrow` crate. You can delete `lucet-wiggle/src/borrow.rs` and wire up `wiggle_borrow::BorrowChecker` here instead please

cfallin: Resolved by awortman in #614; thanks!

cfallin:
Rebased on top of #614; this should be ready for review now (thanks @awortman-fastly for the quick work on that PR!).

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -101,51 +117,20 @@ impl InstanceHandle {
     /// that use `Vmctx::block_on` and provides the trampoline that `.await`s those futures on
     /// behalf of the guest.
     ///
+    /// If `runtime_bound` is provided, it will also pause the Wasm execution and yield a future
+    /// that resumes it after (approximately) that many Wasm opcodes have executed.
```
acfoltzer, line 121: This is a bit unclear to me: if I'm reading correctly, this method is not returning a future, it's running a loop until the instance returns its final result, or an error occurs. The `runtime_bound` makes the `poll` method on the underlying, generated future return every time that many Wasm opcodes have been executed, right?

Sorry if this is splitting hairs, I just want to make sure I understand the intended behavior.

lucetc/src/compiler.rs
```diff
@@ -255,6 +255,22 @@ impl<'a> Compiler<'a> {
...
+    fn get_local_count(body: &FunctionBody, name: &str) -> Result<u32, Error> {
+        let error_mapper = |e| Error::FunctionTranslation {
+            symbol: name.to_string(),
+            source: Box::new(Error::from(e)),
+        };
+        Ok(body
+            .get_locals_reader()
+            .map_err(|source| error_mapper(source))?
+            .into_iter()
+            .map(|result| result.map_err(|source| error_mapper(source)))
```
acfoltzer, line 267: Readability tweak

```suggestion
            .map_err(error_mapper)?
            .into_iter()
            .map(|result| result.map_err(error_mapper))
```

lucetc/src/function.rs
```diff
@@ -22,26 +22,33 @@ pub struct FuncInfo<'a> {
     module_decls: &'a ModuleDecls<'a>,
     codegen_context: &'a CodegenContext,
     count_instructions: bool,
-    scope_costs: Vec<u32>,
+    scope_costs: Vec<(u32, bool)>,
```
acfoltzer, line 25: I'd prefer to see this as a struct with named fields, rather than a tuple. Seeing this `bool` here alone is a head-scratcher

lucetc/src/function.rs
```diff
@@ -22,26 +22,33 @@ pub struct FuncInfo<'a> {
...
         Self {
             module_decls,
             codegen_context,
             count_instructions,
-            scope_costs: vec![0],
+            scope_costs: vec![(0, false)],
             vmctx_value: None,
             global_base_value: None,
             runtime_funcs: HashMap::new(),
+            // variable indices correspond ot Wasm bytecode's index space,
```
acfoltzer, line 48:
```suggestion
            // variable indices correspond to Wasm bytecode's index space,
```

lucetc/src/function.rs
```diff
@@ -89,65 +96,139 @@ impl<'a> FuncInfo<'a> {
...
+    }
+
+    fn load_instr_count(&mut self, builder: &mut FunctionBuilder<'_>) {
+        let (addr, instr_count_offset) = self.get_instr_count_addr_offset(builder);
+        let trusted_mem = ir::MemFlags::trusted();
+
+        // Do the equivalent of:
+        //
+        //    let instruction_count_ptr: &mut u64 = vmctx.instruction_count;
+        //    let instruction_count: u64 = *instruction_count_ptr;
```
acfoltzer, line 119:
```suggestion
        //    let instruction_count_ptr: &mut i64 = vmctx.instruction_count;
        //    let instruction_count: i64 = *instruction_count_ptr;
```

lucetc/src/function.rs
```diff
@@ -89,65 +96,139 @@ impl<'a> FuncInfo<'a> {
...
+        builder.def_var(self.instr_count_var, cur_instr_count);
+    }
+
+    fn save_instr_count(&mut self, builder: &mut FunctionBuilder<'_>) {
+        let (addr, instr_count_offset) = self.get_instr_count_addr_offset(builder);
+        let trusted_mem = ir::MemFlags::trusted();
+
+        // Do the equivalent of:
+        //
+        //    let instruction_count_ptr: &mut u64 = vmctx.instruction_count;
```
acfoltzer, line 134:
```suggestion
        //    let instruction_count_ptr: &mut i64 = vmctx.instruction_count;
```

acfoltzer (changes requested):
Thanks so much for getting this implemented! It's going to make a huge difference for async Lucet embedders.

I have a handful of suggestions here but nothing fundamental.

acfoltzer:
Ah, it occurred to me that we also need a version of `Instance::run_start()` that has the same bounded time properties. @cfallin would you be able to add this before we merge?

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -101,51 +117,20 @@ impl InstanceHandle {
...
-    /// [tokio]: https://docs.rs/tokio/0.2.21/tokio/task/fn.block_in_place.html
-    pub async fn run_async<'a, F>(
+    pub async fn run_async<'a>(
         &'a mut self,
         entrypoint: &'a str,
         args: &'a [Val],
-        wrap_blocking: F,
-    ) -> Result<UntypedRetVal, Error>
-    where
-        F: for<'b> Fn(&mut (dyn FnMut() -> Result<Bounce<'b>, Error>)) -> Result<Bounce<'b>, Error>,
```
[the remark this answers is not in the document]
cfallin: Indeed!

lucet-runtime/lucet-runtime-internals/src/future.rs
cfallin: Yes, that's right; we await on the `YieldNow` future, which results in the `async fn`'s lowering containing a state in which it has returned from `poll()` and will resume execution on the next `poll()`.

lucetc/src/compiler.rs
cfallin: Done!

lucetc/src/function.rs
cfallin: Done.

lucetc/src/function.rs
cfallin: Done.

lucetc/src/function.rs
cfallin: Yup, that makes it much clearer, thanks!

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -412,6 +419,30 @@ impl RunResult {
...
+
+impl InternalRunResult {
+    pub(crate) fn unwrap(self) -> Result<RunResult, Error> {
+        match self {
+            InternalRunResult::Normal(result) => Ok(result),
+            InternalRunResult::BoundExpired => Err(Error::InvalidArgument(
+                "should not have had a runtime bound",
+            )),
+        }
+    }
```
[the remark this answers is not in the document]
cfallin: Done.

cfallin:
Updated, and added the async `Instance::run_async_start()` as well.

acfoltzer (approved):
Thank you for the fast work on this!

benaubin:
Curious - would it be possible to expose these APIs publicly?

cfallin:
@benaubin `run_async()` is a public method on `InstanceHandle` already -- or did you mean some other part of the API?

benaubin:
@cfallin I'd love to have access to the `InternalRunResult` without using `run_async` in order to set an approximate CPU usage limit.

cfallin:
@benaubin I see -- the `InternalRunResult` is an implementation detail (it's how the yield loop in `run_async()` is made aware of time-bound yields), so I don't think it makes sense to expose that type as-is.

If you are not also using the value-yield functionality in hostcalls, you can implement your use-case with `run_async()` as-is, I think: manually inspect the returned future and consider the result to be a timeout if it is not ready after the first invocation.

Otherwise, the right thing to do would be to have a separate `run_with_timeout()` API, I think: this is semantically different than the existing run methods, all of which run to completion (possibly with yields) and so its type needs to be different.

If the latter is what you need, I'm happy to review a PR! Unfortunately I don't have time to work on this myself at the moment though.

benaubin:
Happy to do a PR. Renaming `InternalRunResult` to `BoundedRunResult`and exposing it would provide a perfect `run_with_timeout` API as-is, without adding extra bloat to the library.

benaubin:
Oh, I see what you mean! There would need to be a `run` method that returns `BoundedRunResult`. I'll add a `run_bounded` method.

cfallin:
@benaubin One other thing that I should probably note also, is that longer-term, the plan is to merge Lucet's functionality into Wasmtime (see [blog post from Bytecode Alliance](https://bytecodealliance.org/articles/1-year-update) and [another blog post on the merging efforts](https://www.fastly.com/blog/how-lucet-wasmtime-make-stronger-compiler-together)). There are active efforts to make this happen, e.g. async support in bytecodealliance/rfcs#2 and fast instance allocation in bytecodealliance/rfcs#5. In the meantime Lucet is absolutely supported, and as mentioned I'm happy to review a PR; but this may be relevant for your planning :-)

benaubin:
Thanks for the heads up! We saw that, but currently want lucet's `Send` instances. Our host call surface is purposefully minimal (instances mostly communicate with an external API server) - which should make switching less painful (and sandboxing slightly easier).

benaubin:
@cfallin I've put an initial draft PR together (#625). Adds a few `_bounded` methods, and replaces `InternalRunResult` with an additional Error variant.

benaubin:
@cfallin On second thought, you were probably right that the `InternalRunResult` should remain a private implementation detail. I refactored the `run_async_impl` into a custom Future implementation, which gives the same benefit as exposing the bounded runtime apis, without complicating the API surface.

PR #615, opened 2020-12-03
use wasi sdk 12

pchickey:
https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-12


tschneidereit (approved):
I only skimmed the test changes, so if you feel they should get more scrutiny, then a second review might make sense. Otherwise, r=me!

PR #616, opened 2020-12-07
Add new x64 backend as lucetc option, and add to CI runs.

cfallin:
The new x86-64 backend in Cranelift is now able to run all Lucet tests,
and has shown better performance (compile-time and run-time) than the
current backend. This PR adds a Cargo feature to lucetc to build with
the new backend, and modifies the Makefile and CircleCI config to run
tests with the new backend as well, in order to ensure that it continues
to work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,14 +56,14 @@ jobs:
     steps:
       - full_ci:
           cache_target: true
-          lucetc_features: new-x64-backend
+          new_cranelift_backend: true
 
   new_backend_ci_main:
     executor: linux
     steps:
       - full_ci:
           cache_target: false
-          lucetc_features: new-x64-backend
+          new_cranelift_backend: true
 
 workflows:
   build:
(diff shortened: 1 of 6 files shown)


pchickey (dismissed):

cfallin:
This seems to have fixed the CircleCI yaml issue; but the force-push caused review comment(s) from @pchickey to become inaccessible it seems (sorry!).

cfallin:
The new jobs aren't building with the feature flag in `test-full`; refactoring things a bit in the Makefile to make this happen.

Question: how reasonable would it be to run at least some of the tests on Rust nightly? I think we may need `-Zpackage-features` to specify a particular crate's features (`lucetc/new-x64-backend`) while testing all crates, unless someone else knows a better way!

cfallin:
Tests on new backend regressed due to bytecodealliance/wasmtime#2463, I think, in the stack-limit check code.

pchickey:
I'm fine with running additional tests on nightly, but if there is a way to get new-backend testing on stable, that is preferable, since that is how it will eventually be deployed.

pchickey (approved):

PR #617, opened 2020-12-17
Add differential fuzzing between old and new Cranelift backends.

cfallin:
This PR adds a fuzz target runnable via `cargo-fuzz` (with `libFuzzer`)
that builds an arbitrary Wasm module, produced by `wasm-smith`, with
`lucetc` using both the old and new Cranelift x86-64 backends. It then
instantiates both and invokes the start function in both, comparing the
global values and heap contents after the call returns.

The intent of this differential fuzzing is to validate that the new
backend works as well as the old one for the purposes of Lucet.

This change pulls in a new version of wasmtime and fixes a few small
issues (panics that should be errors) that the fuzzing immediately
discovered.


cfallin:
Should be ready for review now (sorry for churn) -- verified that this is doing interesting things locally, with coverage across old/new backend.

pchickey (approved):

PR #620, opened 2021-01-08
macOS: add LC_BUILD_VERSION to shared objects produced by lucetc

jedisct1:
This adds a new TargetVersion parameter that can be used to specify system-specific versions to compile for.

The command-line tool gets new `--sdk-version` and `--min-os-version` switches, and we default to the current SDK major version and the oldest OSX version it supports.


pchickey (dismissed):
Thank you Frank!

pchickey (approved):

PR #621, opened 2021-01-10
Update the object crate to the same version as wasmtime

jedisct1:
Having different versions of the same dependency is not a good idea, and `lucet-module` couldn't see the `lucet_module_data` symbol any more on macOS.

This repairs signatures on macOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,23 +161,11 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.7.3"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


pchickey (approved):

pchickey:
I wonder if we should get `cranelift-object` to re-export `object` (e.g. `pub use object::*`) so that we can track just one dep here.

jedisct1:
Sure! Re-exporting dependencies would indeed help a lot with such cases.

pchickey:
turns out it already does, ill follow up with a PR that swaps to it

pchickey:
turns out that the cranelift-object re-export does not include the `read` cargo feature, which are what we are actually using here. Rather than make upstream expose just the right feature to our liking (read_core and elf, to avoid the dupe wasmparser dep), we should instead just keep the dep.

PR #623, opened 2021-01-12
📨 Make `InstanceHandle::run_async_start()` future impl `Send`

acfoltzer:
Also adds a test to ensure that both `run_async()` and `run_async_start()` impl `Send`.


cfallin (approved):
Thanks for this!

cfallin:
(And also: the closure capture rules are very subtle; sorry about the troubles with the original version!)

acfoltzer:
No worries at all. This is definitely not the first time this has happened where we only realize something's not `Send` when we try to integrate downstream.

PR #625, opened 2021-01-15
Expose bounded runtime

benaubin:
Exposes the bounded runtime API added in #612 publicly.


benaubin:
Hmm, looks like the only test that fails tests UFFD on MacOS. That seems correct - UFFD doesn't exist on Mac. I'm added `#[cfg(target_os = "linux")]` in b6b7e6b

benaubin:
Closing in favor of #626

PR #626, opened 2021-01-15
Refactor run_async to use custom Future implementation 

benaubin:
This PR replaces the `async fn run_async_internal()` with a custom implementation of Future.

Advantages:

- Significantly less unsafe code than the original implementation of run_async (just one line, used for pinning the future)
a much sounder execution model: the future is executed from within the guest! When host execution is desired tokio::spawn or similar works from within block_on
- Support for defining async fn hostcalls without ever manually calling block_on
- Near-zero overhead for polling an immediately-ready future (only the cost of an Arc.clone()). For example, a future like `async { 5 }` would immediately resolve to 5, without requiring a yield to the host.
- Support for regular yielding from an async execution, even from within block_on
- A defined type for the future returned by `run_async`, which makes it easier for embedders to manage the execution, such as by changing the execution bound between polls.

Semantically, guests behave _exactly_ as normal async functions would, where block_on is equivalent to calling await.

It's now possible to declare an async hostcall like an async function:

```rust
#[lucet_hostcall]
#[no_mangle]
pub async fn hostcall_async(vmctx: &Vmctx) -> u32 {
    for i in 0..6 {
        YieldingFuture { times: 2 }.await
    }

    return 0;
}
```

I also a added new `try_block_on` method, which allows an embedded to define fallback behavior for when an instance is not run from within an async execution context.

Additionally, I removed the inst_count_bound argument in `run_async` and `run_async_start` (added in #612, API change is not released) and replaced it with a method and field on the RunAsync future. This is more flexible for users that need the ability to configure a CPU bound (it lets a user change the CPU bound in-between any poll) while being less verbose (and backward compatible with the old api) for users without that need.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/future.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/future.rs
@@ -18,34 +18,60 @@ pub struct AsyncContext {
 
 const DEFAULT_INST_COUNT_BOUND: u64 = i64::MAX as u64;
 
+/// A value representing that the guest instance yielded because it was blocked on a future.
+///
+/// In the future, we could provide a value from the guest that can be accessed before resuming the future,
+/// such as if we wanted to do something from the host context.
 struct AsyncYielded;
+
+/// Providing the private `AsyncResume` as a resume value certifies that
+/// RunAsync upheld the invarriants necessary to safely resume the instance.
 struct AsyncResume;
+
+/// An error representing a failure of `try_block_on`
+#[doc(hidden)]
+pub enum BlockOnError {
+    NeedsAsyncContext,
+}
+
+impl From<BlockOnError> for TerminationDetails {
+    fn from(err: BlockOnError) -> Self {
+        match err {
+            BlockOnError::NeedsAsyncContext => TerminationDetails::BlockOnNeedsAsync,
+        }
+    }
+}
 
 impl Vmctx {
     /// Block on the result of an `async` computation from an instance run by `Instance::run_async`.
     ///
-    /// Lucet hostcalls are synchronous `extern "C" fn` functions called from WebAssembly. In that
-    /// context, we cannot use `.await` directly because the hostcall is not `async`. While we could
-    /// block on an executor using `futures::executor::block_on` or
-    /// `tokio::runtime::Runtime::block_on`, that has two drawbacks:
+    /// While this method is supported and part of the public API, it's easiest to define the hostcall
+    /// function itself as async. The `#[lucet_hostcall]` macro simply calls this function.
     ///
-    /// - If the Lucet instance was originally invoked from an async context, trying to block on the
-    ///   same runtime will fail if the executor cannot be nested (all executors we know of have this
-    ///   restriction).
-    ///
-    /// - The current OS thread would be blocked on the result of the computation, rather than being
-    ///   able to run other async tasks while awaiting. This means an application will need more
-    ///   threads than otherwise would be necessary.
-    ///
-    /// Instead, this block_on operator is designed to work only when called within an invocation
-    /// of [`Instance::run_async`]. When a future needs to be polled, the instance will yield and
-    /// RunAsync will return control to the async executor. When the future is ready to be polled/make progress,
-    /// it will be polled from within the guest context.
+    /// There's no performance penalty for doing so: futures that are immediately ready without waiting
+    /// don't require a context switch, just like using `.await`.
     ///
     /// Note:
     /// - This method may only be used if `Instance::run_async` was used to run the VM,
     ///   otherwise it will terminate the instance with `TerminationDetails::BlockOnNeedsAsync`.
-    pub fn block_on<R>(&self, mut f: impl Future<Output = R>) -> R {
+    #[doc(hidden)]
+    #[inline(always)]
+    pub fn block_on<R>(&self, f: impl Future<Output = R>) -> R {
+        match self.try_block_on(f) {
+            Ok(res) => res,
+            Err(err) => panic!(TerminationDetails::from(err)),
+        }
+    }
+
+    /// Block on the result of an `async` computation from an instance run by `Instance::run_async`.
+    ///
+    /// The primary reason you may want to use `try_block_on` manually is to provide a fallback
+    /// implementation for if your hostcall is called from outside of an asynchronous context.
+    ///
+    /// If `Instance::run_async` is not being used to run the VM, this function will return
+    /// `Err(BlockOnError::NeedsAsyncContext)`.
+    #[doc(hidden)]
+    pub fn try_block_on<R>(&self, mut f: impl Future<Output = R>) -> Result<R, BlockOnError> {
         // We pin the future to the stack (a requirement for being able to poll the future).
         // By pinning to the stack instead of using `Box::pin`, we avoid heap allocations for immediately-ready futures.
         //
@@ -61,9 +87,7 @@ impl Vmctx {
                 } => cx,
                 State::Running {
                     async_context: None,
-                } => {
-                    panic!(TerminationDetails::BlockOnNeedsAsync)
-                }
+                } => return Err(BlockOnError::NeedsAsyncContext),
                 _ => unreachable!("Access to vmctx implies instance is Running"),
             };
 
@@ -71,11 +95,16 @@ impl Vmctx {
             let mut cx = Context::from_waker(&cx.waker);
 
             match f.as_mut().poll(&mut cx) {
-                Poll::Ready(ret) => return ret,
+                Poll::Ready(ret) => return Ok(ret),
                 Poll::Pending => {
                     // The future is pending, so we need to yield to the async executor
                     self.yield_impl::<AsyncYielded, AsyncResume>(AsyncYielded, false, false);
 
+                    // Providing the private `AsyncResume` as a resume value certifies that
+                    // RunAsync upheld the invarriants necessary for us to avoid a borrow check.
+                    //
+                    // If we resume with any other value, the instance may have been modified, and it is
+                    // unsound to resume the instance.
                     let AsyncResume = self.take_resumed_val::<AsyncResume>();
                 }
             }
@@ -86,21 +115,36 @@ impl Vmctx {
 impl InstanceHandle {
     /// Run a WebAssembly function with arguments in the guest context at the given entrypoint.
     ///
-    /// This method is similar to `Instance::run()`, but allows the Wasm program to invoke hostcalls
-    /// that use `Vmctx::block_on` and provides the trampoline that `.await`s those futures on
-    /// behalf of the guest.
+    /// This method is similar to `Instance::run()`, but allows the Wasm program to invoke async hostcalls
+    /// and provides the trampoline that `.await`s those futures on behalf of the guest.
+    ///
+    /// To define an async hostcall, simply add an `async` modifier to your hostcall:
+    ///
+    /// ```ignore
+    /// #[lucet_hostcall]
+    /// #[no_mangle]
+    /// pub async fn hostcall_async(vmctx: &Vmctx) {
+    ///    foobar().await
+    /// }
+    /// ```
+    ///
+    /// See `[RunAsync]` for details.
     ///
     /// If `runtime_bound` is provided, it will also pause the Wasm execution and yield a future
     /// that resumes it after (approximately) that many Wasm opcodes have executed.
     ///
     /// # `Vmctx` Restrictions
     ///
-    /// This method permits the use of `Vmctx::block_on`, but disallows all other uses of `Vmctx::
+    /// This method permits the use of async hostcalls, but disallows all other uses of `Vmctx::
     /// yield_val_expecting_val` and family (`Vmctx::yield_`, `Vmctx::yield_expecting_val`,
     /// `Vmctx::yield_val`).
     pub fn run_async<'a>(&'a mut self, entrypoint: &'a str, args: &'a [Val]) -> RunAsync<'a> {
         let func = self.module.get_export_func(entrypoint);
-        self.run_async_internal(func, args)
+
+        match func {
+            Ok(func) => self.run_async_internal(func, args),
+            Err(err) => self.run_async_failed(err),
+        }
     }
 
     /// Run the module's [start function][start], if one exists.
@@ -117,41 +161,80 @@ impl InstanceHandle {
     /// runtime between async future yields (invocations of `.poll()` on the
     /// underlying generated future) if `runtime_bound` is provided. This
     /// behaves the same way as `Instance::run_async()`.
+    ///
+    /// Just like `Instance::run_start()`, hostcalls, including async hostcalls,
+    /// cannot be called from the instance start function.
+    ///
+    /// The result of the `RunAsync` future is unspecified, and should not be relied on.
     pub fn run_async_start<'a>(&'a mut self) -> RunAsync<'a> {
-        let func = if self.is_not_started() {
-            self.module
-                .get_start_func()
-                // Invariant: can only be in NotStarted state if a start function exists
-                .map(|start| start.expect("NotStarted, but no start function"))
-        } else {
-            Err(Error::StartAlreadyRun)
+        if !self.is_not_started() {
+            return self.run_async_failed(Error::StartAlreadyRun);
+        }
+
+        let func = match self.module.get_start_func() {
+            Ok(start) => start.expect("NotStarted, but no start function"), // can only be in NotStarted state if a start function exists,
+            Err(err) => return self.run_async_failed(err),
         };
 
         self.run_async_internal(func, &[])
     }
 
-    fn run_async_internal<'a>(
-        &'a mut self,
-        func: Result<FunctionHandle, Error>,
-        args: &'a [Val],
-    ) -> RunAsync<'a> {
-        let state = match func {
-            Ok(func) => RunAsyncState::Start(func, args),
-            Err(err) => RunAsyncState::Failed(Some(err)),
-        };
-
+    /// Returns a `RunAsync` that will asynchronously execute the guest instnace.
+    fn run_async_internal<'a>(&'a mut self, func: FunctionHandle, args: &'a [Val]) -> RunAsync<'a> {
         RunAsync {
             inst: self,
             inst_count_bound: DEFAULT_INST_COUNT_BOUND,
-            state,
+            state: RunAsyncState::Start(func, args),
+        }
+    }
+
+    /// Returns a `RunAsync` that will immediately fail with the given error, without executing the guest instance.
+    fn run_async_failed<'a>(&'a mut self, err: Error) -> RunAsync<'a> {
+        RunAsync {
+            inst: self,
+            inst_count_bound: DEFAULT_INST_COUNT_BOUND,
+            state: RunAsyncState::Failed(err),
         }
     }
 }
 
+/// A future implementation that enables running a guest instance which can call async hostcalls.
+///
+/// Lucet hostcalls are synchronous `extern "C" fn` functions called from WebAssembly. In that
+/// context, we cannot use `.await` directly because the hostcall is not `async`. While we could
+/// block on an executor such as `futures::executor::block_on`, that would block the OS thread,
+/// preventing us from running other async tasks while awaiting. This means an application will need more
+/// threads than otherwise would be necessary.
+///
+/// `RunAsync` allows the guest to call async hostcalls just as if the guest had called the async function
+/// and immediately `.await`ed it.
+///
+/// To define a async hostcall, simply add the async modifier to a hostcall definition:
+///
+/// ```ignore
+/// #[lucet_hostcall]
+/// #[no_mangle]
+/// pub async fn hostcall_async(vmctx: &Vmctx) {
+///    foobar().await
+/// }
+/// ```
+///
+/// Note: Async hostcalls may only be used if `Instance::run_async` was used to run the VM,
+/// otherwise it will terminate the instance with `TerminationDetails::BlockOnNeedsAsync`.
+///
+/// Behind the scenes, lucet polls the future from within the guest execution context. If the future is not immediately ready,
+/// the instance will yield and return control to the async executor. Later, when the future is ready to make progress,
+/// the async executor will return to the guest context, where lucet will poll the future to completion.
+///
+/// Just like `.await`, there is no overhead for futures that are immediately ready (such as `async { 5 }`).
+///
+/// For async hostcalls that may yield to the async executor many times, it's recommended that you use `tokio::spawn`,
+/// or the equivalent from your async executor, which will spawn the task to be run from the host execution context.
+/// This avoids the overhead of context switching into the guest execution context every time the future needs to make progress.
 pub struct RunAsync<'a> {
     inst: &'a mut InstanceHandle,
     state: RunAsyncState<'a>,
-    /// The instance count bound. Can be changed at any time, taking effect on the next guest entry
+    /// The instance count bound. Can be changed at any time, taking effect on the next entry to the guest execution context
     pub inst_count_bound: u64,
 }
 
@@ -165,9 +248,15 @@ impl<'a> RunAsync<'a> {
 
 enum RunAsyncState<'a> {
     Start(FunctionHandle, &'a [Val]),
-    Blocked,
-    Yielded,
-    Failed(Option<Error>),
+    /// The instance is currently blocked on a future.
+    ///
+    /// We keep the async yielded around - although the value of AsyncYielded isn't used for
+    /// anything, there's not much additional overhead in keeping it around, and it gives us
+    /// the option to pass data from the yield and potentially add other types of execution
+    /// in the future (such as bringing back host-context future execution).
+    BlockedOnFuture(Box<AsyncYielded>),
+    BoundExpired,
+    Failed(Error),
 }
 
 impl<'a> Future for RunAsync<'a> {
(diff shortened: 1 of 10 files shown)


benaubin:
In case it's useful, here's one way to implement a CPU time limit with this refactor. 

```
use std::time::Duration;
use cpu_time;

pub struct RunAsyncWithCPULimit<'a> {
    run_async: RunAsync<'a>,
    pub cpu_time_limit: Duration,
    pub cpu_time_used: Duration,
}

struct CPULimitExceeded;

impl<'a> Future for RunAsyncWithCPULimit<'a> {
    type Output = (
        Result<Result<UntypedRetVal, Error>, CPULimitExceeded>,
        Duration,
    );

    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        let start = cpu_time::ThreadTime::now();

        let poll = Pin::new(&mut self.run_async).poll(cx);

        self.cpu_time_used += start.elapsed();

        if self.cpu_time_used > self.cpu_time_limit {}

        poll
    }
}
```

Feel free to use this example however you'd like (I'm hereby releasing this example under the [CC-0 license](https://creativecommons.org/share-your-work/public-domain/cc0/))

cfallin (changes requested):
Hi @benaubin,

Thanks for this PR; however, I'm not sure I understand the reasoning behind some of the changes below.  If I'm understanding correctly, the changes here create an explicit named type for the future, moving the existing `async fn` into a custom trait impl with an explicit state machine.

It might help me to understand the need for these if you could summarize why one can't implement the wall-clock bound example above by wrapping the future returned by the current API. For example, I *think* one could write an `add_timeout<F: Future...>(inner: F) -> TimeoutFuture<F>` and define `TimeoutFuture` to store an `F`, without requiring a named future-state struct on the Lucet side. This is largely like your example above, but with the struct type param to make it work without Lucet changes. Is there a reason this wouldn't work?

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -150,91 +122,120 @@ impl InstanceHandle {
...
+        self.run_async_internal(func, &[])
     }
 
-    /// Shared async run-loop implementation for both `run_async()` and
-    /// `run_start_async()`.
-    async fn run_async_internal<'a>(
+    fn run_async_internal<'a>(
         &'a mut self,
-        func: FunctionHandle,
+        func: Result<FunctionHandle, Error>,
```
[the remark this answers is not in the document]
benaubin: Oh, yeah, this was a bit odd to write. This is to make the future return the error rather than returning the error in a result along with the future.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -150,91 +122,120 @@ impl InstanceHandle {
...
-        &'a mut self,
-        runtime_bound: Option<u64>,
-    ) -> Result<(), Error> {
-        if !self.is_not_started() {
-            return Err(Error::StartAlreadyRun);
-        }
-        let start = match self.module.get_start_func()? {
-            Some(start) => start,
-            None => return Ok(()),
+    pub fn run_async_start<'a>(&'a mut self) -> RunAsync<'a> {
```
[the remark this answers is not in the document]
benaubin: Hmm, I mean it's just `run_async("entrypoint").bound_inst_count(count).await`, rather than `run_async("entrypoint", Some(4)).await`. Shouldn't make things much more fiddly, and, in my opinion, that's a nicer API than having to pass through an optional (closer to a builder-style API). Especially if other runtime options are added in the future, in my opinion, this API is much easier to use.

lucet-runtime/src/c_api.rs
```diff
@@ -553,6 +553,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(target_os = "linux")]
```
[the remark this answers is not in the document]
benaubin: Will do.

benaubin:
> It might help me to understand the need for these if you could summarize why one can't implement the wall-clock bound example above by wrapping the future returned by the current API. For example, I think one could write an add_timeout<F: Future...>(inner: F) -> TimeoutFuture<F> and define TimeoutFuture to store an F, without requiring a named future-state struct on the Lucet side. This is largely like your example above, but with the struct type param to make it work without Lucet changes. Is there a reason this wouldn't work?

To be honest, I completely forgot that you could use generics for this. However, futures created by async/await are `Unpin` - so it would be more difficult to write (although theoretically possible).

The primary thing that is possible to do with this API but not an `async fn` is adjusting the instruction bound between polls. 

Additionally, the custom Future implementation reduces major sharp edges:

- If an instance polls an immediately ready future, the old API would reset the instruction bound and continue running without yielding. While possible to fix within the `async` block, this is a sharp edge that the async fn method hides. A custom implementation makes it much more clear how lucet and the future executor interact.
- If you `run_async` on an instance where `bound_expired()`, the old implementation would resume the instance rather than fail when attempting to run the instance from the given entry point. Solving this requires an additional state machine - another sharp edge removed by the custom future implementation.

In the future, this API makes other features possible:

- It would be possible to add an API that allows yielding from a `run_async` context.
- A fully-synchronous API that does not require a `std::task::Waker` that can drive the `RunAsync`, except when blocking.

lucet-runtime/lucet-runtime-internals/src/future.rs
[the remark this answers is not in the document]
benaubin: Caveat: for the `Box::pin`/custom executor tests, the API usage is indeed more fiddly. That's because calling `Box::pin` on a mutable reference that isn't assigned to a variable causes the lifetime to go out of scope. `.await` does not have the same issue.

The only place this would be fiddly would be if the future returned by `run_async` needs to be moved - something like `tokio::spawn` or similar. But, notably, `run_async` requires a `&mut self` reference, which (if my understanding is correct) limits the lifetime and prevents usage (if my understanding is incorrect, at least prevents usage for any reasonable usecase) with `tokio::spawn` without a custom async block anyways.

benaubin:
@cfallin Hey Chris! I just pushed a new commit that takes advantage of the custom Future implementation to do the first poll of a future without yielding. For futures that are immediately ready, this implementation requires no context switch and no heap allocations - similar to using `.await`.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1203,7 +1205,12 @@ impl Instance {
                 || self.state.is_yielded()
                 || self.state.is_bound_expired()
         );
-        self.state = State::Running { async_context };
+        
+        self.state = State::Running {
+            // SAFETY: We have to lie about the lifetime of async_context.
+            // However, State::Running will only last for as long as this function's lifespan.
+            async_context: unsafe { transmute(async_context) }
```
benaubin, line 1212: I'm going to make the generic types here explicit, to prevent accidental sharp edges if the API evolves.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -49,65 +54,67 @@ impl Vmctx {
...
+        // By overriding the variable named `f`, it is impossible to access f again, except through the pinned reference.
+        let mut f = unsafe { Pin::new_unchecked(&mut f) };
 
-impl YieldNow {
-    fn new() -> Self {
-        Self { yielded: false }
-    }
-}
+        // Poll the future using the std::task::Context, or die if we aren't async
+        match &mut unsafe { self.instance_mut() }.state {
```
benaubin, line 82: Note: this is probably unsound, and needs to be a refcell - re-entrant calls to block_on would cause undefined behavior (two mutable references to state)

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -49,65 +54,67 @@ impl Vmctx {
     ///
     /// Note that this method may only be used if `Instance::run_async` was used to run the VM,
     /// otherwise it will terminate the instance with `TerminationDetails::BlockOnNeedsAsync`.
-    pub fn block_on<'a, R>(&'a self, f: impl Future<Output = R> + Send + 'a) -> R
+    pub fn block_on<R>(&self, f: impl Future<Output = R> + Send) -> R
```
benaubin, line 57: Now that block on is zero-overhead when polling is unnecessary, I'd imagine that it would be possible to make `async fn` hostcalls possible by changing the macro. I'll look into this later on today.

benaubin: Nope, this wouldn't work, because it would be a sharp edge depending on if the future is within the guest execution context

cfallin:
@benaubin Sorry, I haven't gotten back to this as I've been somewhat short on time. To be honest, I still don't quite understand some of the details and reasoning here. For example, could you explain what you mean by: "takes advantage of the custom Future implementation to do the first poll of a future without yielding." -- first poll of which future? AFAICT in the current code, if the called Wasm function returns soon enough (before first bound expiration), we don't yield and require another poll, we just return a completed value. Or do you mean in some other situation?

From above, the only desired behavior I see that is not yet possible is to set the bound to a different amount on every poll. Everything else is possible with the current API. Is that correct?

Overall, I am hesitant to accept a full rewrite of the core async code, especially one that replaces an `async fn` with a custom future implementation and especially one that uses unsafe code, without a really really compelling reason. I'd be happy to look at a much smaller-scoped patch that allows for adjusting the bound on each poll, though, if you'd like. Sorry and thanks for the patience here!

benaubin:
No problem! So my latest changes have a huge advantage - an async function that is immediately ready doesn't require a yield to the executor (such as `async { 5 }`, would return 5 without needing a yield). I'm currently working on reducing the amount of unsafe code necessary to make this work.

benaubin:
@cfallin Just submitted a new commit that removes nearly all of the unsafe code, except one line (used to pin the future for polling). My changes also remove all of the unsafe code and `transmute`s required for the old trampoline-style `run_async`, and a change to the `lucet_hostcall` macro could enable writing `async fn` rather than needing to use `block_on` - there's nearly no overhead to `block_on` if the future does not require waking.

benaubin:
@cfallin Just pushed a new commit that adds support for defining an async hostcall just like an async function:

```
        #[lucet_hostcall]
        #[no_mangle]
        pub async fn hostcall_async(vmctx: &Vmctx, times: u32, times_double: u32) -> u32 {
            for i in 0..times {
                YieldingFuture { times: 2 }.await
            }

            return times * 2;
        }
```

Like Rust's `.await`, there is no overhead for a future that is immediately ready, and all execution of the future takes place from within the guest. To make this work, the executor polls the `RunAsync` future, which resumes the guest, which polls the future from within the guest execution context.

benaubin:
Rebased onto main, consolidated commits (happy to squash into a single commit if requested, but I'm separating out a few of them in case some of the changes are rejected).

@cfallin the changes should be ready for review. Advantages:

- significantly less unsafe code than the original implementation of `run_async` (just one line, used for pinning the future)
- a much sounder execution model: the future is executed from within the guest! When host execution is desired `tokio::spawn` or similar works from within block_on
- support for defining `async fn` hostcalls without ever manually calling `block_on`
- near-zero overhead for polling an immediately-ready future (only the cost of an `Arc.clone()`)
- support for yielding from an async execution, even from within `block_on`

Semantically, guests behave exactly like async functions, where calling an async hostcall is equivalent to immediately calling `await` on the `async fn` hostcall.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -150,91 +169,179 @@ impl InstanceHandle {
...
+///
+/// Behind the scenes, lucet polls the future from within the guest execution context. If the future is not immediately ready,
+/// the instance will yield and return control to the async executor. Later, when the future is ready to make progress,
+/// the async executor will return to the guest context, where lucet will poll the future to completion.
+///
+/// Just like `.await`, there is no overhead for futures that are immediately ready (such as `async { 5 }`).
+///
+/// For async hostcalls that may yield to the async executor many times, it's recommended that you use `tokio::spawn`,
+/// or the equivalent from your async executor, which will spawn the task to be run from the host execution context.
+/// This avoids the overhead of context switching into the guest execution context every time the future needs to make progress.
```
cfallin, line 269: A minor quibble, but a few of these lines are > 120 chars. Could you wrap at 80 or 100 (I usually do the former; I'm not sure if we have a global standard)?

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -150,91 +169,179 @@ impl InstanceHandle {
...
+/// #[no_mangle]
+/// pub async fn hostcall_async(vmctx: &Vmctx) {
+///    foobar().await
+/// }
+/// ```
+///
+/// Note: Async hostcalls may only be used if `Instance::run_async` was used to run the VM,
+/// otherwise it will terminate the instance with `TerminationDetails::BlockOnNeedsAsync`.
+///
+/// Behind the scenes, lucet polls the future from within the guest execution context. If the future is not immediately ready,
```
cfallin, line 261: s/lucet/Lucet/g

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -150,91 +169,179 @@ impl InstanceHandle {
...
+/// the instance will yield and return control to the async executor. Later, when the future is ready to make progress,
+/// the async executor will return to the guest context, where lucet will poll the future to completion.
+///
+/// Just like `.await`, there is no overhead for futures that are immediately ready (such as `async { 5 }`).
```
cfallin, line 265: Slight clarity improvement: "There is no additional return to the executor loop when an async hostcall returns a future that is immediately ready. This behavior is the same as implemented by the `.await` operator."

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -150,91 +169,179 @@ impl InstanceHandle {
...
+            inst: self,
+            inst_count_bound: DEFAULT_INST_COUNT_BOUND,
+            state: RunAsyncState::Failed(err),
+        }
+    }
+}
+
+/// A future implementation that enables running a guest instance which can call async hostcalls.
+///
+/// Lucet hostcalls are synchronous `extern "C" fn` functions called from WebAssembly. In that
```
cfallin, line 239: A few clarity improvements: this doc comment focuses solely on async hostcalls, but this future implements all bounded-execution yielding, right? I think that it would be better to describe the timeslicing and async-passing mechanism here in a bit more detail to avoid missing forest for trees -- maybe something like:

"A Future implementation that enables asynchronous execution of bounded slices of WebAssembly, and of underlying async hostcalls that it invokes.

To execute WebAssembly code in a system that uses an async runtime with an executor loop, we instrument the machine code to periodically yield back to the caller. The on-stack frames are preserved by stack switching: the Wasm code runs on a separate stack.

This allows for yielding from the Wasm guest itself, but we also wish to support yielding from `async` hostcalls. The Lucet hostcall interface is synchronous at the lowest level (C-ABI functions are invoked by generated machine code). To pass a future returned by an async hostcall back to the top-level executor, we store state on the `vmctx`, switch stacks back to the caller, and then return the hostcall's future wrapped by this future back to the caller.

To define an async hostcall..."

Clarity improvement: "At the machine-code level, Lucet hostcalls are `extern "C" fn` functions that are called synchronously from WebAssembly. To "pass through" the future from a hostcall up through WebAssembly to the top-level executor loop, we need to adapt the hostcalls to pass the future through vmctx state, and adapt the `Instance` run method to return a future that returns this state."

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -1,139 +1,158 @@
...
-struct YieldedFuture(BoxFuture<'static, ResumeVal>);
+const DEFAULT_INST_COUNT_BOUND: u64 = i64::MAX as u64;
+
+/// A value representing that the guest instance yielded because it was blocked on a future.
+struct AsyncYielded;
 
-/// A unique type for a boxed return value. The user never sees this type.
-pub struct ResumeVal(Box<dyn Any + Send + 'static>);
+/// Providing the private `AsyncResume` as a resume value certifies that
+/// RunAsync upheld the invarriants necessary to safely resume the instance.
```
cfallin, line 26: s/invarriants/invariants/

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -1,139 +1,158 @@
...
+                    // Check that the async context hasn't changed (this could happen if the instance yielded)
+                    match &self.instance().state {
+                        State::Running { async_context: Some(new_cx) } => {
+                            let same_waker = Arc::ptr_eq(&arc_cx, &new_cx) || arc_cx.waker.will_wake(&new_cx.waker);
+
+                            if !same_waker {
+                                // The AsyncContext changed on us. This is because the instance is running from a new RunAsync.
+                                // This probably happened because the instance yielded and was resumed up by `resume_async`.
+                                //
+                                // Poll the future again before yielding to the executor in order to register the new waker.
```
cfallin, line 110: The safety here is somewhat difficult to reason about for me. Is there a race where the underlying future still has the old waker, we've been resumed with the new one, and we might miss a wakeup? If not, why not (let's include the reasoning here)?

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -1,139 +1,158 @@
...
+    ///
+    /// The primary reason you may want to use `try_block_on` manually is to provide a fallback
+    /// implementation for if your hostcall is called from outside of an asynchronous context.
+    ///
+    /// If `Instance::run_async` is not being used to run the VM, this function will return
+    /// `Err(BlockOnError::NeedsAsyncContext)`.
+    #[doc(hidden)]
+    pub fn try_block_on<R>(&self, mut f: impl Future<Output = R>) -> Result<R, BlockOnError> {
+        // We pin the future to the stack (a requirement for being able to poll the future).
+        // By pinning to the stack instead of using `Box::pin`, we avoid heap allocations for immediately-ready futures.
```
cfallin, line 74: This is a nice improvement!

cfallin (commented):
@benaubin, thanks for the updates and sorry for the delay on the further review!

I think this is definitely going in the right direction, and given the reductions in unsafe code and after reading and understanding further the async-hostcall implementation, I'm inclined toward making sure this gets in.

I do have one pending uncertainty, though, below, regarding a possible race (it's possible you've thought through this already and if so I'd appreciate a clarifying comment :-) ).

I think that w.r.t. the async hostcall implementation we should also get any input that @pchickey might have. It looks reasonable to me but I may be missing some requirement here.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -562,19 +561,21 @@ impl Instance {
...
+            .resume_with_val_impl(Box::new(val), None, None)?
+            .unwrap())
     }
 
-    pub(crate) fn resume_with_val_impl<A: Any + 'static>(
+    pub(crate) fn resume_with_val_impl(
         &mut self,
-        val: A,
-        async_context: bool,
+        val: Box<dyn Any + 'static>,
```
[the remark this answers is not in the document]
benaubin: Mostly, this allows resuming the instance without double-boxing in the `RunAsync` future. The type of a value generally has to be known statically for a box to be allocated, so without this, we'd need to heap allocate twice (`Box<Box<ResumeVal>>`).

There's also no additional type check: because the expected type of the resume value is not known statically, the compiler can't remove the check in the original version. Further, the implementation of `dyn Any::is` is not const, and the implementation is nearly the same:

https://github.com/rust-lang/rust/blob/8fe989dd768f5dfdb0fc90933f3f74fa4579fefd/library/core/src/any.rs#L184-L195

Even then, the cost of a branch should be far less than a heap allocation. Especially because we're checking a developer invariant, branch misses should never happen in correct code. It might be worth adding `#[unlikely]` / the stable equivalent, but also, a branch miss is very little compared to the cost of re-entering an instance, and probably isn't worth the added code complexity.

lucet-runtime/lucet-runtime-internals/src/future.rs
benaubin: Yes, that issue exists. poll is synchronous, so by the time we have reached this point, it's certain if we are executing with a new AsyncContext, and possible that the future will awake (or has awoken) the old waker, which will mean we miss a wakeup.

Rather than try to prevent this from happening at all and writing our own waker (which is, bluntly: not fun, you basically have to manually implement Arc<dyn Fn() -> ()> with a custom vtable), we check if the waker changed from under us, and if so, we simply poll the future again, which registers the new waker, based on the API contract for Future::poll. If the future is Ready, there's no need for another wakeup, because we can make progress without yielding.

Also: Waking a waker unnecessarily will not cause unsafety or logic errors. In the worst case, the async executor may waste CPU time polling pending futures an extra time. It's also valid to hold onto a waker nearly indefinitely (again, they are roughly equivalent to Arc<dyn Fn()>, just with a manual vtable).

I'm realizing that I don't have a test that would cause this branch to execute. I'll write one. Luckily, it's easy behavior to force to happen (simply switch the async runtime between yield points)

lucet-runtime/lucet-runtime-internals/src/future.rs
benaubin: Thanks!

benaubin:
Awesome! Thanks, @cfallin. I'll make the changes tomorrow.

lucet-runtime/lucet-runtime-internals/src/instance.rs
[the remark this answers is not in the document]
benaubin: Also, the address of the typeid of `dyn A` is at a statically known location.

pchickey:
Thanks for working on this! Many of the futures bits of this are over my head, but the change to the macro to accept `async fn` is welcome. I had planned to do something just like that for compatibility with wasmtime :)

lucet-runtime/lucet-runtime-internals/src/instance.rs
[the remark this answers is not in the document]
cfallin: OK, yes, all of this makes sense -- thanks for the detailed explanation!

lucet-runtime/lucet-runtime-internals/src/future.rs
cfallin: (On scanning over this again today I noticed that the second "Clarity improvement" was an earlier version of the comment text I drafted that I forgot to delete. I think the first quote above is probably better.)

lucet-runtime/lucet-runtime-internals/src/future.rs
cfallin: OK, so I think I've convinced myself that this is sound. The potentially problematic case that I was considering earlier went something like: new executor polls our outer future; concurrently, underlying async IO or whatever completes, and the inner future we've wrapped notifies its (old) waker; then, we poll future to "reset to new waker". So we indeed do not set the waker in time, but we've also just polled the future, which is exactly what it wanted anyway. Either the underlying wakeup interleaves before our poll (in which case our poll just happens to be correct) or after our poll (in which case it'll pick up the new waker and trigger another wakeup later).

Could you add a comment here so that a later reader also confused about this gets it more easily? :-)

benaubin:
The changes have been made and the test for yielding from async has been added, although git is struggling to push (internet has been acting up this week due to the Texas blackouts).

cfallin (approved):
Thanks @benaubin -- this LGTM. I think it might be good to have one more sanity-check on the core future/waker logic -- @alexcrichton, would you mind deploying your async expertise here in a final check?

We'll also need to trigger the CircleCI results; usually this occurs when we push to a branch in this repo, but in this case you won't have permissions to do that. @pchickey / @iximeow / @acfoltzer, is there a way around this or will one of us need to push a branch and create a new PR in place of @benaubin?

pchickey:
I am not aware of a workaround to the CircleCI problem, unfortunately - I can push a branch and once it goes green i'll link to the status here and we can merge this PR.

benaubin:
Alternatively, I could edit the PR to merge into a non-main branch, someone could PR to test on Circle?

pchickey:
https://app.circleci.com/pipelines/github/bytecodealliance/lucet/430/workflows/36261c81-0205-41df-8626-2479a9261e76

pchickey:
you'll need to merge the latest `main` to fix the cargo audit error

benaubin:
@pchickey Done.

cfallin (approved):
re-r+'ing for my part after updates.

lucet-runtime/lucet-runtime-internals/src/future.rs
```diff
@@ -1,139 +1,194 @@
...
-        Self { yielded: false }
-    }
-}
+            // build an std::task::Context
+            let mut cx = Context::from_waker(&arc_cx.waker);
+
+            match f.as_mut().poll(&mut cx) {
+                Poll::Ready(ret) => return Ok(ret),
+                Poll::Pending => {
+                    // The future is pending, so we need to yield to the async executor
```
alexcrichton, line 126: Stylistically I might recommend de-indenting all this to the same level as the `match` by having the `Pending` arm just fall-through, that'd help rightward drift here a bit I think

alexcrichton (commented):
As a baseline I'm pretty unfamiliar with the internals of Lucet, and I'm unfortunately finding it difficult to disentangle "just the futures bits" from the rest of the Lucet embedding/runtime. In that sense I can say that after reading this nothing looks obviously wrong but I can't say with conviction that this all looks good to me. 

I'm basically trying to find parallels to the async implementation in wasmtime because I believe at a high-level the two async strategies are the same but I was finding it pretty difficult to draw the parallels because I don't understand enough of Lucet.

That being said it looks like this has gotten reviews from other folks, so that's probably good enough? If needed, though, I can set aside more time to learning more of Lucet first and then digging more into the changes here.

benaubin:
@alexcrichton It's indeed extremely similar to your wasmtime PR. Lucet runs every instance within a context switch, so it's very similar to how wasmtime implements fibers for async instances.

The primary difference in logic is that, in this PR, I avoid having to transmute the `std::task::Context` by cloning the `Waker` and creating a new context from inside the instance.

alexcrichton:
Ah ok that makes sense. While I believe that works well today the intention of `Context` was to abstract over possible extensions in the future, so in the limit of time that may not be an equivalent operation. For example `std` may one day expand the abilities of `Context` with new functionality, and certain features of `tokio` may take advantage of that, but that means that hostcalls using `tokio` futures may unexpectedly not have access to these features since we're recreating `Context`.

I don't think that there's any plans at this time to actually do this so it's a pretty meta-level concern though.

benaubin:
@alexcrichton That's what I figured. My [original implementation](https://github.com/bytecodealliance/lucet/pull/626/commits/e7efc5bc675b7f46c31e98428ca84a3e7990462b) passed the context to the instance with a transmute, but also required _a lot_ of unsafe code blocks propagated throughout lucet.

It wouldn't be too hard to change this if necessary, although to do so soundly, `RunAsync` would need to prevent "yielding" (Lucet instances can yield without blocking on a future and later be resumed from the host). Prior to this PR, that was the case - instances were not allowed to yield when in an async context.

After switching the implementation to clone a waker rather than transmute the lifetime, I also relaxed the restriction on yielding from an async instance.

For future compatibility's sake, it might be worth removing the ability to yield from an async context, or maybe marking it as "unstable".

alexcrichton:
I'm not really sure what it means to prevent yielding or be in or not be an async context myself, but I think it's a reasonable state of affairs to just create a new `Context` for now to prevent a lot of `unsafe` code. I suspect that any change to `Context` is years out at the earliest, if ever, so we'd have plenty of runway to update things if needed.

cfallin (approved):
r+ on latest commit.

pchickey's branch in #633 shows green CircleCI results for `4773acb`, one commit back from this branch's `HEAD`; the last commit is just an early-return out-denting refactor and passes all other CI.

Given that and given the discussion above with alexcrichton, I think we should be good to merge this. I'll go ahead and merge given the manually-verified CI above!

PR #627, opened 2021-02-06
merge latest wasmtime, fixup wasi integration and tests

pchickey:
* the `lucet-wasi` test suite now uses `cap-tempfile` most places! we needed a `Dir` so why not use the better api anyway?
* various errnos have changed in the test suite.
* lucet-wasi's getrusage test no longer makes sense - wasi doesnt pretend to support the process/thread timers anymore. So, I have deleted it.
* I renovated lucet-wasi's `fs.c` test to give unique failures depending where it fails, and isolated the particular behavior I was working on there into the `readdir.c` test
* raw-cpuid has been upgraded to 9.0.0 to resolve a RUSTSEC advisory. we also needed to bump the wasmtime submodule to get the fixed cranelift-native, which removes its dep on a vulnerable raw-cpuid

PR #628, opened 2021-02-09
Upgrade raw-cpuid to version 9

benaubin:
The earlier version has a security vulnerability, making `cargo audit` fail. Wasmtime will need to be upgraded as well in order for the CI checks to pass.


pchickey:
This requires a wasmtime bump to fix the dep that has been removed from cranelift-native. I'll take care of it in #627

PR #629, opened 2021-02-11
use rust 1.50

pchickey:


pchickey:
cargo audit failures will be resolved by #627

acfoltzer (approved):
:crab:

PR #630, opened 2021-02-17
include type of Provided termination details for debugging

iximeow:
an embedder of `lucet` will plausibly want to take some action on the `Provided` details if and when a hostcall hits `lucet_hostcall_terminate!`. to do this, they would have to `downcast` to a known type. if, through a bug, `lucet_hostcall_terminate!` is called with an unknown type, embedders can't get even the _name_ of that type to begin debugging!

at the point you have a `Box<dyn Any>`, i can't find a way to get a type name anyway. thankfully, the real type is available at the point a user calls `TerminationDetails::provide` (through `lucet_hostcall_terminate!` probably. hopefully!), so recording the type at that point is an API-preserving change. except that `TerminationDetails` itself changes. :grimacing:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 - Added `terminate_on_heap_oom` as an option for instances. This causes instances to terminate with an OOM-specific termination value rather than returning `-1` when a `memory.grow` instruction fails.
 
+- Lucet will no longer try to translate `wiggle::Trap` to primitives for `lucet_hostcall_terminate!`. Instead, the underlying `wiggle::Trap` is passed directly to the embedder.
+
 [start-function]: https://webassembly.github.io/spec/core/syntax/modules.html#syntax-start
 
 ### 0.6.1 (2020-02-18)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1102,6 +1102,7 @@ dependencies = [
  "thiserror",
  "tracing",
  "userfaultfd",
+ "wiggle",
 ]
 
 [[package]]
--- a/lucet-runtime/lucet-runtime-internals/Cargo.toml
+++ b/lucet-runtime/lucet-runtime-internals/Cargo.toml
@@ -12,6 +12,7 @@ edition = "2018"
 [dependencies]
 lucet-module = { path = "../../lucet-module", version = "=0.7.0-dev" }
 lucet-runtime-macros = { path = "../lucet-runtime-macros", version = "=0.7.0-dev" }
+wiggle = { path = "../../wasmtime/crates/wiggle", version = "0.22.0" }
 
 anyhow = "1.0"
 bitflags = "1.0"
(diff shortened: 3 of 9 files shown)


awortman-fastly:
rebased on #631, so this can pass the `cargo audit` step. fixed a few straggler cases of `Provided()` which this makes `Provided {}`.

benaubin:
@iximeow You should just be able to call [type_name()](https://doc.rust-lang.org/std/any/fn.type_name.html) on the `Box<dyn Any>`. Rust automatically provides the type name.

iximeow:
@benaubin what would the type parameter be? all i could coax out of `type_name` after-`Any`ing was a [`dyn Any`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=be662e5e479a9d206476174e883df79e):

```rust
use std::any::Any;

pub fn type_name_of_val_but_not_unstable<T: ?Sized>(_: &T) -> &'static str {
    std::any::type_name::<T>()
}

pub fn main() {
    let boxed: Box<dyn Any> = Box::new(1);
    println!("{}", type_name_of_val_but_not_unstable(&boxed));
    println!("{}", std::any::type_name::<dyn Any>());
}
```
->
```
alloc::boxed::Box<dyn core::any::Any>
dyn core::any::Any
```

benaubin:
Oh, yeah, just tried this in the Rust playground. Can't get it to give anything other than Any either 😞. I still really don't love the idea of passing through the type name as a string, though. But yeah, looks like there's no other way around it.

iximeow:
it turns out this was something of a happy accident, in fact. i don't think this should be debug-only because either the field simply doesn't exist in release builds (API's that change between build configs :fearful:) or the field is `Option<&'static str>` or just `""` in release builds (at which point, why?)

there isn't a performance overhead for grabbing the type name since the caller gets monomorphized on the name that it'll pick anyway. so the biggest issue i see is that you can reflect on the type name and do really gnarly stuff.

this is the happy accident: i'd meant to make the type name accessible outside the crate to print when we see an unexpected type somewhere. reading your comment, i realized i didn't add any helper for that. _but_ `Provided` now prints the type of whatever got provided, in its `Debug` impl! so it never needed to be accessible outside the crate in the first place.

what has you wanting type names not present in release builds?

pchickey:
I'll propose that, instead of an `Any` here, we change termination details to be a new `struct Tra

pchickey (dismissed):
sorry for the fat finger before! this is a good stop-gap but I'm going to advocate we figure out a way to mimic `wasmtime::Trap` for termination details and get rid of this `Any`, but thats a separate design discussion and this will help us find bugs now.

benaubin:
@pchickey Oh no! I really hope not: being able to provide our own termination details is extremely useful - we have an enum of custom reasons an instance may terminate, and it's especially useful to be able to send that back to the host context.

iximeow:
once more - had a merge conflict with #632.

@benaubin i think to-Any-or-not-to-Any will end up being a conversation elsewhere (as much as i don't love the `dyn Any`, i also don't particularly want to change it :grin:). i would point out that unless you're not using wiggle/wasi at all, i suspect you _do_ have at least two possible types of values there though: your enum and whatever gets provided to `lucet_hostcall_terminate!` in `proc_exit`. that was an `i32` or a `String`, but #632 makes that `wiggle::Trap`, leaving specific handling left up to users. so if something could possibly `proc_exit` in the webassembly guest, :skull: !

benaubin:
Makes sense. We are not using wasi.

pchickey (approved):

PR #631, opened 2021-02-17
bump wasmtime

iximeow:
this pulls in an updated wasmtime to fix our dependence on `rand_core` that recently had a RUSTSEC advisory. it also updates `raw-cpuid` to `9.0.0` to fix a few soundness issues: these were in functions we don't use (`as_string`/`processor_brand_string` and `cpuid_count`), but cargo audit is a hammer that strikes all nails.


awortman-fastly:
cut corners warning: a full `cargo update` includes bumping `criterion` from 0.3.3 to 0.3.4. this includes deprecation warnings for the various Lucet benchmarks, but i'm pretty uncertain how to rewrite them to the `BenchmarkGroup` API.

pchickey (approved):

PR #632, opened 2021-02-17
just pass the underlying wiggle::Trap to lucet_hostcall_terminate

iximeow:
another small one for the day: by passing `wiggle::Trap` we can ~force~ encourage users of `lucet` to handle the types that can arrive from a `TerminationDetails::Provided` by passing `wiggle::Trap` directly. then, users can be certain that the only types that can show up are their own provided types from `lucet_hostcall_terminate`, or `wiggle::Trap` via the wiggle bindings.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance.rs
@@ -1471,6 +1471,29 @@ impl TerminationDetails {
             _ => None,
         }
     }
+    /// Try to interpret the termination details as a provided exit code.
+    ///
+    /// The most consistent form of `TerminationDetails::Provided` comes from Lucet's
+    /// implementation of `proc_exit`, which exits with a `Provided` holding the given exit code.
+    /// For cases where a Lucet user simply wants "`proc_exit` or continue panicking" behavior,
+    /// `as_exitcode` can simplify handling `TerminationDetails`.
+    pub fn as_exitcode(&self) -> Option<u32> {
+        match self {
+            TerminationDetails::Provided(a) => {
+                // I apologize for this load-bearing `as u32`.
+                // Wasi uses an u32 for the proc_exist status (`lucet_wasi::Exitcode`) in the
+                // witx. However, wasmtime::Trap exit status is an i32, so the
+                // wiggle::Trap::I32Exit variant mirrors Wasmtime. The `as u32` lets this method
+                // return a type equivalent to `lucet_wasi::Exitcode`, but users interested in the
+                // full range of `wiggle::Trap` will have to handle an i32 variant.
+                match a.downcast_ref::<wiggle::Trap>() {
+                    Some(wiggle::Trap::I32Exit(code)) => Some(*code as u32),
+                    _ => None,
+                }
+            }
+            _ => None,
+        }
+    }
 }
 
 // Because of deref coercions, the code above was tricky to get right-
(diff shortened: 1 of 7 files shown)


lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1471,6 +1471,29 @@ impl TerminationDetails {
             _ => None,
         }
     }
+    /// Try to interpret the termination details as a provided exit code.
+    ///
+    /// The most consistent form of `TerminationDetails::Provided` comes from Lucet's
+    /// implementation of `proc_exit`, which exits with a `Provided` holding the given exit code.
+    /// For cases where a Lucet user simply wants "`proc_exit` or continue panicking" behavior,
+    /// `as_exitcode` can simplify handling `TerminationDetails`.
+    pub fn as_exitcode(&self) -> Option<u32> {
```
iximeow, line 1480: the return type here is kinda awkward. i'd _like_ to be able to write `lucet_wasi::types::Exitcode` here, but that would make a circular dependency with `lucet_runtime`.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1471,6 +1471,29 @@ impl TerminationDetails {
...
+        match self {
+            TerminationDetails::Provided(a) => {
+                // I apologize for this load-bearing `as u32`.
+                // Wasi uses an u32 for the proc_exist status (`lucet_wasi::Exitcode`) in the
+                // witx. However, wasmtime::Trap exit status is an i32, so the
+                // wiggle::Trap::I32Exit variant mirrors Wasmtime. The `as u32` lets this method
+                // return a type equivalent to `lucet_wasi::Exitcode`, but users interested in the
+                // full range of `wiggle::Trap` will have to handle an i32 variant.
+                match a.downcast_ref::<wiggle::Trap>() {
```
iximeow, line 1489: and this causes a `wiggle` dependency in `lucet_runtime_internals`. i think that's _okay_, but it's a weird spot where the runtime knows more about the bindings just for a types reason.

lucet-runtime/lucet-runtime-internals/src/instance.rs
pchickey: its fine to use a u32 here, in my opinion. We don't want to take a dep on lucet-wasi (or wasi-common, where that type really comes from) in the runtime.

pchickey: (it is slightly unfortunate that we have to take a dep on wiggle in the runtime, but I think it is fine in this case)

lucet-runtime/lucet-runtime-internals/src/instance.rs
pchickey: the other way to solve this problem would be to define a lucet_runtime_internals::Trap that has the same structure (it is, after all, pretty trivial) and have the lucet-wiggle boundary convert between them. We could then enforce that TerminationDetails is always a lucet_runtime::Trap as well, since this Any business is, in my opinion, not really worth the trouble. the wiggle Trap is designed to mimic wasmtime's and this would make lucet more like wasmtime, which I think is a win.

pchickey (approved):

pchickey (changes requested):
Also needs to be fixed in lucet-wasi/src/main.rs

iximeow:
great catch, updated there too.

pchickey (dismissed):

iximeow:
forgot to push the rustfmt commit..

pchickey (approved):

PR #636, opened 2021-02-22
Don't use non-constant string as panic!()'s first argument

jedisct1:
This is no longer accepted in Rust 2021.

Also remove a column sign after a function definition (how come that did even compile?)


acfoltzer:
These aren't strings, they're values that we use with `Any` in order to control the unwinding of guest termination and should keep them as-is.

We should get rid of that semicolon, though :slightly_smiling_face:

jedisct1:
Thanks!

I guess we have to wait for the next release and use [`panic_any`](https://github.com/rust-lang/rust/issues/78500`), then.

acfoltzer:
Ah yes, looks like it! I didn't know about that API change. Thanks for keeping eyes out for us

jedisct1:
Correct link to the issue: https://github.com/rust-lang/rust/issues/78500

PR #638, opened 2021-02-24
CI: Test uffd backend on github ci

pchickey:
Get rid of the Docker wrapper for Github CI so that we can run the full set of lucet tests on Github.

Using the `userfaultfd` syscall on Github CI wasn't possible when we first introduced that code, so we adopted CircleCI to fill the gap.

Rather than specify the dependencies of the various CI tests via the repo's `Dockerfile`, they are now installed directly in `.github/workflows/main.yml`. This does mean some repetition. 

In the process of putting these deps into the main.yml, I determined that Dockerfile itself contains way more dependencies than we actually need. But, I am not going to prune those today. CircleCI still uses the Dockerfile for its CI environment, and I guess users might use it too? Though I doubt it.

We can figure out if we are going to ditch CircleCI separately.


acfoltzer (approved):
LGTM. As mentioned on Slack I have some concerns about the reproducibility of failures that occur in this environment, but if we are still gating PRs on CircleCI and not GH Actions that shouldn't prevent us from shipping.

PR #643, opened 2021-02-26
Revert #626 (Refactor run_async custom Future): unsound

pchickey:
This reverts commit bf24022cc5e9c7e26ae12dd54c16fdbb26a45ddd, reversing
changes made to ddaf91e4875adfb4912be47f51620c6ffafd818d.

Unfortunately, in testing, we discovered that #626 has unsound interactions with asynchronous signal handling. We want to find a sound way to get these changes back in, but we need to revert the merge while we figure that out.

626 introduced a change to `lucet_runtime_internals::State` where a variant now contains an Arc:
```
    Running {
        async_context: Option<std::sync::Arc<AsyncContext>>,
    }
```

When a signal handler terminates an instance, https://github.com/BytecodeAlliance/lucet/blob/main/lucet-runtime/lucet-runtime-internals/src/instance/signals.rs#L310 will drop this Arc from a signal context, which will in turn call `free` on the underlying memory. If the system allocator is not signal safe (e.g. jemalloc) this can result in a segfault.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/future.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/future.rs
@@ -1,163 +1,112 @@
+use crate::error::Error;
 use crate::instance::{InstanceHandle, InternalRunResult, RunResult, State, TerminationDetails};
 use crate::module::FunctionHandle;
-use crate::val::Val;
+use crate::val::{UntypedRetVal, Val};
 use crate::vmctx::{Vmctx, VmctxInternal};
-use crate::{error::Error, instance::EmptyYieldVal};
+use std::any::Any;
+use std::future::Future;
 use std::pin::Pin;
-use std::sync::Arc;
-use std::task::Context;
-use std::task::Poll;
-use std::task::Waker;
-use std::{any::Any, future::Future};
+use std::task::{Context, Poll};
 
-/// a representation of AsyncContext which can be freely cloned
-#[doc(hidden)]
-#[derive(Clone)]
-pub struct AsyncContext {
-    waker: Waker,
-}
+/// This is the same type defined by the `futures` library, but we don't need the rest of the
+/// library for this purpose.
+type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
 
-const DEFAULT_INST_COUNT_BOUND: u64 = i64::MAX as u64;
+/// A unique type that wraps a boxed future with a boxed return value.
+///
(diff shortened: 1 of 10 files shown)


pchickey:
cc @benaubin

benaubin:
Uh oh! I bet it would be an easy fix to put the async context as a field on the instance instead of on the state. Would be happy to put together a fix :)

iximeow (approved):
:disappointed: i'm very excited about these changes and very _not_ excited about finding out `State` has to be signal-safe on drop. the exact drop is obscured by reassigning rather than some value going out of scope, and i'm not sure we'd even realize `State` has any kind of signal-safety requirements in the first place.

this is also a good time to just rip out the async signalling mechanism, the fuel-based timeout machinery is easy to keep safe by comparison but might need some thought first.

@benaubin i think a patch to move this to `Instance` would be great! i think rebasing #626 on the merge of this PR plus that change would be great for reviewability. looking at `State` i see we had (and will have again) `Box`es on Yielding and Yielded states, so generally i don't love how close we've been to hitting this for unrelated reasons - we'll have to keep an eye out for how `State` changes interact with signal safety in the future. Lucet will never `SIGALRM` a `Yielding` or `Yielded` state, but we've got very little type system guarantee that we wouldn't hit this later. :(

benaubin:
I mean, good to know, at least. #644 Should fix it (required minimal changes), and I'd be happy to rebase upon this PR.

In terms of the out-of-gas mechanism, I believe that an effective solution may be allowing calling a user-defined hostcall for the "on_bound_expired"? I'm not sure about the implications of that from a soundness perspective, but I'd imagine doing so would enable things like "check if cpu time limit expired" without the overhead of a context switch.

acfoltzer:
> this is also a good time to just rip out the async signalling mechanism, the fuel-based timeout machinery is easy to keep safe by comparison but might need some thought first.

This is how I would favor fixing this. It does mean that folks will have to use an async runtime to do timeouts, even if they're not doing any other async stuff. But that can be pretty lightweight between [`futures::executor::block_on()`](https://docs.rs/futures/0.3.13/futures/executor/fn.block_on.html) and [`futures-timer`](https://docs.rs/futures-timer/3.0.2/futures_timer/).

Of course, the current `KillSwitch` mechanism is more general and powerful than just a means to implement timeouts, so removing it could break those use cases. Are we aware of any such use cases? Is this worth a BCA RFC?

pchickey:
There are several ways we can proceed here, lets take our time figuring out what the best path is over on #644

PR #646, opened 2021-03-03
Make the new Cranelift x64 backend the default.

cfallin:
After sufficient testing, we're confident that the new x64 backend does
not have any correctness issues relative to the old backend; and because
it generates better code, we would like to switch the default. This PR
will cause `lucetc` to use the new backend by default. Building `lucetc`
with the `old-x64-backend` Cargo feature will swap the old backend in
again. This PR also removes the separate new-backend CI tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,21 +49,21 @@ jobs:
           cache_target: false
 
   ###################################################
-  # Linux `make test-ci` targets on new x64 backend #
+  # Linux `make test-ci` targets on old x64 backend #
   ###################################################
-  new_backend_ci_branch:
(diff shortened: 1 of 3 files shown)


pchickey:
Can we keep the old backend testing in CI for a little while longer? Instead of deleting the new-backend CI jobs we can change them to be old-backend.

cfallin:
> Can we keep the old backend testing in CI for a little while longer? Instead of deleting the new-backend CI jobs we can change them to be old-backend.

Yes, for sure, that makes sense! Updated.

pchickey (approved):
:tada:

PR #648, opened 2021-03-11
Fix wake in `UffdRegion` for invalid page access on Linux 5.7+.

peterhuene:
On Linux 5.7+, the kernel retries faults which is causing the uffd tests to
hang because the `UffdRegion` wakes invalid page accesses expecting the kernel
to signal `SIGBUS` rather than retry the fault.

This fix attempts to work around the recent kernel behavior by changing the
page protection to `NONE` for an invalid page access. The access is reset when
the `Alloc` is dropped or when the instance is reset.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
@@ -130,11 +130,15 @@ pub struct Alloc {
     pub heap_memory_size_limit: usize,
     pub slot: Option<Slot>,
     pub region: Arc<dyn RegionInternal>,
+    // The uffd region needs to keep track of accesses to pages that were "invalid" (i.e. should
+    // signal SIGBUS when accessed) so that the pages can be reset back to the appropriate
+    // protection level.
+    #[cfg(all(target_os = "linux", feature = "uffd"))]
+    pub invalid_pages: Vec<(usize, usize)>,
 }
 
 impl Drop for Alloc {
     fn drop(&mut self) {
-        // eprintln!("Alloc::drop()");
         self.region.clone().drop_alloc(self);
     }
 }
--- a/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/region/mmap.rs
@@ -8,7 +8,7 @@ use crate::sysdeps::host_page_size;
 use libc::c_void;
 #[cfg(not(target_os = "linux"))]
 use libc::memset;
-use nix::sys::mman::{madvise, mmap, munmap, MapFlags, MmapAdvise, ProtFlags};
+use nix::sys::mman::{madvise, mmap, mprotect, munmap, MapFlags, MmapAdvise, ProtFlags};
 use std::ptr;
 use std::sync::{Arc, RwLock, Weak};
 
@@ -140,6 +140,8 @@ impl RegionInternal for MmapRegion {
             heap_memory_size_limit,
             slot: Some(slot),
             region,
+            #[cfg(all(target_os = "linux", feature = "uffd"))]
+            invalid_pages: Vec::new(),
         };
 
         // Though this is a potential early return from the function, the Drop impl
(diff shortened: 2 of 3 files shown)


peterhuene:
With these changes, the uffd tests now pass on my Ubuntu VM running kernel 5.8.0.

lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
```diff
@@ -130,11 +130,15 @@ pub struct Alloc {
     pub heap_memory_size_limit: usize,
     pub slot: Option<Slot>,
     pub region: Arc<dyn RegionInternal>,
+    // The uffd region needs to keep track of accesses to pages that were "invalid" (i.e. should
+    // signal SIGBUS when accessed) so that the pages can be reset back to the appropriate
+    // protection level.
+    #[cfg(all(target_os = "linux", feature = "uffd"))]
+    pub invalid_pages: Vec<(usize, usize)>,
```
pchickey, line 137: I believe this be made a member of `UffdRegion`, eliminating the need for a cfg here - it looks to me like all the use sites are methods on UffdRegion, or functions called from one.

peterhuene: I think it could if we went through `RegionInternal` to say "mark this page as needing to be reset", as uffd's `page_fault_handler` has no other visibility into the region itself except through `Alloc`, I think; `MmapRegion` could simply no-op that request, perhaps?

Taking a step back though, if we move this into `UffdRegion`, it would necessitate a locking primitive to maintain `Sync` on `UffdRegion`, plus it would probably need to be a map as we need to know which alloc's pages to reset at the appropriate time.

This is working as-is now because we're lying to the compiler when we cast the raw `*mut Instance` to `&mut Instance`, so we can mutate the data directly on `Alloc` from the fault handler thread.  That's not something we can do if we move it into the region.

We do something similar in Wasmtime with the caveat that this is "safe-ish" if we assume the *only* way a page fault relating to an instance can occur is from the thread owning the instance, which we know the kernel has suspended and thus the only thread that *could* mutate the instance-related state is the fault handling thread. In Wasmtime, `Instance` is `!Send+!Sync`, so we depend on rustc to uphold that part of the safety guarantee.  Of course, if the page relating to the instance is faulted from some other memory unsoundness occurring within the runtime this assumption can be violated, but when that happens even the safety guarantees rustc could give us fly out the window anyway.

pchickey: Thank you for the correction! For some reason I forgot the one UffdRegion to many Alloc relationship, and you're right that we wouldn't want to introduce synchronization there when we already have exclusive access to the Alloc.

pchickey (dismissed):

lucet-runtime/lucet-runtime-internals/src/alloc/mod.rs
peterhuene: I was thinking of making this simply an `Option<(usize, usize)>` rather than a `Vec<(usize, usize)` as there *should* be only one invalid page access per instance as the instance shouldn't be reused until reset.  Perhaps we should do it that way and then effectively treat another page fault on an invalid page when there's already been an invalid access for the instance as an error condition?

pchickey: I believe you can get two pages to fault simultaneously if wasm does an unaligned word write that crosses page boundaries

peterhuene: That's an excellent point. We record it with a list in Wasmtime because we don't have any hard requirements on resetting a trapped instance before calling into it again (perhaps we should as the instance would likely be in an indeterminate state anyway).

At any rate, I can go either way re: `Vec<(usize, usize)>` vs. `[Option<(usize, usize)>; 2]`, but I'd like to avoid an allocation in the "invalid page was accessed" path if we can.

pchickey: Lets use a Vec and `tracing::warn!` if we see the vec have greater than 2 elements.

pchickey (approved):
:shipit:

PR #650, opened 2021-03-17
Add custom resource limiting for instances

aturon:
Lucet currently provides configurable limits for things like linear memory growth, but these limits are relatively "static" in that they are set at a specific value when setting up an instance.

For some use-cases, resource limits need to vary based on other conditions on the host, e.g. when attempting to limit both the WebAssembly heap _and_ the allocations resulting from calls into the host.

This PR addresses these additional use-cases by adding an optional, more dynamic `ResourceLimiter` trait that provides hooks for "approving" requests for additional resources dynamically, as they occur. A resource limiter can be configured as part of building an instance.

For now, the only check provided by `ResourceLimiter` is `memory_growing`, invoked during `Instance::grow_memory`.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/instance.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/instance.rs
@@ -76,6 +76,7 @@ pub fn new_instance_handle(
     module: Arc<dyn Module>,
     alloc: Alloc,
     embed_ctx: CtxMap,
+    limiter: Option<Box<dyn ResourceLimiter>>,
 ) -> Result<InstanceHandle, Error> {
     let inst = NonNull::new(instance)
         .ok_or_else(|| lucet_format_err!("instance pointer is null; this is a bug"))?;
@@ -90,7 +91,7 @@ pub fn new_instance_handle(
         needs_inst_drop: false,
     };
 
-    let inst = Instance::new(alloc, module, embed_ctx);
+    let inst = Instance::new(alloc, module, embed_ctx, limiter);
 
     unsafe {
         // this is wildly unsafe! you must be very careful to not let the drop impls run on the
@@ -275,6 +276,8 @@ pub struct Instance {
     /// fails, but this same instruction might also appear when other types of assertions fail,
     /// `panic!()` is called, etc. Terminating allows the error to be more directly identifiable.
     terminate_on_heap_oom: bool,
+
+    resource_limiter: Option<Box<dyn ResourceLimiter>>,
 
     /// `_padding` must be the last member of the structure.
     /// This marks where the padding starts to make the structure exactly 4096 bytes long.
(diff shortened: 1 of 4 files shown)


aturon:
I haven't dug into testing in Lucet yet, but if this PR generally seems acceptable I'm happy to add a test as well.

aturon:
[Related draft PR for wasmtime](https://github.com/bytecodealliance/wasmtime/pull/2736/).

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -76,6 +76,7 @@ pub fn new_instance_handle(
     module: Arc<dyn Module>,
     alloc: Alloc,
     embed_ctx: CtxMap,
+    limiter: Option<Box<dyn ResourceLimiter>>,
```
acfoltzer, line 79: This should probably be an `Arc` instead of a `Box` to make integration with other parts of a system easier and more efficient.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1611,3 +1630,25 @@ pub(crate) struct EmptyYieldVal;
...
+/// Used by hosts to limit resource consumption of instances.
+///
+/// An instance can be configured with a resource limiter so that hosts can take into account non-WebAssembly
+/// resource usage to determine if linear memory should be permitted to grow.
+pub trait ResourceLimiter {
+    /// Called when the the instance's linear memory has been requested to grow.
+    ///
+    /// * `current` gives the current size of the linear memory, in bytes.
+    /// * `desired` gives the desired new size of the linear memory, in bytes.
+    /// * `limit` gives the configured maximum size of the linear memory, in bytes.
```
acfoltzer, line 1643: I had to read through the call site above a couple times because this argument was kind of surprising to me. On a purely intuitive level, it feels like something called a `ResourceLimiter` shouldn't have to be told what its limit is. I wonder if this is a symptom of needing a more evocative name for this trait (perhaps `LimitEnforcer`?) or whether there's a different way of documenting its intended use.

lucet-runtime/lucet-runtime-internals/src/instance.rs
```diff
@@ -1611,3 +1630,25 @@ pub(crate) struct EmptyYieldVal;
...
+    ///
+    /// Returning `Ok` signals that the memory should be permitted to grow; returning an `Err(err)` prevents
+    /// the growth from occurring, yielding an `Error::LimitsExceeded(err)` from the call to grow the memory.
+    fn memory_growing(
+        &self,
+        instance: &Instance,
+        current: u32,
+        desired: u32,
+        limit: u32,
+    ) -> Result<(), String>;
```
acfoltzer, line 1653: `String` feels a bit weird here as an error value, I think because I have a hard time imagining we'd want a call to `memory_growing` to fail with any limit description other than "linear memory". Between this and the above comment about the `limit` argument I'm also wondering if I might just be thinking about this trait incorrectly :slightly_smiling_face:

acfoltzer (commented):
This looks great! I have a few questions about the details of the API, and how I'm maybe not thinking of it in quite the correct way. Also yes, it would be great to add some tests before proceeding.

PR #659, opened 2021-06-08
Update wasmtime, now with shiny new api

pchickey:
Update the wasmtime submodule to the latest `main`, which has landed the new-api PR.

Deletes all of the crates temporarily vendored in #655, points all those dependencies back into the wasmtime submodule.

Some small adjustments for changes between vendoring and landing new-api: I unwound a knot around wiggle-generate's AsyncConf/AsyncFunctions, I had to turn off wiggle's wasmtime integration, and the runtime borrow checker is no longer in a separate crate.

The latest wasmtime submodule was no longer compatible with Rust 1.50.0, so I upgraded to 1.52.1, and fixed all the ensuing warnings which require us to change `panic!(TerminationDetails...)` to `std::panic::panic_any(TerminationDetails...)`.


acfoltzer (approved):
Thanks for splitting up the commits like this, made for an easy review despite the lorge diff

PR #660, opened 2021-06-17
remove wasmtime submodule, depend on latest wasmtime packages from crates.io

pchickey:
Lucet is nearing EOL so we no longer need to track changes between releases in wasmtime. In case we do, we will use `[patch]` in Cargo.toml for a path or git dependency.

Having path dependencies in Lucet is causing integration headaches downstream, because you cannot `[patch]` over a path!

This PR removes the wasmtime submodule, and changes all deps to the latest crates.io release: cranelift-* is at 0.75.0, and the rest are at 0.28.0.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "wasmtime"]
-	path = wasmtime
-	url = https://github.com/bytecodealliance/wasmtime
 [submodule "lucet-spectest/spec"]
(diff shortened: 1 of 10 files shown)


iximeow (approved):
:heart: :crying_cat_face: :wave:

alexcrichton (approved):

PR #667, opened 2021-07-29
Update `userfaultfd` dependency to 0.3.2

acfoltzer:
This is to allow a newer version of bindgen to be picked up in the dependency tree.


acfoltzer:
Cargo audit found some stuff, so I will just combine this with a Rust 1.54.0 update to get all the gardening out of the way together

PR #670, opened 2021-08-11
uffd: fix eager heap init bug, add testing, update defaults to reflect production

pchickey:
* bugfix in the eager heap initialization: guard against initializing zero-sized memory. the zeropage ioctl will helpfully return EINVAL when this code path is exercised with initial_size = 0, e.g. when your wasm has `(memory (export "memory") 0)`
* add better testing coverage, checking the existing uffd test with both lazy (default, as before) and eager heap disposition, and add a test specifically for the bug fixed above.
* update `impl Default for UffdConfig` to the values we use in production.

### DIFF
--- a/lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
+++ b/lucet-runtime/lucet-runtime-internals/src/region/uffd.rs
@@ -710,9 +710,9 @@ impl Default for UffdConfig {
         UffdConfig {
             heap_page_size: HeapPageSize::Wasm,
             heap_init: Disposition::Lazy,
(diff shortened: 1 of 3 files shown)


acfoltzer:
> * add better testing coverage, checking the existing uffd test with both lazy (default, as before) and eager heap disposition, and add a test specifically for the bug fixed above.
> * update `impl Default for UffdConfig` to the values we use in production.

Are there commits missing here?

pchickey:
yep sorry I pushed to the wrong remote.

acfoltzer (approved):
Thanks for catching this!

PR #675, opened 2021-10-06
add maintence mode information to readme

### LINKED ISSUE #673: [Documentation] seems a little bit outdated
Hello Guys, 

I dont know whether this is the right place for this but i try to get my head around the lucet project for quite a while know because i want to create a small rust faas-application which embedding the lucet toolchain. You guys doing a great job! :-) 

But something that makes me a little bit confused is the documentation about this project because some parts of it seem to be very outdated. For example within lucetc there are many different options like witx_specs, wiggle_bindings or veriwasm (there are other, too) which are not documentated (But which are very important). 

In a very rapidly changing ecosystem like wasm this makes it sometimes really hard to stay on track with that project. Maybe there are other people thinking the same? Is there any documentation i maybe not found? Currently the repo is my first source of documentation about lucetc/lucet-wasi/etc. but there are some parts which are hard to understand.

pchickey:
Closes #673

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,15 @@
+## Lucet is in Maintence Mode
+
+Since mid 2020, the Lucet team has been focusing our efforts on the
+[Wasmtime][wasmtime] project. Wasmtime now has support for all of the features
+that previously only Lucet had, such as ahead-of-time (AOT) compilation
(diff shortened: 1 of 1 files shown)


tschneidereit (approved):
Thank you!

PR #686, opened 2021-11-01
upgrade to wasmtime 0.31 & cranelift 0.78

pchickey:
Wasmtime 0.31 and cranelift 0.78 is out :tada:

The big change here is the old backend is no longer in cranelift. This is good news and actually simplifies lucet a bunch! We no longer have to test the old backend in CI, or differentially fuzz against it.

The release was almost perfect, except the `ifcmp_sp` instruction that was dead code from the old backend is still required for lucet trampolines. So, @cfallin graciously added that back in, and we are using a `[patch.'crates.io']` directive in `Cargo.toml` to get that fix.


cfallin (approved):
Happy to see some of the API simplifications work their way down into the consumer end here -- looks good!

PR #692, opened 2022-01-20
update to rust 1.58.1

pchickey:
Update to rust 1.58.1, and fix some new warnings:
* benchmarks: delete inline(never) which is apparently unused
* with_ffi_arcs: Arc::into_raw now requires us to explicitly discard the
  result, which was the intention of the code anyway
* objdump: some of the struct fields were never read, who cares, mark
  the whole file allow(dead_code) because its heading to the bin anyway


cfallin (dismissed):
👍

cfallin (approved):

